PR #20, opened 2018-02-21
Change RainbowMod's name to Fabeline

leo60228:
I don't entirely see what mod name localization support has to do with this.

### DIFF
--- a/RainbowMod/metadata.yaml
+++ b/RainbowMod/metadata.yaml
@@ -1,4 +1,4 @@
-﻿Name: RainbowMod
+﻿Name: Fabeline
 Version: 1.0.0
 DLL: RainbowMod.dll
 Dependencies:


0x0ade:
_Oh,_ I thought you meant the mod options menu title. That's in the RainbowMod English.txt. Do you also want to change that?

leo60228:
I built it and it changed the mod options menu title... :thinking:

0x0ade:
Because it defaults to that when no other name is given. It'd be best to just set the mod options menu title in English.txt, unless you also want to change the settings and cached .dll path.

0x0ade:
I just pushed https://github.com/EverestAPI/Everest/commit/6d201abba4ca09d43a28e9f966dab6b835f08635 to master, which updates English.txt accordingly.

Do we still want to change the config and cached .dll paths by changing the mod name?

PR #21, opened 2018-02-25
Add Dualshock 3 detection for proper button icons

sammko:


0x0ade:
You're missing

```cs
using Microsoft.Xna.Framework.Input;
```

Once you add that, LGTM 👍

PR #25, opened 2018-09-17
[WIP] Fix nix package expression.

d3rped:
I've fixed most issues with the nix package, there is; however, still an error when calling the miniinstaller that I don't really know how to fix.

Changes and Motivations:
- Use fetchNuGet instead of calling nuget restore. Nuget can lead to non deterministic builds.
- Provide a shell.nix file for users (so there is no longer the need to actually install the Everest installer)

Build Instructions:
The project can be built by either invoking nix-shell or by running:
`nix-build -E "with import <nixpkgs> {}; callPackage ./default.nix {}"`

It builds on both nixos-18.09 and nixos-unstable.

ToDo:
- remove old remains of old attempts to get nix working.
- fix error during installation.

Error:
[miniinstaller-log.txt](https://github.com/EverestAPI/Everest/files/2390252/miniinstaller-log.txt)
For reference, here is the file/directory tree of the build/result folder:
[result.txt](https://github.com/EverestAPI/Everest/files/2390270/result.txt)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,4 +1,4 @@
-{ fetchNuGet, buildDotnetPackage, dotnetPackages }:
+{ fetchNuGet, buildDotnetPackage }:
 
 let
   HookedMethod = fetchNuGet {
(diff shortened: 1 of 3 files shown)


0x0ade:
@leo60228 Can you please review this? You're the only other person I know with a nix setup, and you've provided the initial package setup.

leo60228:
Thanks! This is great. I'll test it when I get home in a few hours.

leo60228:
Can you try `https://github.com/leo60228/Everest/tree/nixify`? I'm not home, so I can't test right now, but that should fix the error.

d3rped:
Seems good to me.
I can use the miniinstaller script and when I run the game (with the steam-run chroot) the Mod Options "menu entry" shows up.
Haven't done any further Testing yet but I suppose it works.

Is there anything else that should be done before it can be merged?

leo60228:
Not really, I don't think.

PR #40, opened 2019-04-19
Updated nix build dependencies.

d3rped:
The build dependencies in the default.nix file were out of date so I've updated them.


leo60228:
Thanks for this!

PR #44, opened 2019-06-24
(Partially) fix save loading in Everest 849

max4805:
This [XmlAttribute] somehow became a problem after the dependency upgrade: the save file loading fails with the exception `Cannot serialize member 'Areas' of type Celeste.AreaStats. XmlAttribute/XmlText cannot be used to encode complex types.`

Now the saves are loading (no more showing "Load failed") but there are still some mistakes (at least in strawberry count and unlocked level count).


max4805:
Removed some more [XmlAttribute] since `<TotalStrawberries>` and `<UnlockedAreas>` are not XML attributes

PR #48, opened 2019-08-02
[Suggestion] Display missing dependencies in-game

max4805:
Just a small addition I think would be useful for beginners. When a mod doesn't load because of a missing dependency, they are displayed in-game like this:
image

I went for "[xxx] not found" to keep the message short (if the line is too long it seems to break the display, so I also truncate the list if it is too long).

Don't hesitate with feedback about code style or whatever, I did not do a lot of C# development 😅 and don't hesitate to close the PR if you feel it's wrong, I did not spend that much time on it.

(For the "French translation" part, I'm French so I just translated the English strings. 😃)

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreModule.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModule.cs
@@ -246,7 +246,9 @@ namespace Celeste.Mod.Core {
                                 }
                             }
 
-                            menu.Add(new TextMenuExt.SubHeaderExt(mod.Item1.Name + " | v." + mod.Item1.VersionString + missingDepsString) { HeightExtra = 0f, TextColor = Color.PaleVioletRed });
+                            menu.Add(new TextMenuExt.SubHeaderExt(mod.Item1.Name + " | v." + mod.Item1.VersionString + missingDepsString) {
+                                HeightExtra = 0f, TextColor = Color.PaleVioletRed
+                            });
                         }
                     }
                 }
--- a/Celeste.Mod.mm/Mod/Core/CoreModule.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModule.cs
@@ -224,8 +224,30 @@ namespace Celeste.Mod.Core {
                     if (Everest.Loader.Delayed.Count > 0) {
                         menu.Add(new TextMenuExt.SubHeaderExt(Dialog.Clean("modoptions_coremodule_notloaded_a")) { HeightExtra = 0f, TextColor = Color.OrangeRed });
                         menu.Add(new TextMenuExt.SubHeaderExt(Dialog.Clean("modoptions_coremodule_notloaded_b")) { HeightExtra = 0f, TextColor = Color.OrangeRed });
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/Mod/Core/CoreModule.cs
```diff
@@ -224,8 +224,30 @@ public class CoreModule : EverestModule {
...
+                                    // format their names and versions, and join all of them in a single string
+                                    missingDepsString = string.Join(", ", missingDependencies.Select(dependency => dependency.Name + " | v." + dependency.VersionString));
+
+                                    // ensure that string is not too long, or else it would break the display
+                                    if (missingDepsString.Length > 40) {
+                                        missingDepsString = missingDepsString.Substring(0, 40) + "...";
+                                    }
+
+                                    // wrap that in a " ({list} not found)" message
+                                    missingDepsString = $" ({missingDepsString} {Dialog.Clean("modoptions_coremodule_notloaded_notfound")})";
```
leo60228: Appending to a string is slow. You might want to use https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder instead, but again, not sure it matters if it only executes on startup.

Celeste.Mod.mm/Mod/Core/CoreModule.cs
```diff
@@ -224,8 +224,30 @@ public class CoreModule : EverestModule {
...
                         menu.Add(new TextMenuExt.SubHeaderExt(Dialog.Clean("modoptions_coremodule_notloaded_b")) { HeightExtra = 0f, TextColor = Color.OrangeRed });
-                        foreach (Tuple<EverestModuleMetadata, Action> mod in Everest.Loader.Delayed)
-                            menu.Add(new TextMenuExt.SubHeaderExt(mod.Item1.Name + " | v." + mod.Item1.VersionString) { HeightExtra = 0f, TextColor = Color.PaleVioletRed });
+
+                        foreach (Tuple<EverestModuleMetadata, Action> mod in Everest.Loader.Delayed) {
+                            string missingDepsString = "";
+                            if(mod.Item1.Dependencies != null) {
+                                // check for missing dependencies
+                                List<EverestModuleMetadata> missingDependencies = mod.Item1.Dependencies
+                                    .FindAll(dep => !Everest.Loader.DependencyLoaded(dep));
```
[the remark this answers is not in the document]
0x0ade: It doesn't matter here. Not using Linq is a bonus, but this doesn't run multiple times every frame, so it's perfectly fine :)

Celeste.Mod.mm/Mod/Core/CoreModule.cs
0x0ade: You're right that appending to a string is "slow", but just like the Linq usage, interpolated strings are much, *much* easier to maintain than the "faster" StringBuilder, and the overhead is once again negligible here.

0x0ade:
Other than the remaining comment by leo, LGTM 👍 

By the way @leo60228, I'm more than surprised to see you here 😅

max4805:
I just fixed the "line being too long" issue 😀

PR #51, opened 2019-09-09
Chapter 9 update fixes

max4805:
The vanilla code for the BSideTitle constructor does something like:
```
1/ initialize artist name depending on level
2/ if artist name starts with "image:", load image from GUI atlas
```

So it crashes with a NullReferenceException on step 2 if we are in a custom map.
What this PR does is defining artist before the call to the original constructor.

(Will update this PR with new fixes if I find any before you can review it)


max4805:
Rainbow spinners from Farewell were rendered blue, and there are "stars behaving like blades" that were rendered as actual blades. The third commit fixes that. Thanks for the input on Discord!
(should we add an attribute to allow the star blade things to be used by mods?)

PR #52, opened 2019-09-30
Integrate Update Checker into Everest

max4805:
This aims to integrate https://gamebanana.com/gamefiles/9920 into Everest directly.

The part I'm the most unsure of here is the method extraction in `Everest.Updater`. The code used by the code mod is pretty much the same as Everest's own updater to download the package while keeping track of progress, so I tried extracting it into a method.

**Not ready to be merged yet**, as if you open a save file, close it, then install updates, Celeste will reboot into the save file directly, which isn't really what should happen.

### DIFF
--- a/Celeste.Mod.mm/Celeste.Mod.mm.csproj
+++ b/Celeste.Mod.mm/Celeste.Mod.mm.csproj
@@ -143,6 +143,7 @@
     <Compile Include="Mod\Helpers\EndUserException.cs" />
     <Compile Include="Mod\Helpers\SafeRoutine.cs" />
     <Compile Include="Mod\UI\OuiHelper_ChapterSelect_Reload.cs" />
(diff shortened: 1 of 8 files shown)


max4805:
Opening a save file, then exiting it and pressing Ctrl+F5 also makes the game automatically reload the save.

max4805:
Any advice about how to fix that oddity? This is caused by the fact that even when you exit a save file, `SaveData.Instance` still keeps its value. One way I found to "solve" this is to set it to `null` in `OuiMainMenu.Enter` (since this is always where you get by exiting a save file), but that sounds terrible. Checking in `QuickFullRestart` if we actually are in a save file also sounds painful 🤔

0x0ade:
That oddity is taken care of now :)

max4805:
Cool, thanks :) The update checker should always restart to the title screen now.

Issue #55, opened 2019-10-06
using Debug map in ANY save, causing losing B/C sides MapMetaModeProperties


orbittwz:
as discussed before in the channel...
I discovered it when playtesting my maps which have a special Starting Room setting.
I used the debug map, teleported to a room, and finished the A side from set room...
Entered B/C sides from the beginning, the spawn point was wrong, and was set to the room which was closest to 0,0 and not my Starting Room meta value.
FYI

0x0ade:
Fixed with #56.

PR #56, opened 2019-10-06
Fix F6 on an A-side wiping out B- and C-side properties

max4805:
I'm back at opening PRs for changes I'm unsure about... I feel like it's a more proper way to do this.

The issue here can be reproduced on [this map](https://cdn.discordapp.com/attachments/449297822827937795/629258383161032704/smaerdnogylop.zip): go into the A-side, F6 teleport to the end, then go into the B-side. You spawn near the end, as the "Start Level" property has been wiped.

This is caused by the Modes field in the map metadata being replaced with an array of 3 nulls. Here, I tried to add a check to ensure the array is not only filled with nulls before applying it. This fixes the issue... but I'm PRing this in case this has consequences I don't know about.

(While testing that, I noticed a small issue with the music: on this map, when I F6 teleport from the beginning to the end, the music only gets updated if I hold Ctrl. I might get to it later.)

### DIFF
--- a/Celeste.Mod.mm/Mod/Meta/MapMeta.cs
+++ b/Celeste.Mod.mm/Mod/Meta/MapMeta.cs
@@ -211,7 +211,7 @@ namespace Celeste.Mod.Meta {
                 if (!string.IsNullOrEmpty(Portraits))
                     meta.Portraits = Portraits;
 
-                if ((Modes?.Length ?? 0) != 0)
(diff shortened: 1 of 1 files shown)


0x0ade:
Looks good to me, can I merge this already? :) The music issue seems to be unrelated to the issue in this PR, as it rather seems to be session-related.

This issue fixes #55.

max4805:
Yeah, the music "issue" is not related at all to this. I just noticed it while fixing this, so I just reported it. ~~A PR is probably not the right place to do this though, sorry~~

PR #57, opened 2019-10-10
Automatic dependency downloader?

max4805:
Here is something that's not quite finished yet (there are TODOs for edge cases left in the code this time), but I chose to open a PR anyway to be able to discuss about this (for example, is that a good idea or is that too much of a mess? are the different cases here handled correctly?)

So, what's here is a thing that adds a button in the Mod Options to automatically download missing dependencies if there are any. This uses the same database as the update checker, as it provides what we need here: we can get a download link for a mod based on its name.

~~Note: the only mod that cannot be retrieved automatically by this (and by the update checker as well) is DJ Map Helper, because it has both a Windows and a Linux version.~~ Update (Oct 27): This is not the case anymore. Thanks DJ!

### DIFF
--- a/Celeste.Mod.mm/Celeste.Mod.mm.csproj
+++ b/Celeste.Mod.mm/Celeste.Mod.mm.csproj
@@ -129,6 +129,8 @@
     <Compile Include="Mod\Everest\LuaTypeBuilder.cs" />
     <Compile Include="Mod\Everest\Everest.LuaLoader.HookHelper.cs" />
     <Compile Include="Mod\Everest\Everest.LuaLoader.cs" />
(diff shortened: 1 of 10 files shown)


max4805:
I didn't mention it so I'm doing this now: I'm waiting for feedback before completing the remaining TODOs, because I'm unsure about all of this. There is no hurry at all though :)

max4805:
(latest commit has a typo, was supposed to be "Shorten French translation to _fit_ in the screen")

I just added a check to ensure the version available in the database will actually satisfy the dependency before installing it. 

I don't think checking if the missing dependency is actually installed but blacklisted is really worth it (this would imply checking all blacklisted mods' everest.yamls). @0x0ade if you agree with the latest changes, consider that this PR is ready to be merged :)

PR #60, opened 2019-10-27
Feed back stamp textures (fixes Bingo UI journal page crashes)

max4805:
This change in 1.3.0.0 broke Bingo UI (and Dash Count in Journal but this one got updated):
> Assist Mode, Variant Mode, and Cheat Mode File Slot icons changed from Stamps to Tabs

Bingo UI still tries to render the stamps for Assist Mode and Cheat Mode, and that gives a crash upon viewing the "binoculars" journal page.

I'm opening a PR for this because I'm unsure we want to do this. This fixes the crash, so it is ready to be merged. :)


0x0ade:
Someone might want to inform the dev of the bingo UI mod about this as well. For now, this is perfectly fine.

0x0ade:
uhh... did you force-push right after I merged?

max4805:
I'll make a post about this on the GB page then. The last update to this mod was 6 months ago though :/

max4805:
Also, I think my commit always appears after our comments because my computer's clock did not adjust according to daylight saving time, and this gives some... oddities

image

PR #63, opened 2019-11-10
Optimization in text menu rendering

max4805:
My mod install is quite a stress test for Everest, and quite a few of them have mod options (... like, 15 of them). So, I got pretty bad frame drops in the Mod Options menu (~30 FPS).

This is pretty much a copy-paste of the vanilla method for option rendering, adding a check to skip rendering of off-screen options. This really helps and I now get 60 FPS in the Mod Options menu.

... except on some instances where I get 50 FPS instead? I'm investigating on that. I'll tell you when I'm done with this 🙂

### DIFF
--- a/Celeste.Mod.mm/Patches/TextMenu.cs
+++ b/Celeste.Mod.mm/Patches/TextMenu.cs
@@ -3,6 +3,7 @@
 #pragma warning disable CS0169 // The field is never used
 
 using Celeste.Mod;
+using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Input;
 using Monocle;
 using MonoMod;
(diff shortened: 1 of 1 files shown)


max4805:
Might come from Vanilla Tweaks having two `[SettingRange(0,500)]` options. Continuing to investigate. 🤔

Vawlpe:
THANK YOU, opening mod options is so incredibly laggy that I just avoid it as much as possible...

max4805:
I can confirm the lag I'm still noticing is caused by Vanilla Tweaks having `[SettingRange(0,500)]` options: the game iterates through the 501 options to see which one is the biggest, to adjust the size in consequence. I think fixing the mod is the best option in this case (having 500 options is not exactly very handy anyway...).

So, consider that ready to be merged 🙂

max4805:
Just noticed I didn't put any comment everywhere to explain what the patch is even for. That's done now!

PR #64, opened 2019-11-13
"Auto-update mods on startup" option

max4805:
This was a suggestion on the Discord server: when the game starts, update all your mods automatically if you enable the option in Mod Options.

So, here is the plan:
- When mods are done registering, start checking for updates asynchronously. This way, the game won't take more time to boot if there is no update, as it will continue loading in the meantime.
- Between the intro screen and the title screen, add our own "auto-update" screen (this is what the MonoModRule is for). This is basically the intro screen's background, with the same cogwheel animation as the asset reloader, showing the update progress. I found it cleaner than injecting auto-update code into GameLoader or OverworldLoader for example.

This is not ready to be merged yet, but I'm opening the PR to get advice:
- does this sound like a good plan?
- what would be the best way to share the cogwheel animation? (Also, I'm not very comfortable with rendering code, so if I made mistakes even in my slightly edited copypaste, be sure to let me know!)


max4805:
Note to self (will do this tomorrow): if update checking is done by the time GameLoader ends and there is no update, proceed directly to OverworldLoader to save having to go through AutoModUpdater for nothing.

max4805:
After thinking about it, I'd rather not share code between the mod auto-updater and the asset reloader just for the sake of not having 7 similar lines of code to render the cogwheel...

I kept my dev version of Everest installed for the past 2 days, and it grabbed updates for a few mods without issues. This should now be ready to be merged.

PR #73, opened 2019-12-29
[Suggestion] Add [SettingMaxLength] attribute for string module settings

max4805:
For example, having this in a class extending EverestModuleSettings:
```cs
        [SettingMaxLength(30)]
        public string ThirtyCharacters { get; set; } = "Thirty";
```
will generate a string input screen limited to 30 characters, as opposed to the default 12.

Not using the attribute will still use the default max length of 12.

[Test mod with 3 options in case you want to test that.](https://github.com/EverestAPI/Everest/files/4008548/SettingLengthTest.zip)

### DIFF
--- a/Celeste.Mod.mm/Mod/Module/EverestModule.cs
+++ b/Celeste.Mod.mm/Mod/Module/EverestModule.cs
@@ -421,13 +421,16 @@ namespace Celeste.Mod {
                     ;
 
                 } else if (!inGame && propType == typeof(string)) {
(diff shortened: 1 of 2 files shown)


0x0ade:
Can this be merged already? Not sure whether to merge or not given how your PR is marked as a "suggestion" while it's looking good to me.

max4805:
Yes, this is ready to be merged. Sorry if that "suggestion" tag was confusing.

PR #77, opened 2020-01-07
Add "Star Climb Graphics Controller" for the graphics without the forced music

GreyMario:
Star Climb Controller is a hardcoded piece of shit, and people have been asking about splitting the graphics renderer out of it for use in mapping.

This proposes the Star Climb Graphics Controller as a feature added by Everest, allowing Star Jump Blocks to receive the graphical upgrade seen in 6A's feather introduction without forcing the music to be changed. As a bonus, the color of the rays and the background can be customized.


GreyMario:
Oh, and here's the [Ahorn plugin](https://gist.github.com/GreyMario/e6374d4ead9dd5bcfbc8879abb130ace) that was used for testing so it can be quickly gutted and folded into Ahorn/Maple.

PR #80, opened 2020-01-10
Fix boosts from walljumping on upwards-moving solids

CursedFlames:
Celeste's implementation of Player.WallJump has a bug where it adds 3 to the x coordinate when checking for solids to boost off of, instead of adding in the direction of the wall. This means that under specific circumstances (specifically upwards-moving solids), boosts work off of the left side, but not the right side.

This adds a patch to correctly check for solids in the direction of the wall, meaning boosts work on both sides of moving solids.

### DIFF
--- a/Celeste.Mod.mm/Patches/Player.cs
+++ b/Celeste.Mod.mm/Patches/Player.cs
@@ -131,6 +131,20 @@ namespace Celeste {
                 orig_WindMove(move);
         }
 
+        private extern void orig_WallJump(int dir);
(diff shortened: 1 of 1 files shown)


0x0ade:
You might want to ping @NoelFB about this on the Celeste Discord server as well so that it might get fixed in a future Celeste update.

max4805:
CursedFlames reported this (along with the fix) in the bug reports channel of the Celeste Discord.

PR #82, opened 2020-01-11
StrawberryRegistry for "just works" support for mod berries

GreyMario:
Berries may be added via `Celeste.Mod.StrawberryRegistry.Register(Type type, bool tracked = true, bool alternateCollectionRules = false)`

`type` - The Type of your registered berry, obtained via `typeof(MyBerry)` or similar.
`tracked` - Allow the registered berry to appear in the "maximum" berry count, checkpoint cards, and pause menu tracker.
`alternateCollectionRules` - Whether the berry defers "first in line" to another registered berry which does not have this flag set.

`CoreMapDataProcessor` reads data from the registry to dynamically inject MapDataFixup steps into itself, so all mod berries will correctly auto-assign checkpoints and orders if given `-1:-1`.

### DIFF
--- a/Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
+++ b/Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
@@ -60,20 +60,20 @@ namespace Celeste.Mod
         }
 
         // Register the strawberry or similar collectible with the Strawberry Registry, allowing it to be auto-collected at level end and be trackable.
-        public static void Register(Type type, string name, bool tracked = true, bool alternateCollectionRules = false)
+        public static void Register(Type type, string name, bool tracked = true, bool blocksNormalCollection = false)
         {
-            registeredBerries.Add(new RegisteredBerry(type, name, tracked, alternateCollectionRules));
+            registeredBerries.Add(new RegisteredBerry(type, name, tracked, blocksNormalCollection));
         }
 
         // Register the strawberry or similar collectible with the Strawberry Registry, allowing it to be auto-collected at level end and be trackable.
-        public static void Register(Type type, bool tracked = true, bool alternateCollectionRules = false)
+        public static void Register(Type type, bool tracked = true, bool blocksNormalCollection = false)
         {
             if (Attribute.IsDefined(type, typeof(CustomEntityAttribute)))
             {
                 CustomEntityAttribute attr = Attribute.GetCustomAttribute(type, typeof(CustomEntityAttribute)) as CustomEntityAttribute;
                 foreach (string id in attr.IDs)
                 {
-                    Register(type, id, tracked, alternateCollectionRules);
+                    Register(type, id, tracked, blocksNormalCollection);
                 }
             }
         }
--- a/Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
+++ b/Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
@@ -19,6 +19,7 @@ namespace Celeste.Mod
         private static ReadOnlyCollection<RegisteredBerry> _getRegisteredBerries;
         private static ReadOnlyCollection<RegisteredBerry> _getTrackableBerries;
         private static ReadOnlyCollection<Type> _getBerryTypes;
+        private static ReadOnlyCollection<string> _getBerryNames;
 
         // Return caches or create new ones
         public static ReadOnlyCollection<RegisteredBerry> GetRegisteredBerries()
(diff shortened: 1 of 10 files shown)


GreyMario:
Hold the phone, one last thing to patch.

Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
```diff
@@ -0,0 +1,154 @@
...
+                    !goldenCheck)
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        // Called by CoreMapDataProcessor in order to inject mod berries into the step list for fixup.
+        public static Dictionary<string, Action<BinaryPacker.Element>> GetBerriesToInject(MapDataFixup context)
```
0x0ade: Instead of `CoreMapDataProcessor.Reset` calling this and adding the entries to its steps dict, you could make `CoreMapDataProcessor.Run` check if the berry is trackable and call `Context.Run` instead.

Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
```diff
@@ -0,0 +1,154 @@
...
+            return trackDict;
+        }
+
+        public class RegisteredBerry
+        {
+            // registered berry as a Type
+            public readonly Type berryClass;
+
+            // entity:<passed berry's defined name>
+            public readonly string entityName;
```
0x0ade: As far as I can tell this string will never contain `entity:` at the beginning. Can you please clarify or fix the comment above it?

Celeste.Mod.mm/Patches/HeartGem.cs
```diff
@@ -22,9 +25,43 @@ public patch_HeartGem(Vector2 position)
             // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
         }
 
-        [MonoModIgnore] // We don't want to change anything about the method...
+        /*[MonoModIgnore] // We don't want to change anything about the method...
         [PatchHeartGemCollectRoutine] // ... except for manually manipulating the method via MonoModRules
-        private extern IEnumerator CollectRoutine(Player player);
+        private extern IEnumerator CollectRoutine(Player player);*/
```
0x0ade: Might as well just remove this commented section.

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -53,8 +55,28 @@ class patch_Level : Level {
         public extern void orig_RegisterAreaComplete();
         public new void RegisterAreaComplete() {
             bool completed = Completed;
-            orig_RegisterAreaComplete();
-            if (!completed) {
+            //orig_RegisterAreaComplete();
```
0x0ade: Might as well just remove this comment, the orig_ definition and mark this method with MonoModReplace.

Celeste.Mod.mm/Patches/Strawberry.cs
```diff
@@ -27,5 +27,20 @@ public patch_Strawberry(EntityData data, Vector2 offset, EntityID gid)
...
+        }
+
+        /*[MonoModReplace]
+        private new bool IsFirstStrawberry
+        {
+            get
+            {
+                return StrawberryRegistry.IsFirstStrawberry(this);
+            }
+        }*/
```
0x0ade: Any reason why this is here? 😅

GreyMario:
I believe the _functionality_ is complete now. Reviewing review.

Celeste.Mod.mm/Patches/Strawberry.cs
GreyMario: That was me trying and failing something else until I found out that, for some mysterious reason, the instruction I needed to find and replace was actually a callvirt instead of a call. Not sure how that even happened.

I'll go ahead and run down all the development breadcrumbs like this.

Celeste.Mod.mm/Patches/Level.cs
GreyMario: -pluck-

Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
GreyMario: That is _much_ more reasonable and I hate that I missed that. Consider it done. ..Literally as of this moment.

Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs
GreyMario: Doodled on it with some crayon to make it better.

GreyMario:
I gave it a quick test on the non-opengl version (that's, uh, the XNA version, right?) and everything _seems_ to be in working order.

GreyMario:
Aaaand we're done! (Wait, no we're not. Inconsistent names. Peck!)

A `RegisterStrawberryAttribute` takes required parameters `tracked` and `blocksCollection` in that order.

PR #86, opened 2020-01-12
Antisoftlock change from real time to in-game time

AnInternetTroll:
This change should fix the need to disable antisoftlock when frame skipping with TAS mod in particular, but hopefully it fixes for other mods which freeze time.
Thanks for the help from SuperSupermario24


0x0ade:
Holy heck is that commit log. It doesn't even list a single commit by you as far as I can see.

Please fix it.

max4805:
You made a PR to merge `master` into `stable`. Did you push your code somewhere?

PR #91, opened 2020-01-15
Fixed mooms being spoiled in mod campaign ticket sleeves

GreyMario:
Also fixed secretberries messing up the auto-assignment for checkpoints and IDs, which in turn messed up the tracker.

I'll go ahead and make the merge when I wake up if no one else does. I've tested this fairly thoroughly.


GreyMario:
Thanks, GitHub, for making it weird and difficult to allow collaborators to work together on PRs from their own separate forks.

PR #93, opened 2020-01-17
Unhardcode stamps and compute a lot of totals dynamically

max4805:
- MonoModRules replacing hardcoded values in the save file slot rendering method, to get correct stamps on mod campaigns
- New attributes on LevelSetStats to get maximum values for many things (cassette count, heart count, golden strawberry count)
- New attributes on MapData to have the total number of berries, including untracked ones, and get if a cassette was detected (much like DetectedHeartGem in vanilla)
- Update tickets each time the save slot is shown, in order to better handle switching between mod campaigns: switching level sets then going back to the file select screen has the slot display the campaign that was just chosen.

If you want to check the computed heart/berry/etc amounts are correct, load a save, switch to a mod campaign and use the `print_counts` command.

Note: the flying golden from ch1 is not counted in TotalGoldenStrawberries or in MaxGoldenStrawberries. This is because it is named memorialTextController instead of goldenBerry, and as such, doesn't get added to MapData.Goldenberries. What should be done to address that?

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreMapDataProcessor.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreMapDataProcessor.cs
@@ -197,14 +197,12 @@ namespace Celeste.Mod.Core {
 
         public override void Run(string stepName, BinaryPacker.Element el)
         {
-            if (stepName.Length > 7)
-            {
-                if (StrawberryRegistry.TrackableContains(el))
-                    stepName = "entity:strawberry";
+            if (StrawberryRegistry.TrackableContains(el))
+                stepName = "entity:strawberry";
 
-                if (StrawberryRegistry.GetRegisteredBerries().Any(berry => berry.entityName == el.Name))
-                    TotalStrawberriesIncludingUntracked++;
-            }
+            if (StrawberryRegistry.GetRegisteredBerries().Any(berry => berry.entityName == el.Name))
+                TotalStrawberriesIncludingUntracked++;
+
             base.Run(stepName, el);
         }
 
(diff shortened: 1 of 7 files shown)


max4805:
Am drunk, there are 25 golden berries in vanilla including the flying one. This makes the issue more embarrassing since this will prevent the golden stamp from appearing on vanilla saves (MaxGoldenStrawberries is hardcoded to 25 and TotalGoldenStrawberries doesn't count the flying golden)... **Not ready to be merged**, but worth having a look at.

max4805:
... no, vanilla checks for 25 golden berries before displaying the stamp, but there **are** 26 goldens (including the flying one and Farewell). So this just replaces an issue with another...?

GreyMario:
The winged golden is a weird case; it's technically a "gold berry", but if vanilla doesn't count it for the purposes of "total golden strawberries", maybe we *should* consider it "secret" for the purposes of tracking on save data? Goldens are goldens; Strawberry Registry doesn't provide any way to differentiate between goldens and secrets and considers them both "secret" so that they won't ever appear in the redberry trackers.

max4805:
... now it should be much better. Goldens (no matter if dashless or not) are considered "untracked" and "blocking collection" by the strawberry registry, so `MaxStrawberriesIncludingUntracked` will find them as well. Also did a fix to include dashless goldens to `Max/TotalGoldenStrawberries` so that the golden stamp appears only when they are collected as well.

Remaining question: vanilla is wrong, should we fix it? The golden stamp appears at 25 golden berries, but there are 26 in vanilla. As is, this PR doesn't fix it (following a "don't mess with vanilla" rule).

GreyMario:
I'd say let "fixing it" go to the Vanilla Tweaks mod, since fixing it *would* actually change vanilla behavior in the vanilla campaign.

Did another skim, nothing appears sad and broken to me. Are mod campaigns getting all shiny now? 👀

max4805:
Then I'd say this is ready to merge. :smile:

... well, the only golden campaigns _I_ get are the ones with no golden berries 😓 
image

(here are the stats behind that Path of Hope save. I got the cassette, completed the chapter and got all 7 strawberries, but not anything else:)
image

max4805:
"Fun" fact: this PR is as massive as the Update Checker PR. Good luck reviewing _that_.

PR #102, opened 2020-02-02
Fix Ctrl-F5 and updating on linux

PenguinOwl:
Unix-likes now fork-and-die when quick-restarted, and now clean env
variables left over by mono in between downloading the update and running
the installer.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
@@ -353,12 +353,16 @@ namespace Celeste.Mod {
 
                 progress.Progress = 1;
                 progress.ProgressMax = 1;
-                progress.LogLine("Restarting");
+                String action = "Restarting";
+                if (Environment.OSVersion.Platform == PlatformID.Unix) {
+                    action = "Updating";
+                }
+                progress.LogLine(action);
                 for (int i = 3; i > 0; --i) {
-                    progress.Lines[progress.Lines.Count - 1] = $"Restarting in {i}";
+                    progress.Lines[progress.Lines.Count - 1] = $"{action} in {i}";
                     Thread.Sleep(1000);
                 }
-                progress.Lines[progress.Lines.Count - 1] = $"Restarting";
+                progress.Lines[progress.Lines.Count - 1] = action;
 
                 // Start MiniInstaller in a separate process.
                 try {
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
```diff
@@ -368,13 +372,24 @@ public class Source {
...
                         }
                     }
                     installer.StartInfo.WorkingDirectory = extractedPath;
-                    installer.Start();
+                    if (Environment.OSVersion.Platform == PlatformID.Unix) {
+                        installer.StartInfo.UseShellExecute = false;
+                        installer.StartInfo.RedirectStandardOutput = true;
+                        installer.OutputDataReceived += (sender, args) => progress.LogLine(args.Data);
+                        installer.Start();
+                        progress.LogLine("Patching the game in-place...");
```
0x0ade: Patching the game in-place reportedly caused the game process to hang for some people, while for others, the installer was unable to write to the running assembly at all (just like on Windows). Is that still a concern to have with your in-place patching code?

Celeste.Mod.mm/Mod/Everest/Everest.cs
```diff
@@ -622,6 +619,15 @@ public static partial class Everest {
                 patch_Audio.System?.release();
             };
 
+            // Unix-likes can just fork-and-die to start the new game
+            if (Environment.OSVersion.Platform == PlatformID.Unix ||
+                Environment.OSVersion.Platform == PlatformID.MacOSX) {
+                Logger.Log(LogLevel.Info, "info", Path.Combine(PathGame, "Celeste"));
+                Process fork = new Process();
+                fork.StartInfo.FileName = Path.Combine(PathGame, "Celeste");
+                fork.Start();
```
0x0ade: Any reason why the process is being forked in `_QuickFullRestart` and not on shutdown like before? Doing it too early might cause Steam to see the game as "still alive" and prevent it from running a second time.

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
PenguinOwl: It shouldn't, but if that ends up being a problem again we could add a slight delay to the installer.

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
```diff
@@ -368,13 +372,24 @@ public class Source {
                     if (Type.GetType("Mono.Runtime") != null) {
                         installer.StartInfo.FileName = "mono";
                         installer.StartInfo.Arguments = $"\"{installerPath}\"";
-                        if (File.Exists("/bin/sh")) {
-                            installer.StartInfo.FileName = "/bin/sh";
-                            installer.StartInfo.Arguments = $"-c \"cd '{extractedPath}'; mono MiniInstaller.exe\"";
+                        if (File.Exists("/bin/bash")) {
+                            installer.StartInfo.FileName = "/bin/bash";
+                            installer.StartInfo.Arguments = $"-c \"cd '{extractedPath}'; unset MONO_PATH LD_LIBRARY_PATH LC_ALL MONO_CONFIG; /usr/bin/mono MiniInstaller.exe &> log.txt\"";
```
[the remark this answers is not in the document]
PenguinOwl: Fixed in 84f78d78fd42d2ad90f535b6cd76f5632f18aaef, environment variables are unset to clean up the mono environment so that it finds the system libraries rather than conflicting with Celeste's. (starting mono in a mono process without a clean environment is not a fun time)

Celeste.Mod.mm/Mod/Everest/Everest.cs
PenguinOwl: Yes. Shutdown is too late for Linux for whatever reason, and in all of my testing Steam waits plenty long enough for the parent Celeste process to die.

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
```diff
@@ -368,13 +372,24 @@ public class Source {
                     if (Type.GetType("Mono.Runtime") != null) {
                         installer.StartInfo.FileName = "mono";
                         installer.StartInfo.Arguments = $"\"{installerPath}\"";
-                        if (File.Exists("/bin/sh")) {
-                            installer.StartInfo.FileName = "/bin/sh";
-                            installer.StartInfo.Arguments = $"-c \"cd '{extractedPath}'; mono MiniInstaller.exe\"";
+                        if (File.Exists("/bin/bash")) {
```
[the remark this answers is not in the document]
PenguinOwl: I was using bash to pipe the log to `everest-update/log.txt`, but that was just mainly for debug purposes. Hooking into input in the parent process is not a great idea as the parent can die before input is over, leading to a broken pipe.

PR #105, opened 2020-02-12
Fix seekers crashing in rooms bigger than 200 tiles (width or height)

max4805:
**The following point should be considered before merging**: this makes the game laggy if used in rooms that are _too_ big. See [this video of Farewell with two seekers](https://cdn.discordapp.com/attachments/663557244620767264/677128175280586762/Celeste_2020-02-12_13-22-02.mp4). One seeker is perfectly fine (constant 60 FPS), but two of them is nearly unplayable.

So, this should be addressed:
- do we want to do that? 
- if so, do we want to put a higher limit, but still put a limit to prevent abuse?

Other than that, it has no impact on vanilla maps and existing custom maps, since it literally does nothing in this case (the array doesn't need to be extended). afaik, only the part of the array _within rooms boundaries_ are used when pathfinding, so extending it won't impact the seekers' behavior. Mods that patch the game to be able to spawn seekers in bigger rooms (that is, Crow Control and Extended Variants) don't need updating after this as well.


0x0ade:
> do we want to do that?

If anyone wants to optimize the path finder or use their own pathfinder for large rooms, they're welcome to do so and I'll happily merge it into Everest. As such, *performance* in large rooms shouldn't be *our* concern right now.

> if so, do we want to put a higher limit, but still put a limit to prevent abuse?

If anyone is going to abuse it, I assume people will just end up staying away from the map abusing it.

PR #106, opened 2020-02-17
Make TotalHeartGems "campaign-local" and patch some usages of it

max4805:
Getting 3 hearts in vanilla and 1 in some mod campaign allowed to enter Core... that felt _wrong_.

This PR aims at making `TotalHeartGems` only count crystal hearts for the current campaign, and also creates another property named `TotalHeartGemsInVanilla`, to be used in some cases when considering vanilla crystal hearts only makes more sense.

**`TotalHeartGems` usages:**
- `HeartGemDoor.HeartGems`: number of hearts collected to the heart doors' point of view. We want the heart count for the campaign here
- `AutoSplitterInfo.Update`: info for the auto-splitter...? We certainly want to keep the heart count for the active campaign here.
- `Commands.CmdHearts`: just displays the total heart count in the console. Unsure we care about that enough to patch it to something else as the heart count for the current mod campaign?
- `SaveData.CompletionPercent`: used to compute the completion percent for a campaign. Of course we want the heart count for the campaign here

**`TotalHeartGemsInVanilla` usages:**
- `OverworldLoader.CheckVariantsPostcardAtLaunch`: checks for 24 hearts on startup to unlock variants
- `OverworldLoader.Routine`: checks for 24 hearts upon exiting a level to unlock variants
- `HeartGem.RegisterAsCollected`: checks for 24 hearts to unlock the C-Sides achievement
- `OuiFileSelectSlot constructor`: checks for 24 hearts, again, to unlock variants

Another usage for `TotalHeartGems` was replaced with a check on `UnlockedModes`: this is used in the `OuiJournalGlobal` class to display or hide the "Total Golden Strawberries" row in the global Steam stats, only if goldens are unlocked (more than 16 heart gems). (Note: I got someone to send me an itch.io Celeste.exe long, long ago, and yes, `OuiJournalGlobal` is in this version as well. The journal page is just hidden.)

### DIFF
--- a/Celeste.Mod.mm/Celeste.Mod.mm.csproj
+++ b/Celeste.Mod.mm/Celeste.Mod.mm.csproj
@@ -193,6 +193,7 @@
     <Compile Include="Patches\Maddy3D.cs" />
     <Compile Include="Patches\ModeProperties.cs" />
     <Compile Include="Patches\MountainRenderer.cs" />
+    <Compile Include="Patches\OuiJournalGlobal.cs" />
     <Compile Include="Patches\Pathfinder.cs" />
     <Compile Include="Patches\CheckpointData.cs" />
     <Compile Include="Patches\SlashFx.cs" />
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -199,6 +199,24 @@ namespace MonoMod {
     [MonoModCustomMethodAttribute("PatchPathfinderRender")]
     class PatchPathfinderRenderAttribute : Attribute { };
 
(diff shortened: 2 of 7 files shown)


max4805:
I opened a PR to be sure we are agreeing on the rules about crystal hearts:
- 4-heart doors only open when you got 4 hearts in the **current** campaign
- variants only unlock when you got all hearts in the **vanilla** campaign
- the C-sides achievement is only unlocked when you got all hearts in the **vanilla** campaign 
- the "total golden berries" row in the Global Steam Stats journal page is only displayed if you unlocked goldens in the **current** campaign

If you agree with all of that, this is ready to be merged!

PR #112, opened 2020-03-04
Don't slow down the wake up animation when in space

max4805:
`UpdateSprite` finishes with this piece of code:
```cs
if (StateMachine.State != 11) {
	if (level.InSpace) {
		Sprite.Rate = 0.5f;
	} else {
		Sprite.Rate = 1f;
	}
}
```
so, in space, animations are twice as slow.

This causes the "intro wake up" animation to be twice as long, making it quite annoying. This patch turns the sprite rate back to `1f` when Madeline's state is `IntroWakeUp`.


max4805:
Small nitpick: I replaced "15" with "StIntroWakeUp". This is ready to be merged.

PR #119, opened 2020-03-21
Call a Discord webhook on successful stable builds

max4805:
Here is a small suggestion of webhook to announce new stable versions in #modding_updates on Discord. I did a bunch of experiments on a sandbox repo yesterday (using the same vm-image as Everest and a webhook posting to an empty Discord server) to ensure it works.

As we already figured when we published 1451, we don't _really_ need to announce any features since they already get announced sonner or later when they get on master. So, here is how it looks (build 17+42):
image

Of course I'm open to suggestions about the message itself etc.

If we want to set this up, I can create a webhook on the Celeste Discord, and this webhook URL will have to be set up as a secret variable in Azure Pipelines.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -51,6 +51,7 @@ steps:
     artifactName: 'main'
     publishLocation: 'Container'
 
+# Post-build steps (obsolete).
 - task: PowerShell@2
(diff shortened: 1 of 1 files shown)


max4805:
Taking it out of draft state, the pipeline ran as expected (the webhook step was skipped).

PR #121, opened 2020-03-23
Added optional "mods/modoptionsorder.txt" to manually sort ModOptions menu

coloursofnoise:
Uses the same format as "mods/blacklist.txt".


coloursofnoise:
I've never really used GitHub, so if any of this is incorrect, please let me know.

max4805:
Wondering if it would be good for it to create a sample file if it doesn't exist, like what Everest does for blacklist.txt?

PR #124, opened 2020-03-29
Fix unhandled error during ctrl-f5 on unix-likes

PenguinOwl:
This shitty patch was brought to you by linux gang™.
Time to go back into hiding.


0x0ade:
As already explained on Discord, a solution that doesn't just shove the exception under the rug would be more welcome.

PR #127, opened 2020-04-01
Updated modding_help channel name

coloursofnoise:


coloursofnoise:
I would wait to make a pull request until I have more changes to propose, but Github difficult

max4805:
https://github.com/EverestAPI/Everest/search?q=game_modding

There is another (way less visible) reference hiding in the in-game updater

PR #129, opened 2020-04-03
EverestCore can be ordered by modoptionsorder.txt using "Everest"

coloursofnoise:
"Update Everest" and "Install Dependencies" are always on top


Celeste.Mod.mm/Mod/UI/OuiModOptions.cs
```diff
@@ -41,9 +41,66 @@ public interface ISubmenu { }
...
+                    }));
+                }
+            }
+
             List<EverestModule> modules = new List<EverestModule>(Everest._Modules);
             if (Everest.Loader._ModOptionsOrder != null && Everest.Loader._ModOptionsOrder.Count > 0) {
                 foreach (string modName in Everest.Loader._ModOptionsOrder) {
+                    //Hack to allow user to position Everest Core in ModSettings
+                    if (modName.ToLower().Equals("everest")) {
+                        modules.Find(mod => mod.Metadata.Name == "Everest").CreateModMenuSection(menu, inGame, snapshot);
```
max4805, line 101: Everest doesn't get removed from the `modules` list, so the Everest Core options will get rendered twice, won't they?

PR #130, opened 2020-04-03
Command to load maps by SID from the debug console

max4805:
Here is an idea which... was not quite suggested, but that I found nice anyway. Some users were wondering how to load custom maps from the debug console, this is a simpler solution for them than using numerical IDs that can change when installing new maps.

This creates a `load_map` command, allowing to load a map based on its SID. For example: 
- `load_map Celeste/3-CelestialResort B 02` does pretty much the same as `hard 3 02`: load 3B on screen 02
- `load_map 1up/pathofhope C` loads Path of Hope C. The vanilla command to do so on my install is `rmx2 11`.

This PR also implements autocomplete on the SID for this command: typing `load_map 1` and pressing Tab will give `load_map 1up/pathofhope`.

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -253,6 +253,12 @@ namespace MonoMod {
     [MonoModCustomMethodAttribute("PatchCelesteMain")]
     class PatchCelesteMainAttribute : Attribute { };
 
+    /// <summary>
+    /// Removes the [Command] attribute from the matching vanilla method in Celeste.Commands.
+    /// </summary>
+    [MonoModCustomMethodAttribute("RemoveCommandAttributeFromVanillaLoadMethod")]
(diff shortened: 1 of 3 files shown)


0x0ade:
Personally, it'd make more sense to "merge" this into the existing `load` command, trying to parse the argument as a number first, interpreting it as a SID otherwise.

max4805:
What made me tilt towards a new command is the fact I needed to change the signature of the methods handling load/hard/rmx2 so that the first parameter is a string instead of an int... but that's definitely doable, I guess I was just too lazy 😅 I'll do that

max4805:
Fixed! The vanilla `load`, `hard` and `rmx2` commands now support SIDs instead. This is done by moving the [Command] attribute to a new method, that has a (string, string) signature. If the first parameter can be parsed as an int, the vanilla command will be called, or else a method that will do some checks and will convert the SID to a numerical ID will be called instead.

PR #140, opened 2020-04-12
Expose more elements of KeyboardConfigUI

PenguinOwl:
This patch makes adding new keyboard config screens much easier.
There might be a dead variable or two from the orig class, but most of
it should be good.

### DIFF
--- a/Celeste.Mod.mm/Patches/KeyboardConfigUI.cs
+++ b/Celeste.Mod.mm/Patches/KeyboardConfigUI.cs
@@ -166,7 +166,7 @@ namespace Celeste {
         /// </summary>
         /// <param name="mapping">The mapping</param>
         /// <returns>true if the key supports multiple bindings, false otherwise</returns>
-        private bool SupportsMultipleBindings(int mapping) {
+        protected virtual bool SupportsMultipleBindings(int mapping) {
             return !keysWithSingleBindings.Contains((Mappings) mapping);
         }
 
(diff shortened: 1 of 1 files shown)


max4805:
image

This changes the labels in the vanilla key binding menu.

max4805:
[I'm currently working on that, just to add comments and such.]

max4805:
When I wrote the docs for `GetRemapList` I noticed it was quite weird... since it did the remapping for keys that do not support multiple bindings (directional keys in vanilla), but just returned the list of _existing_ bindings for keys that support multiple bindings.

So I replaced this with a `RemapKey` method that does the remapping itself.

Also fixed the labels in the vanilla key binding menu by making the `GetLabel` default implementation call the vanilla `Label` method.

max4805:
I'm... pretty much done with my vacuum cleaning job here. Leaving that for review now!

PenguinOwl:
Thanks max for the clean-up!

One thing, I am actually still in favor of using `GetRemapList` over having modders remap the keys by themselves. It takes out a lot of the bloat in the custom class, as they only have to return the field in their settings file. It also saves them from implementing having to implement additive keys. For example,
```cs
protected override List<Keys> GetRemapList(int remapping, Keys newKey)
{
    Mappings mapping = (Mappings)remapping;
    if (mapping == Mappings.SpawnJelly)
    {
        return LiquidMod.Settings.SpawnJelly;
    }
    return null;
}
```
An easy way to document it would be
```
// Return the list to remap with. This should be a field in your settings class. To prevent the remap from happening, return null.
```

max4805:
Agreed for `GetRemapList`: even if the vanilla implementation is weird (and also _sets_ key bindings), this makes sense for modders implementing it and spares them ~5 lines of code they would have to write all the time to actually update the bindings.

Plus, I made `SupportsMultipleBindings` that is supposed to be overriden... private. Good job Max. :sweat:

max4805:
Looks good to me now. Thanks for fixing my fix 😅

PR #143, opened 2020-04-12
Prevent loading "assets" in hidden directories

PenguinOwl:
Helps mods that use SCM or ship with source code to easily hide their
git tree or source from being loaded by Everest as an asset.


0x0ade:
Implemented without a lengthy regex in 3cf90e8a6ba63225861670bca89997a9064e3a14.

PR #144, opened 2020-04-13
Custom 'color' attribute for fake hearts

### LINKED ISSUE #135: Fake heart colors are random
Fake hearts have randomized colors which change when editing a map, but seem consistent under no edits.

I suppose it would be worth adding an option that sets a specific heart color for individual fake hearts.

max4805:
Closes #135.

Having no "color" attribute or setting it to "Random" will leave the vanilla (pseudo) random value. Setting it to an area mode (Normal, BSide or CSide) will determine the heart color. In this case, the random value will still get generated, so that calls to the RNG that follow don't get impacted by it (1 random value is "consumed" by the heart, no matter if the heart color is customized or not).

### DIFF
--- a/Celeste.Mod.mm/Patches/FakeHeart.cs
+++ b/Celeste.Mod.mm/Patches/FakeHeart.cs
@@ -1,0 +1,39 @@
+﻿using Microsoft.Xna.Framework;
+using Monocle;
+using MonoMod;
+
+namespace Celeste {
+    public class patch_FakeHeart : FakeHeart {
+        public patch_FakeHeart(Vector2 position)
+            : base(position) {
+            // dummy constructor
+        }
+
+        // null is the vanilla (random) color.
+        private AreaMode? color;
+
+        [MonoModConstructor]
+        [MonoModIgnore]
+        public extern void ctor(Vector2 position);
+
+        [MonoModConstructor]
+        [MonoModReplace]
+        public void ctor(EntityData data, Vector2 offset) {
+            ctor(data.Position + offset);
+
+            if (data.Has("color") && data.Attr("color") != "Random") {
+                color = data.Enum<AreaMode>("color");
+            } else {
+                color = null;
+            }
+        }
+
+        [MonoModIgnore]
+        [PatchFakeHeartColor] // adds a call to _getCustomColor to override the random color
+        public extern override void Awake(Scene scene);
+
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/FakeHeart.cs
```diff
@@ -0,0 +1,39 @@
...
+        [MonoModConstructor]
+        [MonoModReplace]
+        public void ctor(EntityData data, Vector2 offset) {
+            ctor(data.Position + offset);
+
+            if (data.Has("color") && data.Attr("color") != "Random") {
+                color = data.Enum<AreaMode>("color");
+            } else {
+                color = null;
+            }
```
0x0ade, line 28: `EntityData.Enum` is case-insensitive, so the `Random` check should be as well.

```suggestion
            if (data.Has("color") && data.Attr("color").ToLower() != "random") {
                color = data.Enum<AreaMode>("color");
            } else {
                color = null;
            }
```

As discussed on Discord, it'd be better to default to random on invalid color values anyway.

```suggestion
            color = data.Enum<AreaMode>("color", (AreaMode) (-1));
```

Celeste.Mod.mm/Patches/FakeHeart.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Celeste {
+    public class patch_FakeHeart : FakeHeart {
+        public patch_FakeHeart(Vector2 position)
+            : base(position) {
+            // dummy constructor
+        }
+
+        // null is the vanilla (random) color.
+        private AreaMode? color;
```
0x0ade, line 13:
```suggestion
        // -1 is the vanilla (random) color.
        private AreaMode color;
```

Celeste.Mod.mm/Patches/FakeHeart.cs
```diff
@@ -0,0 +1,39 @@
...
+                color = null;
+            }
+        }
+
+        [MonoModIgnore]
+        [PatchFakeHeartColor] // adds a call to _getCustomColor to override the random color
+        public extern override void Awake(Scene scene);
+
+        private static AreaMode _getCustomColor(AreaMode vanillaColor, patch_FakeHeart self) {
+            return self.color ?? vanillaColor;
```
0x0ade, line 36:
```suggestion
            return self.color != (AreaMode) (-1) ? self.color : vanillaColor;
```

PR #145, opened 2020-04-14
Input prompt on AreaComplete screen

coloursofnoise:


Celeste.Mod.mm/Patches/AreaComplete.cs
```diff
@@ -59,6 +67,18 @@ public patch_AreaComplete(Session session, XmlElement xml, Atlas atlas, HiresSno
             DisposeAreaCompleteInfoForEverest();
         }
 
+        [MonoModReplace]
+        private void RenderUI() {
```
0x0ade, line 71: Any reason why you're replacing the method instead of reusing the orig_inal code? It seems like you're effectively adding code after orig_RenderUI anyway, keeping the rest unmodified. Or I'm blind 😅

PR #146, opened 2020-04-14
Custom Tutorial Bird (+ Trigger) pulled from Spring Collab 2020

max4805:
Differences are:
- more comments
- the CustomBirdTutorial.Awake method was added to show the bubble if no trigger is placed (for cases a trigger isn't necessary)
- CustomBirdTutorial.FindById: a helper method in case a mod wants to find a bird by birdId.

[Demo video](https://cdn.discordapp.com/attachments/445236692136230943/699619827798638672/Celeste_2020-04-14_15-49-49.mp4)

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/CustomBirdTutorial.cs
+++ b/Celeste.Mod.mm/Mod/Entities/CustomBirdTutorial.cs
@@ -1,0 +1,132 @@
+﻿using Microsoft.Xna.Framework;
+using Monocle;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


max4805:
Of note that I didn't carry on the entity name from Spring Collab 2020, since this one behaves _slightly_ differently, and that the collab bird is likely to stay in the collab helper anyway. Is that sounding good?

0x0ade:
Compatibility being broken hurts a little, but it's completely understandable.

PR #148, opened 2020-04-16
In-game blacklist.txt editor (aka "Mod toggler")

max4805:
Good bad idea of the day: have a big list of all your mods in Mod Options and be able to turn some of them off / cherry pick which ones you want to turn on.

image

**This is not a true mod toggler**, as in, it won't load/unload mods at runtime. When you exit the menu, it will write all disabled mods to blacklist.txt, delete whitelist.txt, and restart the game. And as such, I don't know if that's such a good idea or not. So, feel free to suggest stuff or to trash the idea entirely.

### DIFF
--- a/Celeste.Mod.mm/Content/Dialog/English.txt
+++ b/Celeste.Mod.mm/Content/Dialog/English.txt
@@ -176,8 +176,9 @@
 
 # Mod Toggle Menu
 	MODOPTIONS_MODTOGGLE=				TOGGLE MODS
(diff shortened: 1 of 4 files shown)


max4805:
New commit: from now, only the blacklist is taken into account by this screen. When a whitelist is active, a warning is displayed about it, and it is ignored (not overwritten, not removed, nothing). This should be a better base for discussion now, and it can be merged if everything looks good!

PR #155, opened 2020-04-22
Add Disable Transition Fading attribute to Fake Walls

max4805:
Per Cruor's request, add an attribute for preventing fake walls from fading in and out.


max4805:
After some discussion, turns out there is no real situation where this is useful. Closing.

Issue #158, opened 2020-04-30
Debug map explorer: key to re-center view on the player's current room


GreyMario:
In short, this is the problem you get when you peek at the F6 screen at the start of Farewell, then play through to the moon berry and then press F6 again--oh, hey, it's still showing the first room.

This is fine and all, but having a button to re-center the view is helpful.

PR #169, opened 2020-05-17
Fix for #158 (center map editor on respawn point)

DashingCat:
If the map loaded in the editor is the one currently being played:
- opening the map editor will center the view on the current respawn point.
- pressing F2 after moving in the editor will also re-center the view to the current respawn point.

### DIFF
--- a/Celeste.Mod.mm/Patches/MapEditor.cs
+++ b/Celeste.Mod.mm/Patches/MapEditor.cs
@@ -31,6 +31,7 @@ namespace Celeste.Editor {
                                           "Cancel:       Exit debug map\n" +
                                           "Q:            Show red berries\n" +
                                           "F1:           Show keys\n" +
+                                          "F2:           Center on current respawn point\n" +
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/MapEditor.cs
```diff
@@ -56,30 +57,18 @@ public patch_MapEditor(AreaKey area, bool reloadMapData = true)
...
-            Vector2 pos;
-            if (CurrentSession.RespawnPoint != null) {
-                pos = CurrentSession.RespawnPoint.Value;
-            } else {
-                Point lvlCenter = CurrentSession.LevelData.Bounds.Center;
-                pos = new Vector2(lvlCenter.X, lvlCenter.Y);
-            }
-
             Camera.CenterOrigin();
             Camera.Zoom = 6f;
```
DemoJameson, line 70: I think it would be better not to reset Camera.Zoom every time.

PR #172, opened 2020-05-18
Greatly speed up texture loading for runtime mod loading

max4805:
This is a track to speed up runtime mod loading.

To try that out, I started up the game with the collab and all its dependencies minus Extended Variants, then headed to Mod Options to install the missing dependency. Results:
- without this patch: about 1 minute and 5 seconds
- with this patch: about **15 seconds**

This code doesn't look good though... but it greatly speeds up things, because Everest isn't limited to loading 1 texture per frame. So, this is at least a track to explore! Please tell me what can be done to do this in a cleaner way.

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/MainThreadHelper.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/MainThreadHelper.cs
@@ -115,12 +115,19 @@ namespace Celeste.Mod {
         }
 
         public override void Update(GameTime gameTime) {
-            while (Queue.Count > 0) {
-                Action action;
-                lock (Queue) {
-                    action = Queue.Dequeue();
+            if (Queue.Count > 0) {
+                // run as many tasks as possible in 10 milliseconds (a frame is ~16ms).
+                Stopwatch stopwatch = new Stopwatch();
+                stopwatch.Start();
+                while (stopwatch.ElapsedMilliseconds < 10) {
+                    Action action = null;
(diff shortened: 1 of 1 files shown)


0x0ade:
> This code doesn't look good though...

What do you mean? It looks very similar to how I would've implemented it 😅 please stop your stopwatches though. And the Stopwatch class has got a static method to create + start a new stopwatch in one call.

max4805:
I just tend to find "actively wait in a for loop for something to happen" kind of code quite bad 😅 but since this lasts for 10ms and that the queue is locked when checked, this might be fine.
I'll apply your feedback in about an hour 👍

0x0ade:
Ah, I see what you mean now 😅 did you notice any issues when leaving the loop early instead of waiting for the next queued action?

max4805:
This speed up uses the fact that we have the time to load more than 1 texture during a frame; without that loop, what happens is "we load a texture then get out of the loop > the thread that was waiting for the texture continues and queues loading the next one > it is picked up on the next frame". That limited texture loading to 60 textures/second, which is pretty noticeable for the collab.

I didn't notice issues running that code, I just didn't like the principle of it but couldn't think of anything I liked better. 😅

0x0ade:
Guess we can always improve this later if it ever becomes an issue. A few milliseconds going to waste to wait for new tasks while already on a loading screen doesn't sound bad anyway - this would've been different if it had to load mid-game or something :P

PR #177, opened 2020-05-25
Ensure that Celeste is not modded before patching

DashingCat:
Check if the Celeste assembly has been patched before actually patching it.

(Slightly changed wording so users should be able to identify the root cause of the installation failure more easily)

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -382,11 +382,18 @@ namespace MonoMod {
                 version = new Version(versionInts[0], versionInts[1], versionInts[2], versionInts[3]);
             }
 
+            // Check if Celeste version is supported
             Version versionMin = new Version(1, 3, 1, 2);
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -382,11 +382,18 @@ static class MonoModRules {
                 version = new Version(versionInts[0], versionInts[1], versionInts[2], versionInts[3]);
             }
 
+            // Check if Celeste version is supported
             Version versionMin = new Version(1, 3, 1, 2);
             if (version.Major == 0)
                 version = versionMin;
             if (version < versionMin)
-                throw new Exception($"Unsupported version of Celeste: {version}");
+                throw new Exception($"Unsupported version of Celeste: {version}. The minimum required version of Celeste is: {versionMin}.");
```
0x0ade, line 390: The minimum required version was not shown specifically to throw off any pirates unable to find the actually latest version.

Issue #101, opened 2020-01-27
Personal to-do tweaks for Star Jump blocks


GreyMario:
- Option to disable the railing per block
- *Not* visually connect adjacent sinking/stationary blocks
- Maybe connect adjacent sinking blocks like vanilla does with floaty space?
- Inside corner tiles.

bigkahuna443:
Determined to be out of scope for Everest, see #208.

PR #182, opened 2020-05-29
Make StarJumpBlock act similar to FloatySpaceBlock

coloursofnoise:
Add "hasRail" option
Remove visual connection between moving and non-moving blocks
Add StarJumpThru


GreyMario:
Who huh what

Sorry, I'm not in the best position right now to really sniff at this, can't put this through its paces. Nice to see new QoL upgrades for starjumps, though.

A couple of questions, just to verify what's going on:

- Do floaty and static starjumps now ignore the opposite type when determining what tiles to draw?
- Any weirdness with attaching other objects, such as a spinner at an inside corner?
- Do rails end cleanly when a sequence of blocks goes rail-norail-rail horizontally?
- Was a plan considered to allow for starjumps that do *not* connect when adjacent, whether or not they are floaty? Just a thought, don't mind me.

GreyMario:
(Screw the mobile interface.)

GreyMario:
Oh, and it looks like this closes #101. Neat. 📸

PR #185, opened 2020-06-03
Add support for custom vignettes on level entry.

coloursofnoise:
LoadingVignetteScreen uses the same format as CompleteScreen

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/CustomTextVignette.cs
+++ b/Celeste.Mod.mm/Mod/Entities/CustomTextVignette.cs
@@ -1,0 +1,188 @@
+﻿using FMOD.Studio;
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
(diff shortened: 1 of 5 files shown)


Celeste.Mod.mm/Mod/Entities/CustomTextVignette.cs
```diff
@@ -0,0 +1,188 @@
...
+            started = true;
+            exiting = true;
+        }
+
+        private void ReturnToMap() {
+            StopSfx();
+            menu.RemoveSelf();
+            menu = null;
+            exiting = true;
+            bool toAreaQuit = ((patch_SaveData) SaveData.Instance).Areas_Safe[0].Modes[0].Completed && Celeste.PlayMode != Celeste.PlayModes.Event;
```
0x0ade, line 148: It should be possible to clean up the cast + \_Safe access to `SaveData.Instance.Areas`

PR #191, opened 2020-06-09
Add advanced option for AppDomain restart on Ctrl+F5

maddie480:
Add a flag in mod settings to allow picking between AppDomain restart and process restart when hitting Ctrl+F5 and similar.

I don't really know what the default value for this should be... AppDomain restarting is slightly faster (1 or 2 seconds from my testing), but it is also leaking memory. If someone has many mods and updates their mods, the game will restart and use way more memory than it should (or run out of memory if they really have many and/or use ANGLE), so it might be a good idea to make AppDomain restarting opt-in instead of opt-out. Any thoughts about that? 

Of note that process restarting does not trigger the "preparing to launch Celeste" thing from Steam.

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
@@ -183,7 +183,7 @@ namespace Celeste.Mod.Core {
         [SettingNeedsRelaunch]
         [SettingInGame(false)]
         [SettingIgnore] // TODO: Show as advanced setting.
-        public bool RestartAppDomain { get; set; } = true;
+        public bool RestartAppDomain_WIP { get; set; } = false;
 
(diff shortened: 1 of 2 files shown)


0x0ade:
> Of note that process restarting does not trigger the "preparing to launch Celeste" thing from Steam.

Is that with the vanilla Steamworks.NET.dll?

> so it might be a good idea to make AppDomain restarting opt-in instead of opt-out. Any thoughts about that?

Opt-in until it's fixed (which makes fixing it less critical, but eh). Maybe add a _WIP suffix to the property like with code reloading?

max4805:
Yes, I'm using the vanilla Steamworks.NET DLL (I'm getting the popup when the game starts up, and the overlay). I'm guessing this is because the new Celeste process is a child of the old one?

I added _WIP at the end of the setting name, and made it opt-in.

PR #196, opened 2020-06-14
Try fixing IL hook for Rumble Trigger in Steam FNA

maddie480:
The strategy is copying over the instructions _before_ the existing ones instead of after, since the instructions at the end of an if condition vary between Steam FNA and other versions: Steam FNA stores the check result in a bool local variable and checks it again right after instead of just branching.

Pushing as a draft PR because the instructions look fine, but **the branch targets turn into garbage** (for example `IL_0056: br.s IL_ffffffd9`) and I can't figure out why. I'm pushing this for someone else to pick up, since I have other topics to handle and that I'm starting to need a break. 😅


coloursofnoise:
Fixed by bulk replacing any short-form opcodes that are now out of bounds with their long-form counterparts.

PR #209, opened 2020-07-06
Custom Memorial: add a dreamingMode setting to override the metadata 'dreaming' setting

maddie480:
As it turns out, there is a "dreaming" checkbox in Ahorn, but the custom memorial ignores it, so it doesn't work. :maddyS:

So, here is a PR that adds a "tri-state" called dreamingMode and that can take 3 values: On, Off or BasedOnMetadata which is the default and is what the custom memorial currently does.

This PR is a reopening of PR #205, that got closed after the branch renaming.

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/CustomMemorial.cs
+++ b/Celeste.Mod.mm/Mod/Entities/CustomMemorial.cs
@@ -15,9 +15,14 @@ namespace Celeste.Mod.Entities {
     /// Checks for the following new attributes:
     /// - `string dialog` (default: `memorial`)
     /// - `string sprite` (default: `scenery/memorial/memorial`)
+    /// - `float spacing` (default: 8)
+    /// - `DreamingMode dreamingMode` (default: BasedOnMetadata)
     /// </summary>
     [CustomEntity("everest/memorial")]
     public class CustomMemorial : Entity {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/Entities/CustomMemorial.cs
```diff
@@ -46,18 +56,17 @@ public CustomMemorial(Vector2 position, string texture, string dialog, float spa
             Add(loopingSfx = new SoundSource());
         }
 
-        public CustomMemorial(EntityData data, Vector2 offset)
-            : this(data.Position + offset, data.Attr("sprite", "scenery/memorial/memorial"), data.Attr("dialog", "memorial"), data.Float("spacing", 8f)) {
-        }
+        public CustomMemorial(EntityData data, Vector2 offset) : this(data.Position + offset, data.Attr("sprite", "scenery/memorial/memorial"),
+            data.Attr("dialog", "memorial"), data.Float("spacing", 8f), data.Enum("dreamingMode", DreamingMode.BasedOnMetadata)) { }
```
0x0ade, line 60: cc @Cruor is this fine?

maddie480:
This will be moved in a helper if needed, and the currently broken "dreaming" checkbox will be removed in the next Ahorn update.

PR #213, opened 2020-07-14
Add support for custom overworld music through map metadata

maddie480:
Can be used like this:
```cs
Mountain:
    BackgroundMusic: event:/max480/test_music
```

Calling Audio.SetMusic repeatedly is fine, since this method doesn't do anything if the music you're asking for is already playing.


maddie480:
> Calling Audio.SetMusic repeatedly is fine, since this method doesn't do anything if the music you're asking for is already playing.

Nevermind that, restoring normal music all the time destroys menus that change the music, like credits and the sound test. So I'm now using a flag to restore normal music only if it has been changed.

PR #218, opened 2020-07-29
Add Logger.SetLevel and associated functionality

rhelmot:
This allows mods to have very verbose logging (think: hundreds of megabytes) that is disabled by default without having to comment anything out.


maddie480:
I started working on something on my side yesterday to set the logger level _depending on the tag_ (with regex). The issue I can see here is that you're setting the logger level **for all mods**, which other mods might not want (especially since the default log level is verbose).

rhelmot:
That does make sense :(

maddie480:
should we close this PR and work on #219 instead?

rhelmot:
yep!

PR #219, opened 2020-07-29
Customize log level through modsettings-Everest.celeste and code

maddie480:
Allows customizing log levels through two ways:
- **in code**: a mod can call this in its Load method:
```cs
Logger.SetLogLevel(new Regex("^MyCoolMod/"), LogLevel.Info);
```
This will make only Info and higher logs appear in log.txt for tags starting with "MyCoolMod/" by default.
- **in modsettings-Everest.celeste**: by adding this to the file:
```yaml
LogLevels:
  ^ExtendedVariantMode/: Info
  ^DJMapHelper/: Info
```
someone can customize the log levels for each mod.

YAML log levels take priority over log levels set through code: the idea is that the developer of the mod can set the log level to Verbose while working on it, and have the default log level set to Info through code. This way, verbose logs only appear in the developer's log, and don't clutter other people's logs.

@rhelmot would this cover what you need? 😅

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Logger.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Logger.cs
@@ -5,32 +5,31 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace Celeste.Mod {
     public static class Logger {
 
-        private static Dictionary<Regex, LogLevel> minimumLevels = new Dictionary<Regex, LogLevel>();
-        private static Dictionary<Regex, LogLevel> minimumLevelsFromEverestSettings = new Dictionary<Regex, LogLevel>();
+        private static Dictionary<string, LogLevel> minimumLevels = new Dictionary<string, LogLevel>();
+        private static Dictionary<string, LogLevel> minimumLevelsFromEverestSettings = new Dictionary<string, LogLevel>();
         private static Dictionary<string, LogLevel> minimumLevelsCache = new Dictionary<string, LogLevel>();
 
         /// <summary>
-        /// Sets the minimum log level to be written in the logs for lines matching the given tag.
-        /// When using this, make sure the tag regex is restrictive enough not to impact other mods
-        /// (for example, if all your tags follow the format MyMod/xxx, use "^MyMod/" as a regex).
+        /// Sets the minimum log level to be written in the logs for lines matching the given tag prefix.
+        /// When using this, make sure the tag prefix is restrictive enough not to impact other mods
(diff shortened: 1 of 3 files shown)


rhelmot:
ooh, yes, this would definitely work. My only concern is that adding a regex match to every single logging statement would slow things down a bit, would it be worth it to switch to exact string matching instead? It seems like there's a convention I wasn't aware of of logging with a label separated by slashes, so I guess if a lot of mods do that then regexes would be important... but on the other hand we could do what python logging does and institutionalize the hierarchy of logging so for example setting the loglevel for foo/bar would affect the logging on foo/bar and foo/bar/baz.

But that's all very complicated and if the regex is fast enough then the regex is fast enough.

maddie480:
I tried mitigating the issue with regex by creating a "cache" that would store the log level for each exact tag... but when thinking about it, we could work with prefixes instead of regex (for example, "ExtendedVariantMode/" could set the logging level of everything in Extended Variants), and in case of multiple matches, the longest prefix matching wins. I could implement that, would be less heavy than regex and still cover most cases I think.

maddie480:
So new examples are:
- in code
```cs
Logger.SetLogLevel("MyCoolMod/", LogLevel.Info);
```
- in yaml
```yaml
LogLevels:
  ExtendedVariantMode/: Info
  ExtendedVariantMode/ExtendedVariantsModule: Verbose
```
("I want verbose logging for ExtendedVariantsModule, Info logging for the rest of Extended Variants")

Celeste.Mod.mm/Mod/Everest/Logger.cs
```diff
@@ -10,6 +10,70 @@
...
+        private static Dictionary<string, LogLevel> minimumLevelsCache = new Dictionary<string, LogLevel>();
+
+        /// <summary>
+        /// Sets the minimum log level to be written in the logs for lines matching the given tag prefix.
+        /// When using this, make sure the tag prefix is restrictive enough not to impact other mods
+        /// (for example, if all your tags follow the format MyMod/xxx, use "MyMod/" as a prefix).
+        /// </summary>
+        /// <param name="tagPrefix">The prefix of the log tags to affect with this log level</param>
+        /// <param name="minimumLevel">The minimum level of logs to print out in the logs</param>
+        /// (useful to make levels set by code overridable through modsettings-Everest.celeste)</param>
```
coloursofnoise: Please fix the xml tags in this comment, and consider adding some `<br/>` or `<para />` tags to the summary

coloursofnoise (changes requested):

Celeste.Mod.mm/Mod/Everest/Logger.cs
maddie480: Done, this line was actually a leftover from an extra parameter that was removed during development.

coloursofnoise (approved):

PR #225, opened 2020-08-24
Add option for accepting keyboard text input

rhelmot:
This makes more sense than arbitrarily disabling keyboard input if a controller is attached; I play with a controller and I would like to be able to use my keyboard for text input.


rhelmot:
while dogfooding this I noticed that backspace was backspacing two letters at once, and figured out that backspace is hardcoded to be bound to MenuCancel so it will always cause a backspace event anyway, and thus commented out the code that handles the backspace scancode as a backspace. backspace.

rhelmot:
Before this gets merged: I'm seeing more issues in the form of "if a key is bound to menuconfirm, pressing it will input both the keycode and the currently selected letter" which I will try to fix. If you have any advice it would be appreciated...

rhelmot:
Okay, should be fixed and ready to merge. Turns out ignoring code you don't understand is a bad look.

rhelmot:
Actually, one important question: should the default be off or on? I had it set to on because I thought it didn't disable the normal gamepad input method, but it turns out that that needs to be disabled in order for it to work. Should it be off by default to preserve the vanilla behavior?

PR #232, opened 2020-09-07
Add new NonGameplayLazyLoading hidden setting

maddie480:
Some try at a middle ground between "let's load everything" and "load stuff as it's needed, even if it causes stutter during gameplay".

Here is an option, enabled by default, that enables lazy loading for _some_ assets, everything excluding:
- everything from base game
- everything in the Gameplay, Misc and Mountain atlases
- all color grades

This aims at minimizing RAM usage by, for example, not loading all the maps' checkpoint images, while not introducing stutter in gameplay since gameplay assets are not lazy loaded. On my test install with 150 mods, this cut down RAM usage from ~3 to ~2 GB.

Main questions are:
- good idea or bad idea? 🤔 
- enabled or disabled by default?
- anything I overlooked?


maddie480:
Might have unwanted side effects and seems pretty unnecessary (vs just blacklisting mods), closing

PR #234, opened 2020-09-15
Make watchtower can be placed behind Solid.

DemoJameson:
In the vanilla, placing watchtower behind `Solid` would crash the game on `TalkComponentUI.awake()`.

	System.NullReferenceException: 未将对象引用设置到对象的实例。
	   在 Celeste.TalkComponent.TalkComponentUI.Awake(Scene scene)

```csharp
// Lookout
public override void Update() {
	...
	if (talk != null && CollideCheck<Solid>()) {
		Remove(talk);
		talk = null;
	}
    // after patch
	if (false && CollideCheck<Solid>()) {
		Remove(talk);
		talk = null;
	}
}
```

```csharp
// TalkComponentUI
public override void Awake(Scene scene) {
	base.Awake(scene);
	if (Scene.CollideCheck<FakeWall>(Handler.Entity.Position)) {  // <-- Handler.Entity == talk.Entity == null
		alpha = 0f;
	}
}
```

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -313,6 +313,12 @@ namespace MonoMod {
     [MonoModCustomMethodAttribute("PatchOuiChapterPanelRender")]
     class PatchOuiChapterPanelRenderAttribute : Attribute { };
 
+    /// <summary>
+    /// Don't remove TalkComponent even watchtower collide solid, so that watchtower can be hidden behind Solid.
+    /// </summary>
+    [MonoModCustomMethodAttribute("PatchLookoutUpdate")]
+    class PatchLookoutUpdateAttribute : Attribute { };
+
     static class MonoModRules {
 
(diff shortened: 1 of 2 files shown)


maddie480:
It looks like the intent behind the code that is getting removed here is to hide the bubble if the watchtower is inside a solid (for example, an exit block that solidified). Wouldn't patching that out make the talk component appear no matter if the watchtower is in a solid or not?

DemoJameson:
The bubble will always be there, but since `Solid` defaults to a smaller `Depth` value than `Lookout`, the bubble can be hidden, with a few exceptions like `MoveBlock`

maddie480:
I will test it out to be sure, but since the speech bubble is UI, I think it is on a different layer and that it will render on top of "gameplay" without caring about depth. That depth means that the watchtower itself will be behind most solids though.

DemoJameson:
You are right, so we need to hide watchtower's talk UI when it collides with solid.

PR #237, opened 2020-09-22
Destroy static movers on disabled GoldenBlocks

### LINKED ISSUE #236: Spikes don't properly attach to GoldenBlocks
When spikes, or anything is placed on GoldenBlocks, they don't disappear if the player doesn't have a Golden Berry.
Screenshot 2020-09-16 204441
Screenshot 2020-09-16 204446

coloursofnoise:
Also disable them while the block lerps in, then re-enable them.

Closes #236


Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2534,6 +2540,30 @@ bgt.un.s	ldloca.s
             }
         }
 
+        public static void PatchGoldenBlockStaticMovers(MethodDefinition method, CustomAttribute attrib) {
+            if (!method.HasBody)
+                return;
+
+            MethodDefinition m_Platform_DisableStaticMovers = MonoModRule.Modder.Module.GetType("Celeste.Platform").FindMethod($"System.Void DisableStaticMovers()");
+            MethodDefinition m_Platform_DestroyStaticMovers = MonoModRule.Modder.Module.GetType("Celeste.Platform").FindMethod($"System.Void DestroyStaticMovers()");
```
0x0ade: No need for `$"` here, a simple `"` should be enough.

PR #243, opened 2020-10-20
Allow setting music params in sound test

maddie480:
Press Journal (Tab by default on keyboards) to bring up the list of music params for the currently selected event and be able to change them. The changes will apply directly.

If the span between the minimum setting and the maximum one is strictly less than 3, the slider will go by steps of 0.1. Otherwise, it will go by steps of 1, using the IntSliders provided by Everest (meaning holding a direction will make the digits scroll by faster).

This allows to test out music params quickly, to turn on or off layers, and to listen to the different progresses of songs. It even allows to listen to cassette music, if you feel like spamming your Right key with the right timing. 😅 

Demo video here: https://cdn.discordapp.com/attachments/445631337315958796/768187050146005062/Celeste_2020-10-20_20-50-56.mp4

### DIFF
--- a/Celeste.Mod.mm/Content/Dialog/English.txt
+++ b/Celeste.Mod.mm/Content/Dialog/English.txt
@@ -91,7 +91,8 @@
 
 # Sound Test
 	SOUNDTEST_TITLE=	SOUND TEST
-	SOUNDTEST_AUDIOPARAMS_OPEN=	Press Journal to edit Audio Params
+	SOUNDTEST_AUDIOPARAMS_OPEN_1=	Press
+	SOUNDTEST_AUDIOPARAMS_OPEN_2=	to edit Audio Params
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/Content/Dialog/English.txt
```diff
@@ -91,6 +91,8 @@
 
 # Sound Test
 	SOUNDTEST_TITLE=	SOUND TEST
+	SOUNDTEST_AUDIOPARAMS_OPEN=	Press Journal to edit Audio Params
```
0x0ade: It'd be great if `Journal` could be replaced with the currently active Journal key binding, although it'd require splitting the text into two. Maybe an icon + key binding combo like on the map select screen would do the job too?

maddie480:
Here is what the prompt looks like now:

image

PR #250, opened 2020-11-16
Update Chinese font to add missing characters

WEGFan:


maddie480:
The characters added are: 台邀链传卡卸脉

What are those for? Are they used in Everest or the game itself?

maddie480:
They're indeed all used in Everest's `Simplified Chinese.txt`... let me check if anything else is missing

maddie480:
Okay looks good! (Please include a short description next time though)

PR #251, opened 2020-11-16
Use last write time to name log rotation files

WEGFan:
Some file systems have low resolution on last access time (e.g. FATs are 1 day), so if someone starts the game in that file system twice in this range, the game will crash with this exception because a file with same name exists. (it happened to a friend of mine, which I guess he's starting from a USB device or something?)

```
System.IO.IOException: Cannot create a file when that file already exists.

   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.__Error.WinIOError()
   at System.IO.File.InternalMove(String sourceFileName, String destFileName, Boolean checkHost)
   at System.IO.File.Move(String sourceFileName, String destFileName)
   at Celeste.Celeste.Main(String[] args) in D:\a\1\s\Celeste.Mod.mm\Patches\Celeste.cs:line 35
   at Celeste.Mod.BOOT.Main(String[] args) in D:\a\1\s\Celeste.Mod.mm\Mod\Everest\BOOT.cs:line 0
```

And since the `log.txt` can't be moved to `LogHistory`, the game will always crash unless the file with same name is deleted or the last access time of `log.txt` is changed

Snipaste_2020-11-17_01-52-27

But even with last write time, same thing might happen for unknown reason, so I guess adding a number suffix to the backup if same name exists is better, like `log_20201177_022500.txt` and `log_20201177_022500_1.txt`? Or just directly add a timestamp of current time to the end.


0x0ade:
It would be nice if you could please implement your suffix idea as well :)

WEGFan:
Now if a file with same datetime exists, it will add a suffix index to the end of file name (such as `log_20201208_005102_1.txt`). The index is chosen by iterating all **valid** files with same datetime, find the max index among them and plus one. We assume the index of a file without suffix is 0.

For example: (datetime omitted for simplicity)

- If there is `log.txt`, the new file will be `log_1.txt`
- If here are `log_3.txt` and `log_10.txt`, the new file will be `log_11.txt`

A file is valid if:

- It doesn't have a suffix and it's datetime can be parsed as `yyyyMMdd_HHmmss`.
- It has a suffix, the suffix is in the range of `ulong`, and it's datetime can be parsed in the same way as above.

When deleting outdated files, this sorting method is used:

1. Take two files A and B.
2. If A or B or both are invalid, compare them by alphabetical order instead, the smaller will be older.
3. The file with earlier datetime will be older one, if the datetime are same, smaller index will be older one.

The first file after sorting will be the oldest one.

Some test cases, from oldest to newest:

| File name                   | Valid | Date time           | Index |
| --------------------------- | ----- | ------------------- | ----- |
| log_20201207_112233.txt     | True  | 2020-12-07 11:22:33 | 0     |
| log_20201207_112233_.txt    | False | -                   | -     |
| log_20201207_112233_11a.txt | False | -                   | -     |
| log_20201207_112233_2.txt   | True  | 2020-12-07 11:22:33 | 2     |
| log_20209999_999999.txt     | False | -                   | -     |
| log_20221122_112233.txt     | True  | 2022-11-22 11:22:33 | 0     |
| log_20221122_112233_0.txt   | True  | 2022-11-22 11:22:33 | 0     |
| log_20221122_112233_000.txt | True  | 2022-11-22 11:22:33 | 0     |
| log_20221122_112233_1.txt   | True  | 2022-11-22 11:22:33 | 1     |
| log_20221122_112233_001.txt | True  | 2022-11-22 11:22:33 | 1     |
| log_20221122_112233_2.txt   | True  | 2022-11-22 11:22:33 | 2     |
| log_20221122_112233_10.txt  | True  | 2022-11-22 11:22:33 | 10    |
| log_20221122_112233_100.txt | True  | 2022-11-22 11:22:33 | 100   |

PR #254, opened 2020-11-29
Offer extra save file slots so that the last slot is always empty

maddie480:
Essentially Infinite Saves, except integrated in Everest and actually infinite.

This will show as many slots as you have save files + 1, so that you always have a spare save file.

- When entering file select, Everest goes through the Saves folder and gets the highest file slot. Then, it rebuilds the Slots array to give it as many slots as required + 1 before the game fills it, so the game instantiates the slots itself. (No need to refresh the number of slots, since it is rebuilt each time you come back to the overworld, and creating a new save _requires_ you to leave the overworld.)
- When going through the saves, scrolling has been set up if there are more than 3 slots.
- The file select enter and leave animations were patched so that the animations of off-screen slots don't cause delay.

Question: how do we deal with Infinite Saves if it is installed?

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -102,6 +102,8 @@ namespace Celeste.Mod {
                 { "Elemental Chaos", new Version(1, 0, 0, 0) },
                 { "BGswitch", new Version(0, 1, 0, 0) },
 
+                // Infinite Saves 1.0.0 does not work well with the "extra save slots" feature of Everest
+                { "InfiniteSaves", new Version(1, 0, 1) }
             };
 
             /// <summary>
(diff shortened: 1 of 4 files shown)


WEGFan:
Maybe read the settings in Infinite Saves if installed, and take `Math.Max(maxSaveFile + 2, InfiniteSaveModuleSettings.SaveFileCount)`?

- If the setting in Infinite Saves is 10 and we have 5.celeste, we have 10 slots
- If the setting is 10 and we have 9.celeste, we have 11 slots because we want the last slot be empty
- If the setting is 10 and we have 19.celeste, we have 21 slots

JaThePlayer:
I'm personally not against just perma-blacklisting Infinite Saves once this gets merged (I'm the mod dev), since ultimately this is just a better version of that mod.

maddie480:
Added Infinite Saves to the blacklist in the PR.

PR #255, opened 2020-12-01
Fix C-Side button visible in chapter panel if B-Side doesn't have heart and the player got heart in A-Side

### LINKED ISSUE #228: Running C-side opens B-side instead
If your map don't have hearts in B-side, than it is possible to unlock C-side without unlocking B-side, because in general you can gather heart of A-side without gathering cassette.
However, If in that case you try to run that C-side, it will run B-side instead.
Reproduced in that minimal mod
[Testmod.zip](https://github.com/EverestAPI/Everest/files/5144779/Testmod.zip)

WEGFan:
Fixes #228.

Originally whether C-Side should be displayed is by checking `TotalHeartGems` and `MaxHeartGemsExcludingCSides`. In this case, if there is no heart in B-Side and the player collected the heart in A-Side, these two values are equal, so the button displayed on the chapter panel. But the game used selected index while entering a level, so actually B-Side is entered.

This patch changes the way on checking whether C-Side should unlock. A B-Side is counted as completed if it has heart and the player has collected it, or the player completed it without heart by triggering events if it hasn't. If all B-Sides are completed, then C-Sides should be unlocked.

### DIFF
--- a/Celeste.Mod.mm/Patches/SaveData.cs
+++ b/Celeste.Mod.mm/Patches/SaveData.cs
@@ -700,7 +700,23 @@ namespace Celeste {
             get {
                 int offset = AreaOffset;
 
-                if (TotalHeartGems >= MaxHeartGemsExcludingCSides) {
+                bool completedAllBSides = true;
+                for (int i = 0; i <= MaxArea; i++) {
+                    AreaData areaData = AreaData.Areas[offset + i];
+                    if (!areaData.HasMode(AreaMode.BSide)) {
+                        continue;
+                    }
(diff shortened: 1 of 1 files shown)


0x0ade:
I could see this being a breaking change if the C side was intended to be unlocked via B side "secret exit" hearts, but the only way to find out is to listen to user feedback.

monroth:
Well, in my map i planned that B-side and C-side could be entered independedly from each other (B-side if you find cassette, C-side if you find heart). So not a breaking change (as it did not work previously), but it is a fix that does not make it work either :)
But if there is no ways to make it somehow optional, and there is a reason to make it the way you wrote (so it counts B-sides even if it does not have any hearts), than don't mind me, I'll think of a way around it

PR #257, opened 2020-12-06
Delete titlescreen warning about save data loss + create mod save data on startup if Everest was just updated

maddie480:
The first commit is just a revert of #242 to delete the mod save data loss warning when you restart into vanilla.

The second commit makes sure the mod save data backups are created after you update Everest: if you restart into vanilla and play a save you didn't open since when you upgraded Everest, you'll still lose your mod save data. So, we want to back up mod save data from existing files on first startup to make sure it exists if the user directly restarts into vanilla.

### DIFF
--- a/Celeste.Mod.mm/Patches/GameLoader.cs
+++ b/Celeste.Mod.mm/Patches/GameLoader.cs
@@ -164,7 +164,7 @@ namespace Celeste {
                 previousVersion = new Version(0, 0, 0);
             }
 
-            if (previousVersion < new Version(1, 2104, 0)) {
+            if (previousVersion < new Version(1, 2109, 0)) {
                 // user just upgraded: create mod save data backups.
                 // (this is very similar to OverworldLoader.CheckVariantsPostcardAtLaunch)
                 Logger.Log("core", $"User just upgraded from version {previousVersion}: creating mod save data backups.");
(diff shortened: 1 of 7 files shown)


Celeste.Mod.mm/Patches/GameLoader.cs
```diff
@@ -149,5 +151,37 @@ class patch_GameLoader : GameLoader {
...
+
+                for (int i = 0; i < 3; i++) { // only the first 3 saves really matter.
+                    if (!UserIO.Exists(SaveData.GetFilename(i))) {
+                        continue;
+                    }
+                    SaveData saveData = UserIO.Load<SaveData>(SaveData.GetFilename(i), backup: false);
+                    saveData.AfterInitialize();
+                    if (saveData != null) {
+                        UserIO.Save<ModSaveData>(SaveData.GetFilename(saveData.FileSlot) + "-modsavedata", UserIO.Serialize(new ModSaveData(saveData as patch_SaveData)));
+                    }
```
0x0ade: You're calling AfterInitialize on saveData even if it might be null. Also, what speaks against using BeforeSave?

maddie480: Calling BeforeSave() without actually saving felt weird 😕 I just tried calling it, but it crashed because orig_BeforeSave() tried accessing SaveData.Instance.

PR #261, opened 2020-12-14
Some fixes for the 1.3.3.0 update

maddie480:
Done:
- update lib-stripped
- Some "shims" so that mods using those methods don't crash (Everest is made to use those shims so that it still works on 1.3.1.2 since it will run the actual methods if installed on there)
- patches to RidgeGate now take the new "ridge" parameter into account: if "texture" is not specified, it should be used to pick the texture between `objects/ridgeGate` and `objects/farewellGate`.

To do:
- the --dump-all option crashes
- BirdTutorialGui now seems to take ButtonPrompts instead of VirtualButtons, which potentially breaks custom bird tutorials and anything else using those bubbles, so something should be done about that


maddie480:
Seems like everything in here was done in dev already. 😅

PR #272, opened 2021-01-16
Allow duplicate Decal attributes

coloursofnoise:
Changes DecalInfo.CustomProperties from a `Dictionary` to a `List<KeyValuePair>` allowing for the same attribute to be applied more than once to a decal.

Also adds a "light" DecalRegistry attribute to add a `VertexLight` to the decal.


maddie480:
`DecalInfo` is passed as a parameter to the `OnHandleDecalRegistry` Everest event, so I would check for side effects before merging this. I'll check if any mod I have installed uses this event.

DemoJameson:
I'm ready.
https://github.com/EverestAPI/CelesteTAS-EverestInterop/commit/d4b34d172b9b353d6453837c88854b73a8a2329b#diff-9699a4699f631a5aaa3309745e240d6aaf4e80f02fcd30b91d059d950967ecc1R252

maddie480:
No mod I have installed seem to use `OnHandleDecalRegistry` or `CustomAttributes`. I have 2 mods that have custom decal registry attributes (the Spring Collab and Sardine7) but both use `DecalRegistry.AddPropertyHandler`.

PR #277, opened 2021-02-02
Unused code cleanup & Discord Rich Presence toggle

Lollyde:
I implemented the toggle in a way where it *does* show up in the ingame mod menu - it doesnt cause any issues (except for it doesnt actually update the the rich presence text for a while), but I'm not sure if it'd be better suited as an option that *doesnt* show up ingame.

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
@@ -222,6 +222,20 @@ namespace Celeste.Mod.Core {
             }
         }
 
+        private bool disableDiscordRichPresence = false;
+
+        public bool DisableDiscordRichPresence {
+            get => disableDiscordRichPresence;
+            set {
+                if (value) {
+                    Everest.Discord.Disable();
+                } else {
+                    Everest.Discord.Initialize();
+                }
+                disableDiscordRichPresence = value;
+            }
+        }
+
 
         [SettingIgnore]
         public int DebugRCPort { get; set; } = 32270;
--- a/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
@@ -222,17 +222,16 @@ namespace Celeste.Mod.Core {
             }
         }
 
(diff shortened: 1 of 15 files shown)


Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
```diff
@@ -222,6 +222,20 @@ public class CoreModuleSettings : EverestModuleSettings {
...
+        public bool DisableDiscordRichPresence {
+            get => disableDiscordRichPresence;
+            set {
+                if (value) {
+                    Everest.Discord.Disable();
+                } else {
+                    Everest.Discord.Initialize();
+                }
+                disableDiscordRichPresence = value;
+            }
```
0x0ade: There's no protection against this being set to true / false consecutively multiple times. Either implement something in Everest.Discord to take this into account, or (cheaper) `if (value != disableDiscordRichPresence) { ... }`

0x0ade (changes requested):

Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
Lollyde: added a check to skip the initialization/shutdown if Everest.Discord is already initialized/shut down 👍

Lollyde:
actually, this commit breaks initializing Discord Rich Presence somehow - do not merge yet

Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
```diff
@@ -222,17 +222,16 @@ public class CoreModuleSettings : EverestModuleSettings {
...
         }
 
-        private bool disableDiscordRichPresence = false;
-
         public bool DisableDiscordRichPresence {
-            get => disableDiscordRichPresence;
+            get => !Everest.Discord.Initialized;
             set {
+                if (value != Everest.Discord.Initialized)
+                    return;
```
0x0ade, line 229: This check should ideally be moved to `Everest.Discord` as there *could* be f.e. a mod calling `.Initialize()` without first checking `.Initialized` - it is fine as is though.

0x0ade (approved):

PR #282, opened 2021-02-15
Add line editing to the debug console

rhelmot:
- Arrow keys to move the cursor
- Character insert at any point
- Backspace and delete at any point
- Ctrl + arrow keys, backspace, and delete moves an entire word
- Hold-to-repeat works
- Home and End keys move to beginning and end of line


0x0ade:
At this point I'd like if someone replaced the vanilla keys -> character conversion with Everest's TextInput helper 😅

Would you be willing to do that?

rhelmot:
Oh hell yeah

PR #286, opened 2021-02-22
Fix rare missing file slots bug

WEGFan:
NoelFB/Celeste#4

When file slots are missing, the missing slots is not in the scene's entity list, so they are never get updated and rendered, which is why selecting them will cause a soft lock.

image

After logging all `Monocle.Entity.Added()` and `Monocle.EntityList.Add(Monocle.Entity)` calls (log file [here](https://pastebin.com/iB47jjEG), changed to 100 saves for easier reproduce), I found that the missing slot (at line 148) was adding to the scene but didn't get added. It's because that after pressing "CLIMB", the game starts a thread to load the save files, but the thread is also removing and adding entities, so two threads can access to `Monocle.EntityList.entities` at the same time, which can cause issues. (The main thread updates the game every frame, which calls `Monocle.Engine.Update(GameTime)`, which calls `Monocle.Scene.BeforeUpdate()`, which calls `Monocle.EntityList.UpdateLists()`, which actually adds/removes entities that get added/removed during this frame and sorts the entities by depth, thus `Monocle.EntityList.entities` is changed)

The solution is simply moving code that adds and removes entities out of the loading thread, so only one thread is accessing the entity list at the same time. The patched code should be similar to this:

```diff
*** Celeste/OuiFileSelect.cs
@@ -12,12 +12,20 @@ public class OuiFileSelect : Oui
 public override IEnumerator Enter(Oui from) {
     this.SlotSelected = false;
     if (!OuiFileSelect.Loaded) {
+        for (int i = 0; i < this.Slots.Length; i++) {
+            if (this.Slots[i] != null) {
+                base.Scene.Remove(this.Slots[i]);
+            }
+        }
         RunThread.Start(new Action(this.LoadThread), "FILE_LOADING", false);
         float elapsed = 0f;
         while (!OuiFileSelect.Loaded || elapsed < 0.5f) {
             elapsed += Engine.DeltaTime;
             yield return null;
         }
+        for (int i = 0; i < this.Slots.Length; i++) {
+            base.Scene.Add(ouiFileSelectSlot);
+        }
         if (!this.loadedSuccess) {
             FileErrorOverlay error = new FileErrorOverlay(FileErrorOverlay.Error.Load);
             while (error.Open) {
@@ -65,9 +73,6 @@ public class OuiFileSelect : Oui
 private void LoadThread() {
     if (UserIO.Open(UserIO.Mode.Read)) {
         for (int i = 0; i < this.Slots.Length; i++) {
-            if (this.Slots[i] != null) {
-                base.Scene.Remove(this.Slots[i]);
-            }
             OuiFileSelectSlot ouiFileSelectSlot;
             if (!UserIO.Exists(SaveData.GetFilename(i))) {
                 ouiFileSelectSlot = new OuiFileSelectSlot(i, this, false);
@@ -81,7 +86,6 @@ private void LoadThread() {
                 }
             }
             this.Slots[i] = ouiFileSelectSlot;
-            base.Scene.Add(ouiFileSelectSlot);
         }
         UserIO.Close();
         this.loadedSuccess = true;
```

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -363,6 +363,18 @@ namespace MonoMod {
         }
     };
 
+    /// <summary>
+    /// Patches OuiFileSelect.Enter to fix file slots missing bug.
+    /// </summary>
+    [MonoModCustomMethodAttribute("PatchOuiFileSelectEnter")]
+    class PatchOuiFileSelectEnterAttribute : Attribute { }
+
+    /// <summary>
+    /// Patches OuiFileSelect.LoadThread to fix file slots missing bug.
+    /// </summary>
+    [MonoModCustomMethodAttribute("PatchOuiFileSelectLoadThread")]
+    class PatchOuiFileSelectLoadThreadAttribute : Attribute { }
+
     static class MonoModRules {
 
         static bool IsCeleste;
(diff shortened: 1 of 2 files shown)


NoelFB:
Fixed, will be in our next patch. Thanks!
https://github.com/NoelFB/Celeste/issues/4#issuecomment-809641793

PR #293, opened 2021-03-07
Add "Find Mod ButtonBinding" format to the CustomBirdTutorial controlString

coloursofnoise:
Given a controlString like `mod:MaxHelpingHand/ShowHints`, search for the `EverestModule` from the mod named "MaxHelpingHand" (as defined in the everest.yaml) and try to retrieve the `ButtonBinding` property named "ShowHints" located in its settings.


0x0ade:
I'd personally prefer if the syntax could be consistent with the content API's "absolute path" syntax (`{Mod}:/{Name}`), but I could also see this being too confusing. Feel free to discuss this on Discord 😅

Other than that, 👍

PR #294, opened 2021-03-08
Fix check for vanilla behavior in the coroutine delay hackfix

maddie480:
When `ForceDelayedSwap` is null (so in any case except in a coroutine from an old mod that is not a hook), the delay will apply if the old coroutine and the new coroutine are **both** from vanilla (same assembly as Celeste itself), and if they are **not** the wrap coroutine from `CoroutineDelayHackfixHelper` (because hook coroutines from mods pretty much get replaced by this one... and it is declared in Celeste.exe so it is ""vanilla"" 😓)

In order to check for the `Wrap` coroutine, I went for MonoModRule magic that replaces `typeof(CoroutineDelayHackfixHelper)` with `typeof(CoroutineDelayHackfixHelper.<Wrap>d__2)`.

The mod that caused issues hooked `Player.PickupCoroutine` and did `yield return orig(self)`. Before applying this patch, TAS would desync at c-00 (first room where a jelly appears) with this mod installed, and after applying this the TAS reaches the end of Farewell with no issues.

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -375,6 +375,12 @@ namespace MonoMod {
     [MonoModCustomMethodAttribute("PatchOuiFileSelectLoadThread")]
     class PatchOuiFileSelectLoadThreadAttribute : Attribute { }
 
+    /// <summary>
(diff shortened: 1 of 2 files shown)


maddie480:
Closing because it doesn't fix all issues or even creates new ones. :a:

PR #302, opened 2021-05-27
Add mouse button bindings

coloursofnoise:
Add support for binding mouse buttons to inputs.
Mouse buttons are set in the Keyboard config menu, and Keys take priority wherever necessary.

Still missing are icons for the buttons to display in the config menus [[1]](https://github.com/EverestAPI/Everest/blob/f05acb8d90d4738462d1c88f2b26023fe877a44d/Celeste.Mod.mm/Patches/TextMenu.cs#L329)


coloursofnoise:
This PR is on hold pending the merge of 
- #327

WEGFan:
I recently watched [this video](https://youtu.be/d6GtGbI-now) and there's a [button prompt pack](https://thoseawesomeguys.com/prompts/) in it, can it fit our needs for the config menu?

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -339,9 +339,10 @@ class PatchAreaCompleteVersionNumberAndVariantsAttribute : Attribute { };
 
     /// <summary>
     /// Patches {Button,Keyboard}ConfigUI.Update (InputV2) to call a new Reset method instead of the vanilla one.
+    /// Also implments mouse button remapping.
```
Kalobi:
```suggestion
    /// Also implements mouse button remapping.
```

Celeste.Mod.mm/Patches/Settings.cs
```diff
@@ -157,5 +159,45 @@ class patch_Settings : Settings {
...
+        public new static void Initialize() {
+            orig_Initialize();
+            // Load Mouse Button Bindings, which are applied to the already-loaded settings.
+            if (UserIO.Open(UserIO.Mode.Read)) {
+                (UserIO.Load<VanillaMouseBindings>("modsettings-Everest_MouseBindings") ?? new VanillaMouseBindings().Init()).Apply();
+                UserIO.Close();
+            }
+        }
+
+        public void ClearMouseControls() {
```
bigkahuna443, line 173: I feel like this would be a better fit as a method in `VanillaMouseBindings` that you call from here.

Celeste.Mod.mm/Patches/OuiOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using Celeste.Mod;
+using Celeste.Mod.Helpers;
+using MonoMod;
+using System.Collections;
+
+namespace Celeste {
+    class patch_OuiOptions : OuiOptions {
+        public extern IEnumerator orig_Leave(Oui next);
+        public override IEnumerator Leave(Oui next) {
```
bigkahuna443: Re: discussion in Discord - this only gets called from the main menu options. If you want settings to apply after changes from the in-game menu, you'd probably want to add this in `UserIO` instead

Celeste.Mod.mm/Patches/Settings.cs
coloursofnoise: This matches the vanilla `Settings.SetDefault{Keyboard/Button}Controls`, and I would prefer to keep the scope for `VanillaMouseBindings` restricted to a serialization helper.

Celeste.Mod.mm/Patches/UserIO.cs
```diff
@@ -54,6 +55,13 @@ public static string GetSaveFilePath(string name = null)
             if (Saving)
                 QueuedSaves.Enqueue(Tuple.Create(file, settings));
 
+            if (settings && UserIO.Open(UserIO.Mode.Write)) {
+                // VanillaMouseBindings extracts the data from the Settings class
+                byte[] data = UserIO.Serialize(new VanillaMouseBindings().Init());
+                UserIO.Save<VanillaMouseBindings>("modsettings-Everest_MouseBindings", data);
+                UserIO.Close();
+            }
+
```
bigkahuna443: If you save settings here it bypasses the save thread and any error handling. Might make more sense to add the serialization in `SaveRoutine` and do the saving in `orig_SaveThread` next to where it does that for vanilla settings?

coloursofnoise: Serializing and Saving have been separated, with saving happening in the `SaveThread`, and mirrors how Mod SaveData is handled.

bigkahuna443:
Tested new version, binds work as expected, and they save and load correctly from in-game menu and overworld menu.

Re: Discord discussion, I think it would be better if names were changed to better distinguish from EverestModule settings since mouse bindings are more extensions of vanilla settings (and also mod settings are saved independently of everything else).

Alternatively, the mouse bindings could be bundled into the CoreModule settings, but this would require shuffling some logic around.

bigkahuna443 (approved):

PR #316, opened 2021-06-09
Apply DecalRegistry tags to multiple Decals

### LINKED ISSUE #315: Apply DecalRegistry tags to multiple Decals
Allow a single set of [DecalRegistry](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs) properties to be applied to multiple decals in the following ways:
- Ending the `path` attribute with a `/` will cause it to be treated as a folder, and all decals in the folder will be affected
- Ending the `path` attribute with a `*` will act as a wildcard, and match all decals that begin with that path

catapillie:
These changes make apply decal registry tags easier, specifically when more than one decal need the same tags. Could be really handy for people who use a lot of decals that need the same properties.

* Ending the path attribute with a `/` will cause it to be treated as a folder, and all decals in the folder will be affected.
* Ending the path attribute with a `*` will act as a wildcard, and match all decals that begin with that path.

Decals in subfolders, in both cases, are not affected.
This now makes `DecalRegistry.ReadDecalRegistryXml` only return a list of `KeyValuePair`s that contain the path to decals with the corresponding `DecalInfo` data. Those are all sorted so that the path attributes that concern a folder are applied first, then the ones acting as a wildcard, and then the single decal paths. And then they are all actually applied in `DecalRegistry.ApplyDecalRegistry`.

This also makes in-game mod updating reload every decal registry files, so that the decal tags are registered the same way that they are on startup consistently.

_Closes #315._

### DIFF
--- a/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
+++ b/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
@@ -181,36 +181,34 @@ namespace Celeste.Mod {
                     }
 
                     if (decalPath.EndsWith("*") || decalPath.EndsWith("/")) {
+                        // Removing the '/' made the path wrong
                         decalPath = decalPath.TrimEnd('*');
                         int pathLength = decalPath.Length;
 
-                        foreach (string subDecalPath in 
+                        foreach (string subDecalPath in
                             GFX.Game.GetTextures().Keys
                             .Select(str => str.StartsWith("decals/") ? str.Remove(0, 7) : null)
                             .Where(str => str != null && str.StartsWith(decalPath))
                         ) {
                             // Decals in subfolders are considered as unmatched
-                            if (!subDecalPath.Remove(0, pathLength).Contains("/")) {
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -156,11 +156,13 @@ public static class DecalRegistry {
         /// <summary>
         /// Reads a DecalRegistry.xml file's contents
         /// </summary>
-        public static void ReadDecalRegistryXml(string fileContents) {
+        public static List<KeyValuePair<string, DecalInfo>> ReadDecalRegistryXml(string fileContents) {
```
0x0ade, line 159: Personally not a huge fan of changing the method behaviors and signatures in general, as mods could be depending on this method already. In this specific case though, it makes sense, and I'd be willing to hotglue around this with some MonoModLinkFrom if a mod ever crashes from this API change 😅

PR #319, opened 2021-06-13
Improve performance when entering large menus

WEGFan:
### Issue

If players install a lot of mods, the game may freeze for a short time when they enter a menu with a lot of items (e.g. mod options, map search, map list).

### Reason

The game calls `TextMenu.RecalculateSize()` and calculates all items' width and height every time an item is added or removed. It is unnecessary because this happened in one frame and the calculation is only needed after all operations are done.

### Solution

This patch adds a batch mode to the text menu. When it is enabled, recalculating menu size is skipped when adding and removing menu items, which can significantly increase performance when players enter a large menu.

If `Width` / `Height` / `LeftColumnWidth` / `RightColumnWidth` is accessed in batch mode, `TextMenu.RecalculateSize()` will be called to make sure the result is correct. If `TextMenu.RecalculateSize()` is called in batch mode, `recalculatingSizeInBatchMode` will be set to `true` to prevent infinite recursive calls.

The patch also optimizes `Emoji.Apply(string)` by using two pointers to iterate the string only once and replace the emojis, to also improve performance in large menus, since `TextMenu.RecalculateSize()` is called every time the menu renders. Although there aren't many mods that register emojis, CelesteNet used emojis to display avatars, so if someone has been online for hours, about 50 avatars can be registered as emojis, causing lags on menus.

### Result

Suppose that we have a text menu with `n` items, `a` is an array of length `n`, containing each item's text length, and there are `m` emojis registered. The time complexity before and after the patch is:

|        | Best                | Average             | Worst                 |
| ------ | ------------------- | ------------------- | --------------------- |
| Before | `O(n * m * sum(a))` | `O(n * m * sum(a))` | `O(n^2 * m * max(a))` |
| After  | `O(sum(a))`         | `O(sum(a))`         | `O(sum(a))`           |

Call count and time comparison when entering map list with 20 level sets and 400 maps installed, all maps' name is 20 characters long without emojis, and there are 30 emojis registered: (Note that call time is much longer than in real game due to the profiler)

Before:

before

After:

after

`Emoji.Apply()` benchmark result before and after optimizing ([code](https://www.pastery.net/mgajdh/)):

```text
BenchmarkDotNet=v0.10.3.0, OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-7700HQ CPU 2.80GHz, ProcessorCount=8
Frequency=2742190 Hz, Resolution=364.6720 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.8.4010.0
  DefaultJob : Clr 4.0.30319.42000, 64bit RyuJIT-v4.8.4010.0
```

| Method | Emoji Count | Emoji in Text |               Mean |          StdErr |          StdDev | Scaled |
| ------ | ----------: | ------------: | -----------------: | --------------: | --------------: | -----: |
| Before |          10 |            10 |      3,095.7799 ns |      24.8549 ns |      96.2625 ns |   1.00 |
| After  |          10 |            10 |        912.1202 ns |       9.0746 ns |      48.0184 ns |   0.29 |
| Before |          10 |           100 |     27,512.6470 ns |     225.4702 ns |     873.2423 ns |   1.00 |
| After  |          10 |           100 |      9,595.4766 ns |      88.8038 ns |     486.3986 ns |   0.35 |
| Before |          10 |          1000 |    261,457.5667 ns |   2,535.8593 ns |  11,340.7077 ns |   1.00 |
| After  |          10 |          1000 |     95,803.9844 ns |     772.2700 ns |   2,990.9888 ns |   0.37 |
| Before |         100 |            10 |     12,256.1060 ns |      87.8604 ns |     340.2819 ns |   1.00 |
| After  |         100 |            10 |        865.7724 ns |       8.3928 ns |      41.1159 ns |   0.07 |
| Before |         100 |           100 |    190,369.1403 ns |   1,087.8736 ns |   4,213.3164 ns |   1.00 |
| After  |         100 |           100 |      8,615.6526 ns |      82.9249 ns |     388.9523 ns |   0.05 |
| Before |         100 |          1000 |  2,671,179.0389 ns |  31,067.6264 ns | 194,017.2648 ns |   1.00 |
| After  |         100 |          1000 |     97,382.9423 ns |     808.6068 ns |   3,025.5296 ns |   0.04 |
| Before |        1000 |            10 |    105,540.6256 ns |   1,045.1672 ns |   5,628.3978 ns |   1.00 |
| After  |        1000 |            10 |        943.0521 ns |       9.2539 ns |      47.1859 ns |   0.01 |
| Before |        1000 |           100 |    534,928.4297 ns |   4,262.8933 ns |  16,510.1148 ns |   1.00 |
| After  |        1000 |           100 |      9,600.3199 ns |      64.1450 ns |     248.4326 ns |   0.02 |
| Before |        1000 |          1000 | 16,723,940.9119 ns | 161,419.3316 ns | 665,548.9541 ns |   1.00 |
| After  |        1000 |          1000 |     84,667.3785 ns |     827.3982 ns |   3,309.5929 ns |   0.01 |

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Emoji.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Emoji.cs
@@ -163,8 +163,7 @@ namespace Celeste.Mod {
                                 now head and tail are pointing to colons so we need to check if the text inside colons is an emoji name
                             */
                             string name = text.Substring(head + 1, (tail - 1) - (head + 1) + 1);
-                            bool exist = _IDs.TryGetValue(name, out int value);
-                            if (exist) {
+                            if (_IDs.TryGetValue(name, out int value)) {
                                 // if it is, we need to first append the text before emoji
                                 resultBuilder.Append(text, appendStartIndex, (head - 1) - appendStartIndex + 1);
                                 // then append the emoji itself
@@ -193,7 +192,7 @@ namespace Celeste.Mod {
                 /*
                         aaaa:1111:2222:bbbb
                     (6)               H^S  ^T
-                    there are still text left since last append, so we need to append them 
+                    there are still text left since last append, so we need to append them
                 */
                 if (appendStartIndex < text.Length) {
                     resultBuilder.Append(text, appendStartIndex, (text.Length - 1) - appendStartIndex + 1);
--- a/Celeste.Mod.mm/Mod/Everest/Emoji.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Emoji.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Text;
 using System.Xml;
 
 namespace Celeste.Mod {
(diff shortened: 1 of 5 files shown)


Celeste.Mod.mm/Mod/Everest/Emoji.cs
```diff
@@ -146,10 +148,58 @@ public static bool IsMonochrome(char c)
...
+                        if (head >= 0 && text[head] == ':') {
+                            /*
+                                    aaaa:1111:2222:bbbb
+                                (2) ^S  ^H   ^T   ^
+                                (4) ^S       ^H   ^T
+                                now head and tail are pointing to colons so we need to check if the text inside colons is an emoji name
+                            */
+                            string name = text.Substring(head + 1, (tail - 1) - (head + 1) + 1);
+                            bool exist = _IDs.TryGetValue(name, out int value);
+                            if (exist) {
```
coloursofnoise: These lines can be merged without sacrificing readability
```suggestion
                            if (_IDs.TryGetValue(name, out int value)) {
```

Celeste.Mod.mm/Patches/TextMenu.cs
```diff
@@ -15,11 +15,84 @@ public class patch_TextMenu : TextMenu {
...
+            [MonoModReplace]
+            get {
+                if (batchMode && !recalculatingSizeInBatchMode) {
+                    RecalculateSize();
+                }
+                return width;
+            }
+            [MonoModReplace]
+            private set => width = value;
+        }
```
coloursofnoise, line 55: Does replacing this also remove the compiler-generated backing field?

Celeste.Mod.mm/Patches/TextMenu.cs
```diff
@@ -150,6 +227,39 @@ public class patch_TextMenu : TextMenu {
...
+            item.ValueWiggler.UseRawDeltaTime = item.SelectWiggler.UseRawDeltaTime = true;
+            if (Selection == -1) {
+                FirstSelection();
+            }
+            if (!BatchMode) {
+                RecalculateSize();
+            }
+            item.Added();
+            return this;
+        }
```
coloursofnoise, line 247: This could be changed through MonoModRules to avoid re-implementing, but it may not be worth the effort for something this short.

coloursofnoise (approved):
Looks good to me, comments are just nitpicks

Celeste.Mod.mm/Patches/TextMenu.cs
0x0ade: It doesn't - it'll simply go unused, which is acceptable, given how it's inaccessible anyway.

PR #322, opened 2021-06-17
Don't catch OutOfMemoryException when loading maps

coloursofnoise:
When loading maps all exceptions are caught, but if the game has run out of memory then it is in a presumably unrecoverable state. When this happens while loading maps it will continue to try/catch until all maps have been attempted as seen [here](https://cdn.discordapp.com/attachments/514006912115802113/854098111101141002/log.txt).
This patch throws any OOM exceptions encountered when loading maps.

### DIFF
--- a/Celeste.Mod.mm/Patches/MapData.cs
+++ b/Celeste.Mod.mm/Patches/MapData.cs
@@ -96,6 +96,8 @@ namespace Celeste {
 
             } catch (Exception e) {
                 Mod.Logger.Log(LogLevel.Warn, "misc", $"Failed loading MapData {Area}");
+                if (e is OutOfMemoryException)
+                    throw e; // Unrecoverable
                 e.LogDetailed();
             }
         }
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/MapData.cs
```diff
@@ -96,6 +96,8 @@ public patch_MapData(AreaKey area)
 
             } catch (Exception e) {
                 Mod.Logger.Log(LogLevel.Warn, "misc", $"Failed loading MapData {Area}");
+                if (e is OutOfMemoryException)
+                    throw e; // Unrecoverable
```
0x0ade: The `throw;` syntax rethrows the caught exception without altering its stack trace.

Better yet, it should be possible to use [`when` to let the runtime filter the exception before catching it.](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when)

PR #330, opened 2021-06-26
Add Custom Height Display

coloursofnoise:
Needs some cleanup, this is something I wrote up probably close to a year ago and never finished.
I couldn't decide if the trigger should be placed in the room you want the display in, or in the previous room to be displayed on room transition. Probably the first option is the best.

Some initial feedback from @0x0ade:
- Console.WriteLine might be replaced with logger log
- I don't understand why setAudioProgression = !progressAudio just for StepAudioProgression to !setAudioProgression

I don't have a lot of time to work on things at the moment so if someone is interested in finishing this up in the mean time, feel free to comment/assign yourself.

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/CustomHeightDisplay.cs
+++ b/Celeste.Mod.mm/Mod/Entities/CustomHeightDisplay.cs
@@ -5,6 +5,7 @@ using System.Collections;
 
 namespace Celeste.Mod.Entities {
     public class CustomHeightDisplay : Entity {
+
         private bool drawText => ease > 0f && !string.IsNullOrEmpty(Text);
 
         public string Text { get; }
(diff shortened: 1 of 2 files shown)


coloursofnoise:
Should still be thoroughly tested, but works in theory.

Ahorn plugin: 
```jl
module HeightDisplayTrigger

using ..Ahorn, Maple

@mapdef Trigger "everest/CustomHeightDisplayTrigger" Display(
	x::Number,
	y::Number,
	height::Number=Maple.defaultTriggerHeight,
	width::Number=Maple.defaultTriggerWidth,
	vanilla::Bool=false,
	target::Int=0,
	from::Int=0,
	text::String="",
	progressAudio::Bool=false,
	displayOnTransition::Bool=false,
)

const placements = Ahorn.PlacementDict(
	"Height Display Trigger (Everest)" => Ahorn.EntityPlacement(
		Display,
		"rectangle",
	)
)

end
```

PR #337, opened 2021-07-03
Precompute LevelSetStats.MaxArea and AreaOffset instead of computing it on each usage

maddie480:
The goal of this change is to reduce the time it takes to the game to compute the total amount of cassettes the player has, which appeared to take a lot of time on each frame with a lot of mods installed (which I discovered while doing profiling sessions in a map for an unrelated issue). It is done on each frame due to the auto splitter info class.

I preferred doing that over caching cassette counts because this will help with other things, and the chapter list doesn't change (and when it does AfterInitialize is called anyway) contrary to cassette counts that can change if the player changes save files, activates cheat mode or ... collects a cassette.

**This affects AltSidesHelper** as it hooks get_MaxArea so make sure to let Luna know when it's merged.

------

Before this change:

image

The game spent 30.3% of the time being idle, and 17.1% of the time computing cassette count

------

After this change:

image

The game spent 40.9% of the time being idle, and 3% of the time computing cassette count

### DIFF
--- a/Celeste.Mod.mm/Patches/SaveData.cs
+++ b/Celeste.Mod.mm/Patches/SaveData.cs
@@ -7,6 +7,7 @@ using MonoMod;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Xml;
 using System.Xml.Serialization;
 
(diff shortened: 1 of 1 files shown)


coloursofnoise (approved):
Looks good to me :+1: This reminds me that the `SaveData` class needs some documentation... as does a lot of other stuff

PR #341, opened 2021-07-05
Make hearts command better to support custom maps

DemoJameson:
The `mainmenu` command is also added for quick return to the main menu

### DIFF
--- a/Celeste.Mod.mm/Patches/Commands.cs
+++ b/Celeste.Mod.mm/Patches/Commands.cs
@@ -261,12 +261,12 @@ namespace Celeste {
 
             int num = 0;
             foreach (patch_AreaStats areaStats in saveData.Areas_Safe.Cast<patch_AreaStats>().Where(stats => stats.LevelSet == levelSet)) {
-                foreach (AreaModeStats areaStatsMode in areaStats.Modes) {
+                foreach (AreaModeStats areaModeStats in areaStats.Modes) {
                     if (num < amount) {
-                        areaStatsMode.HeartGem = true;
+                        areaModeStats.HeartGem = true;
                         num++;
                     } else {
-                        areaStatsMode.HeartGem = false;
+                        areaModeStats.HeartGem = false;
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/Commands.cs
```diff
@@ -241,5 +241,35 @@ private class MeshData {
         private static float Float(string data) {
             return float.Parse(data, CultureInfo.InvariantCulture);
         }
+
+        [Command("mainmenu", "go to the main menu")]
+        private static void CmdMainMenu() {
+            Engine.Scene = new OverworldLoader(Overworld.StartMode.MainMenu);
```
WEGFan, line 247: Maybe add these before `Engine.Scene` to be consistent with vanilla `CmdOverworld()`? Although I don't know what's the use of it either *:theo:*

```suggestion
            if (SaveData.Instance == null) {
                SaveData.InitializeDebugMode(true);
                ((patch_SaveData) SaveData.Instance).CurrentSession_Safe = new Session(AreaKey.Default);
            }
            Engine.Scene = new OverworldLoader(Overworld.StartMode.MainMenu);
```

DemoJameson: I don't think it's necessary. I was referring to vanilla `CmdTitlescreen()`, which is just one line of code

Celeste.Mod.mm/Patches/Commands.cs
```diff
@@ -241,5 +241,35 @@ private class MeshData {
...
+        }
+
+        [MonoModReplace]
+        [Command("hearts", "gives a certain number of hearts for the specified level set (default all hearts and current level set)")]
```
0x0ade:
```suggestion
        [Command("hearts", "sets the amount of obtained hearts for the specified level set to a given number (default all hearts and current level set)")]
```

0x0ade (changes requested):

0x0ade (approved):

PR #267, opened 2021-01-12
Update MonoModRules to use ILCursor

coloursofnoise:
Update MonoModRules to use ILCursor where possible
- Some methods are either cleaner without, or too monolithic to deal with at the moment

Replace static shim methods with instance counterparts
- The only remaining one is `MapData._Process` which hasn't been converted to using ILCursor 

Clean up resulting code where possible

:warning:  These changes may break some IL hooks in mods

PR #351, opened 2021-07-29
MonoModRules cleanup

WEGFan:
- [X] Remove FNADroid related patches
- [X] Add attribute to `orig_` methods directly when patching them
- [X] Use a util method to find `MoveNext()` when patching routines
- [ ] ~~Correct multiple `IteratorStateMachine` attributes on a single iterator method~~
  It will break a lot of mods, so it's better to only include this after we have a solution to migrate.
- [x] Correct wrong `ILCursor cursor = new ILCursor(new ILContext(method))` usages to `new ILContext(method).Invoke(context => { ILCursor cursor = new ILCursor(context); ... }`
- [x] Add more fail-fast in IL patches
- [x] Correct copy-pasting code/comments without changes
- [x] Move methods and attribute declarations to `Patches` folder if the patch isn't used in multiple places
- [x] Fix ILSpy failed to decompile enumerator after MonoMod patching

May be added in future:

- Detect MonoMod/MonoMod#73 and crash when patching to remove any potential game crashes after someone adds members in the middle of a class

---

Feel free to comment if I missed something or you think there's more to be done.


coloursofnoise:
Another thing to add to the list would be to use `ILLabel`s to make emitting `br_` instructions more linear, f.e. replacing
```cs
Instruction target = cursor.Clone().GotoNext(
   instr => instr.OpCode == OpCodes.Ldarg_0,
   instr => instr.MatchLdfld("Celeste.Player", "StateMachine")
).Next;
cursor.Emit(OpCodes.Beq_S, target);
```
with
```cs
ILLabel label = cursor.DefineLabel();
cursor.Emit(OpCodes.Beq_S, label);
cursor.GotoNext(
   instr => instr.OpCode == OpCodes.Ldarg_0, 
   instr => instr.MatchLdfld("Celeste.Player", "StateMachine")
).MarkLabel(label);
```

WEGFan:
Hmm, I've never know that you could emit a label first then mark label. I always did the reverse way...

WEGFan:
This PR also includes a fix for ILSpy failed to decompile some enumerator methods (not sure if it's "enumerator inside enumerator" methods) after installing Everest.

Take `IEnumerable<T> Celeste.BackdropRenderer.GetEach<T>()` as an example, this is what it looks like before fixing:



and after the fix:



It's because there are operators in `MoveNext()` changed from `call` to `callvirt` after patching, since ILSpy is expecting a `call` [here](https://github.com/icsharpcode/ILSpy/blob/d91732f82a58dde8f47b6503a40db4232b531d86/ICSharpCode.Decompiler/IL/ControlFlow/StateRangeAnalysis.cs#L194), it then failed to decompile the method.

```diff
    .class nested private auto ansi sealed beforefieldinit '<GetEach>d__9`1'<class T> {
        .method private final hidebysig newslot virtual instance bool MoveNext () cil managed {
            .try {
                ...
                // this.<>m__Finally1();
                IL_0083: ldarg.0
-               IL_0084: call instance void class Celeste.BackdropRenderer/'<GetEach>d__9`1'<!T>::'<>m__Finally1'()
+               IL_0084: callvirt instance void class Celeste.BackdropRenderer/'<GetEach>d__9`1'<!T>::'<>m__Finally1'()
                ...
            } fault {
                IL_0099: ldarg.0
-               IL_009a: call instance void class Celeste.BackdropRenderer/'<GetEach>d__9`1'<!T>::System.IDisposable.Dispose()
+               IL_009a: callvirt instance void class Celeste.BackdropRenderer/'<GetEach>d__9`1'<!T>::System.IDisposable.Dispose()
                // }
                IL_009f: endfinally
            }
            ...
        }
    }
```

WEGFan:
I used this python script to check instruction differences before and after the patch. First you need to install ilspycmd by `dotnet tool install ilspycmd -g`, then make two folders `before` and `after` besides the script, put patched Celeste.exe in it, and run the script. It will generate ILs for every class in exe so it's easier to know which are changed when diffing.

```python
import re
import subprocess
from pathlib import Path

GAME_PATH = Path(r'E:\SteamLibrary\steamapps\common\Celeste')
FOLDERS = [
    'before',
    'after',
]


def export_il():
    print('Exporting IL')
    for folder in FOLDERS:
        print(folder)
        for exe in Path(folder).glob('*.exe'):
            print(exe)
            subprocess.run([
                'ilspycmd',
                str(exe),
                '-il',
                '-r', str(GAME_PATH),
                '-o', str(exe.parent)
            ])


def process_il():
    print('Processing IL')
    for folder in FOLDERS:
        print(folder)
        for file in Path(folder).glob('*.il'):
            print(file)
            with open(file, 'r+', encoding='utf-8') as f:
                s = f.read()

                s = re.sub(r'// Method begins at RVA 0x.*$', '// Method begins', s, flags=re.MULTILINE)
                s = s.expandtabs(4)

                f.seek(0)
                f.truncate()
                f.write(s)


def split_il():
    print('Splitting IL')
    for folder in FOLDERS:
        print(folder)
        for file in Path(folder).glob('*.il'):
            print(file)
            with open(file, 'r', encoding='utf-8') as f:
                lines = f.readlines()
            lines_count = len(lines)
            (begin, end) = (0, 0)
            while end < lines_count:
                try:
                    while not lines[begin].startswith('.class'):
                        begin += 1
                    end = begin
                    while not lines[end].startswith('} // end of class'):
                        end += 1
                except IndexError as e:
                    break

                class_name = re.match(r'} // end of class (\S+)' + '\n', lines[end]).group(1)
                (namespace, _, class_name) = class_name.rpartition('.')

                folder = Path(file.with_suffix(''), namespace)
                folder.mkdir(parents=True, exist_ok=True)
                file_name = re.sub(r'\W', '_', class_name)
                with open((folder / file_name).with_suffix('.il'), 'w', encoding='utf-8') as f:
                    f.writelines(lines[begin:end + 1])

                begin = end


export_il()
process_il()
split_il()
```

After the patch only changes to instruction bodies are the ILSpy decompilation fixes, and there are some other MonoMod custom method attributes changes that won't affect the game.

coloursofnoise (approved):
Separating MonoModRules into each patch file makes it more tedious to look through them for reference, but I think the improved association between IL and `patch_` code is worth it.

microlith57:
currently several of the MonoModRules use `for (int instri = 0; instri < instrs.Count; instri++) { ... }` to patch method instructions, but this both less readable and less safe than using an `ILCursor`, both because inserting instructions this way is more likely to lead to infinite loops and other unexpected behaviour, and because these rules do not count the number of matches (leading to the possibility of builds breaking silently).
are there any advantages that the `for` method has that `ILCursor`s do not?
would it be in-scope for this PR to migrate to using `ILCursor`s in these cases?

WEGFan:
@coloursofnoise thought they are either cleaner without using `ILCursor`, or too monolithic to deal in #267, so I didn't change them in this PR.

PR #359, opened 2021-08-10
Attempt to fix InvalidOperationException when reloading maps

### LINKED ISSUE #356: Game Crashing when Mapping
So when mapping with Ahorn via Olympus, when i save my map, my game crashes when i click on the celeste icon again. The thing is, it doesn't happen all the time. Sometimes i can save a lot of times before crashing and sometimes it crashes almost all the time.
I'm pretty sure it started happening this week and i don't know what i did to trigger the crashes.

Here is the error log:
[error_log.txt](https://github.com/EverestAPI/Everest/files/6947533/error_log.txt)

coloursofnoise:
- hopefully fixes #356 

I was unable to reproduce the crash so I just added a lock to `VirtualContent.assets` to hope that fixes it

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/VirtualContent.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/VirtualContent.cs
@@ -73,12 +73,14 @@ namespace Monocle {
         /// Unload all overworld-related content.
         /// </summary>
         public static void UnloadOverworld() {
-            foreach (VirtualAsset asset in assets) {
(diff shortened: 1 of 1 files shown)


0x0ade:
Sadly locking the list in a single method doesn't change the fact that it is being modified on another thread elsewhere *without* the lock 😅 I really wish fixing this was as easy as merging this PR. I'll sadly need to reject this, thanks for the effort though! Maybe a more aggressive locking PR will have better chances at fixing this.

PR #372, opened 2021-08-28
Add lib-stripped artifact to build pipeline

coloursofnoise:
Generates an artifact using `mono-cil-strip` of the Celeste libs post-patch
For use by mods as part of their build pipelines


WEGFan:
Maybe also split the process of patching to an individual one so we can add a patch check for commits across all versions of Celeste? (although only FNA is available since it's on Ubuntu)

coloursofnoise:
> Maybe also split the process of patching to an individual one so we can add a patch check for commits across all versions of Celeste? (although only FNA is available since it's on Ubuntu)

Definitely in favour of this, including moving the patching to happen *before* creating artifacts, since if it fails patching then we probably shouldn't ship the build :sweat_smile: 
For this PR though I'd rather just get the lib-stripped artifact out and usable, and I'll work on a second PR for proper patch checks at a later date.

PR #374, opened 2021-09-03
Patch DeathEffect.Update to fix Percent check

coloursofnoise:
Original code, `Percent` will never actually be greater than 1f:
```cs
if (Percent > 1f)
...
Percent = Calc.Approach(Percent, 1f, Engine.DeltaTime / Duration);
```

Patched code:
```cs
if (Percent >= 1f)
```

Originally added in 7df0542f0676bec58c83dda69c9abd715e2780bd, reverted in 7279867a6cf6d405cdf5732f8553073cd7ce8021 because of a NRE involving `PlayerDeadBody.Render` calling `deathEffect.Render` after the `DeathEffect` had removed itself.

This is fixed by adding a null check to `DeathEffect.Render`. Other fixes are possible, but likely more invasive.

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -394,6 +394,9 @@ namespace MonoMod {
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchMountainRendererUpdate))]
     class PatchMountainRendererUpdate : Attribute { }
 
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchDeathEffectUpdate))]
+    class PatchDeathEffectUpdateAttribute : Attribute { }
+
     static class MonoModRules {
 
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2249,6 +2252,12 @@ bgt.un.s	ldloca.s
                 throw new Exception("MountainRenderer failed to patch key presses for keys: " + pressedKeys.Keys);
         }
 
+        public static void PatchDeathEffectUpdate(ILContext context, CustomAttribute attrib) {
+            ILCursor cursor = new ILCursor(context);
+            cursor.GotoNext(instr => instr.OpCode == OpCodes.Ble_Un_S);
```
0x0ade, line 2257: Any reason why `MatchBleUn` wouldn't do its job here? 😅

PR #395, opened 2021-10-31
Warn if mod assembly doesn't contain an EverestModule

Popax21:
Log a warning if the assembly of a code mod doesn't contain an `EverestModule`. All code mods should have one, and the ones who don't can bypass this warning by creating an `EverestModule` which either does nothing or subclasses `NullModule`. This can help developers catch errors when they either break the module, or it somehow fails to load (in which case it's not returned by `asm.GetTypesSafe()`).

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -576,7 +576,7 @@ namespace Celeste.Mod {
 
                     if (typeof(EverestModule).IsAssignableFrom(type) && !type.IsAbstract) {
                         foundModule = true;
-                        if(!typeof(NullModule).IsAssignableFrom(type)) {
+                        if (!typeof(NullModule).IsAssignableFrom(type)) {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -570,15 +570,23 @@ public static class Loader {
...
                 for (int i = 0; i < types.Length; i++) {
                     Type type = types[i];
 
-                    if (typeof(EverestModule).IsAssignableFrom(type) && !type.IsAbstract && !typeof(NullModule).IsAssignableFrom(type)) {
-                        EverestModule mod = (EverestModule) type.GetConstructor(_EmptyTypeArray).Invoke(_EmptyObjectArray);
-                        mod.Metadata = meta;
-                        mod.Register();
+                    if (typeof(EverestModule).IsAssignableFrom(type) && !type.IsAbstract) {
+                        foundModule = true;
+                        if(!typeof(NullModule).IsAssignableFrom(type)) {
```
0x0ade:
```suggestion
                        if (!typeof(NullModule).IsAssignableFrom(type)) {
```

0x0ade (changes requested):

0x0ade (approved):

PR #400, opened 2021-11-14
Retrieve everest_update.yaml in gzipped form to spare bandwidth

maddie480:
Currently everest_update.yaml is 221.31 KiB / 43.27 KiB gzipped (which is a 80% save), is downloaded ~6000 times a day, and getting it in gzipped form is just about including two extra headers in the request. 😅


0x0ade:
I'm a bit paranoid about this not working fine with MonoKickstart on macOS / Linux - if you've got the time, can you please give it a try? 😅

maddie480:
Just tested it on macOS, it works :smile:

Issue #307, opened 2021-06-07
Replace DiscordRPC with GameSDK


coloursofnoise:
The [Everest DiscordRPC](https://github.com/EverestAPI/Everest/blob/0d098aab868084d5b5fc649fd15a35edd6e21557/Celeste.Mod.mm/Mod/Everest/Everest.Discord.cs) integration is broken on MacOS, and the [DiscordRPC Library](https://github.com/discord/discord-rpc) has been deprecated in favour of [Discord's GameSDK](https://discord.com/developers/docs/game-sdk/sdk-starter-guide).

endorpersand:
I believe I have figured out why Everest's DiscordRPC integration is broken on macOS.

In the current Everest code, the platform is being checked like so:
```cs
if (!string.IsNullOrEmpty(CoreModule.Settings.DiscordLib))
    lib = CoreModule.Settings.DiscordLib;
else if (Environment.OSVersion.Platform == PlatformID.Win32NT)
    lib = "discord-rpc.dll";
else if (Environment.OSVersion.Platform == PlatformID.MacOSX) {
    lib = "libdiscord-rpc.dylib";
    // FIXME: macOS doesn't see libdiscord-rpc.dylib wherever Celeste.exe is.
} else if (Environment.OSVersion.Platform == PlatformID.Unix)
    lib = "libdiscord-rpc.so";
```

However, on macOS, `Environment.OSVersion.Platform` is `Unix`, not `MacOSX`. So, Everest tries to use `libdiscord-rpc.so` rather than `libdiscord-rpc.dylib`, causing rich presence to fail on macOS.

At the moment, a workaround for current macOS users is to open `Saves > modsettings-Everest.celeste` and replace the line `DiscordLib: ''` with `DiscordLib: libdiscord-rpc.dylib`.

A possible permanent solution would be to try `libdiscord-rpc.so` on Unix platforms and then try `libdiscord-rpc.dylib` if `libdiscord-rpc.so` fails.

KTibow:
reposting this since someone from the discord recommended me to
it would be nice if when you folks get around to replacing the discord rich presence component, you could also improve the design a bit, in a couple ways
1. remove the original rich presence (if possible); having 2 rich presences are bad
2. use the image; it makes things feel more aligned, gives an opportunity to add art, and lets you add another text field (see https://discord.com/developers/docs/rich-presence/best-practices#use-all-of-the-fields-where-applicable)
[...]

PR #401, opened 2021-11-15
Fix Discord rich presence on macOS

endorpersand:
This commit doesn't rewrite Everest's DiscordRPC integration with Discord's GameSDK (like #307 requests), but it fixes rich presence not showing up on macOS and shouldn't interfere with rich presence on other platforms (though I haven't tested that).


0x0ade:
Thanks for submitting the PR and reminding me that the "standard" way to check for macOS is broken with MonoKickstart 😅

Can you please give `PlatformHelper.Is` (available in the `MonoMod.Utils` namespace from the .dll of the same name) a try? It should be able to handle the macOS platform check just fine.

endorpersand:
Done. I can also confirm that this commit works on macOS.

0x0ade:
Thank you :)

PR #402, opened 2021-12-03
Improvements for cassette music maps

Vexatos:
This adjusts two things in `CassetteBlockManager`:
- `BeatIndexOffset` can now be set in map metadata instead of being forced to 5 for levels where the cassette track is the main music event (it's set to 5 in `Awake`)
- the `beatIndex` adjustment is no longer hardcoded to 8 but will adjust to always use the second-to-last cassette colour if the beat is within the second half of a swap and the last colour if it's within the first half.

`BeatIndexOffset` should probably default to 0 for all non-vanilla levels. I believe this is what it should be, since it starting at 5 for levels with the cassette track as the main track is somewhat arbitrary (I believe only Farewell uses that in Vanilla). I wonder whether changing this default will break something; Is the offset of 5 beats compensated for within the FMOD project? Cassette swap cycles are usually eight beats, so that doesn't align with 5 at all.

### DIFF
--- a/Celeste.Mod.mm/Mod/Meta/MapMeta.cs
+++ b/Celeste.Mod.mm/Mod/Meta/MapMeta.cs
@@ -527,6 +527,7 @@ namespace Celeste.Mod.Meta {
         public int TicksPerSwap { get; set; } = 2;
         public int Blocks { get; set; } = 2;
         public int BeatsMax { get; set; } = 256;
+        public int BeatIndexOffset { get; set; } = 0;
         public bool OldBehavior { get; set; } = false;
 
         public void Parse(BinaryPacker.Element meta) {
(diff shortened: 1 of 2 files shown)


0x0ade:
It would be great if the Everest-side default value preserves old behavior as not to risk messing with existing maps. Map editors can use a different default value for newly created maps though.

Vexatos:
Right now, the default value is being used _unless_ the map has `CassetteModifier` set in its `.meta.yaml`. Is that enough? If the `CassetteModifier` group is present, checking whether or not to use default might be difficult since it has to happen in each room while the metadata is parsed on map load. Do you have any suggestion for doing that?

0x0ade:
> Right now, the default value is being used _unless_ the map has `CassetteModifier` set in its `.meta.yaml`. Is that enough?

This behavior sounds good enough to me, I simply forgot about how it behaves in the absence of the cassette modifier group 😅

PR #404, opened 2021-12-23
Disable anchors in serialized yaml

WEGFan:
If the object has multiple same values (checked by `Equals()` and `GetHashCode()`), the serializer will use anchors by default to reduce file size. However, those values will share the same reference after deserializing, causing some unexpected behaviors.

```cs
internal class Program {
    public record Point {
        public int X { get; set; }
        public int Y { get; set; }

        public Point() {
        }

        public Point(int x, int y) {
            X = x;
            Y = y;
        }
    }

    public record Node {
        public Point Position { get; set; }
        public int Scale { get; set; }
    }

    public static void Main(string[] args) {
        List<Node> nodes = new List<Node> {
            new Node {Position = new Point(0, 0), Scale = 1},
            new Node {Position = new Point(0, 0), Scale = 1},
            new Node {Position = new Point(0, 0), Scale = 2}
        };

        string yaml = YamlHelper.Serializer.Serialize(nodes);
        // Outputs:
        // - &o0
        //   Position: &o1
        //     X: 0
        //     Y: 0
        //   Scale: 1
        // - *o0
        // - Position: *o1
        //   Scale: 2
        Console.WriteLine(yaml);

        List<Node> result = YamlHelper.Deserializer.Deserialize<List<Node>>(yaml);
        // Outputs: True True
        Console.WriteLine(ReferenceEquals(result[0], result[1]));
        Console.WriteLine(ReferenceEquals(result[0].Position, result[2].Position));
    }
}
```

Suppose the player wants to change the scale of the first node in mod settings, and the code is `result[0].Scale = 2`, then the second node is also changed.

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/YamlHelper.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/YamlHelper.cs
@@ -6,7 +6,7 @@ namespace Celeste.Mod {
     public static class YamlHelper {
 
         public static IDeserializer Deserializer = new DeserializerBuilder().IgnoreUnmatchedProperties().Build();
-        public static ISerializer Serializer = new SerializerBuilder().ConfigureDefaultValuesHandling(DefaultValuesHandling.Preserve).Build();
+        public static ISerializer Serializer = new SerializerBuilder().ConfigureDefaultValuesHandling(DefaultValuesHandling.Preserve).DisableAliases().Build();
 
         /// <summary>
         /// Builds a deserializer that will provide YamlDotNet with the given object instead of creating a new one.


0x0ade:
This seems like a possible YamlDotNet upstream issue to me, which should be using ReferenceEquals instead of Equals and GetHashCode for this kind of thing. Maybe worth noting in https://github.com/aaubry/YamlDotNet/issues/571?

WEGFan:
Actually non-record classes have this issue too:

```cs
public class Point : IEquatable<Point> {
    public int X { get; set; }
    public int Y { get; set; }

    public Point() {}
    public Point(int x, int y) {
        X = x;
        Y = y;
    }

    public bool Equals(Point other) {
        if (ReferenceEquals(null, other)) {
            return false;
        }
        if (ReferenceEquals(this, other)) {
            return true;
        }
        return X == other.X && Y == other.Y;
    }

    public override bool Equals(object obj) {
        if (ReferenceEquals(null, obj)) {
            return false;
        }
        if (ReferenceEquals(this, obj)) {
            return true;
        }
        if (obj.GetType() != this.GetType()) {
            return false;
        }
        return Equals((Point)obj);
    }

    public override int GetHashCode() {
        unchecked {
            return (X * 397) ^ Y;
        }
    }
}

public static void Main(string[] args) {
    List<Point> points = new List<Point> {
        new Point(0, 0),
        new Point(0, 0)
    };
    // - &o0
    //   X: 0
    //   Y: 0
    // - *o0
    YamlHelper.Serializer.Serialize(points);
}
```

It uses a dictionary to store anchors in [code](https://github.com/aaubry/YamlDotNet/blob/41e334c7f14eec1458c3926a576b09df3699ed2c/YamlDotNet/Serialization/ObjectGraphVisitors/AnchorAssigner.cs#L36), and uses the object as a key, so `Equals()` and `GetHashCode()` is used.

According to the [doc](https://github.com/aaubry/YamlDotNet/wiki/Serialization.Serializer#disablealiases):

> When the same object appears multiple times in the serialized graph, the default behaviour is to emit it only once, assigning it an anchor, and use aliases to reference it on its subsequent occurrences.

I assume this is intended to decrease the output size (and maybe to solve cyclic reference issue?), rather than to let the output document have a closer structure with the original object. 😅

0x0ade:
The docs also talk about "duplicate references", which means that they really should be using an equality comparer which uses `ReferenceEquals` 😅

0x0ade:
Given that there haven't been any developments upstream regarding this / the related issue, I'd like to revisit the idea mentioned on (I think it was) Discord before falling back to merging this PR. Would you or anyone else be willing to try implementing a fix that changes the value equality checks to reference equality checks? Or should I attempt it in my spare time?

WEGFan:
I'll try it these days.

WEGFan:
Currently stuck because of aaubry/YamlDotNet#677, if there's still no response then I guess I need to use reflections to solve that. 😅

PR #406, opened 2022-01-07
Adds UpdatePreceder and UpdateFinalizer functionality through List<Action>

vivianlonging:
The main purpose of this is specifically to have a scenario where you can add some function to the end of an Update call from the perspective of EntityList.Update, without having to hook every scenario you want it in.
Example:
```
Glider.Update {
  (A1)
  ... (B1)
  Actor.Update
  ... (C1)
  (D1)
}

Seeker.Update {
  (A2)
  ... (B2)
  Actor.Update
  ... (C2)
  (D2)
}
```
let's say, I'd want my "hook" to work on both Seeker and Glider at the point D. Unless I hook Seeker and Glider's Update functions, this simply will not be able to work, and if I don't know what the entity I want to "hook" _is_, then there's currently no way to do this.
Real world scenarios of this are specifically Flag Toggle Containers and Floaty Containers in EeveeHelper, which get around this by using a second entity that "contains" the information of the entity and changes it over time, which has recently caused some high memory usage (possibly a memory leak? researching atm), and this can also offer a resolution to those bugs.

Current methodology is using two List<Action>'s for Preceding Actions and Proceeding (Finalizing) Actions, since this isn't expected to be a commonplace thing, which might call for the use of implementing UpdateFinalizers and UpdatePreceders with Components (in practice, you could just add the Actions through the components)

The major issue with this currently that I can see is that the features aren't IL optimized at all due to in-patch construction, and given the extreme load scenario this could give, I think implementing these features with IL directly isn't a bad call. Let me know what you think, please.

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -2410,37 +2410,6 @@ namespace MonoMod {
             foreach (TypeDefinition nested in type.NestedTypes)
                 PostProcessType(modder, nested);
         }
-        /*
-        public static void PatchEntityListUpdate(ILContext context, CustomAttribute attrib) {
-            TypeDefinition entityType = MonoModRule.Modder.FindType("Monocle.Entity").Resolve();
-            FieldDefinition entity_UpdateFinalizer = entityType.FindField("UpdateFinalizer");
-            GenericInstanceType action_entity = new GenericInstanceType(MonoModRule.Modder.FindType("System.Action"));
-            action_entity.GenericArguments.Add(entityType);
-            ILCursor cursor = new ILCursor(context);
-            ILLabel iterateLoop = cursor.DefineLabel();
-            cursor.GotoNext(MoveType.Before, instr => instr.MatchBr(out iterateLoop)); //Retrieve loop iterator code (at the end of the loop)
-            cursor.GotoNext(MoveType.After, instr => instr.MatchLdloc(1), instr => instr.MatchCallvirt("Monocle.Entity", "Update")); // Go after entity.Update
-            //entity.UpdateFinalizer?.Invoke();
-            //entity to stack
-            cursor.Emit(OpCodes.Ldloc_1);
-            //entity.UpdateFinalizer to stack
-            cursor.Emit(OpCodes.Ldfld, entity_UpdateFinalizer);
-            //entity.UpdateFinalizer to stack x2
-            cursor.Emit(OpCodes.Dup);
-            //Mark this as a label 
-            ILLabel passCheck = cursor.MarkLabel();
-            //Invoke the UpdateFinalizer
-            cursor.Emit(OpCodes.Ldloc_1);
-            cursor.Emit(OpCodes.Callvirt, action_entity_Invoke);
-            cursor.Emit(OpCodes.Nop); //Necessary according to SharpLab. idfk
-            //Goes behind the passCheck label, after the Dup instruction but before the lable for callvirt. This lets me insert code at this point
-            cursor.GotoLabel(passCheck, MoveType.Before);
-            //takes the latter of the entity.UpdateFinalizer on the stack and compares to null, branch to passCheck if true
-            cursor.Emit(OpCodes.Brtrue, passCheck);
-            //Assuming this false, it pops the latter of the UpdateFinalizers off stack and branches to the iteration loop, the instruction after this point
-            cursor.Emit(OpCodes.Pop);
-            cursor.Emit(OpCodes.Br, iterateLoop);
-        }*/
 
         public static void PatchEntityListUpdate(ILContext context, CustomAttribute attrib) {
             TypeDefinition Entity = MonoModRule.Modder.FindType("Monocle.Entity").Resolve();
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2408,5 +2411,29 @@ bgt.un.s	ldloca.s
...
+            TypeDefinition Entity = MonoModRule.Modder.FindType("Monocle.Entity").Resolve();
+            MethodDefinition entity_UpdatePreceder = Entity.FindMethod("UpdatePreceder");
+            MethodDefinition entity_UpdateFinalizer = Entity.FindMethod("UpdateFinalizer");
+
+            ILCursor cursor = new ILCursor(context);
+            ILLabel activeBranchLabel = cursor.DefineLabel();
+            ILLabel loopBranchLabel = null;
+            if (!cursor.TryGotoNext(MoveType.Before, instr=>instr.MatchBr(out loopBranchLabel))) {
+                throw new Exception("Label not found. This is a temporary error that shouldn't be shown to you unless you know what you're doing.");
+            }
```
0x0ade, line 2424: Why not use a simple `GotoNext` which is already throwing? Also, please add spaces around the `=>`

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -13,5 +15,26 @@ class patch_Entity : Entity {
         internal void DissociateFromScene() {
             Scene = null;
         }
+
+        public List<Action> UpdatePrecederActions;
+        public List<Action> UpdateFinalizerActions;
+
+        //For some reason this compiles to be significantly slower than it could be in theory, most likely to do patch structure in MonoMod.
+        //If you want me to update this to be written in pure IL I can do that as well, just let me know in PR review
```
0x0ade, line 23: Please elaborate in what you mean by those comments. Specifically, I'm interested in how fast it should be in theory and how MonoMod is holding you back here.

Please keep in mind that C# / IL that looks "less optimized" can still be JIT-compiled to native code that's more optimized than the "optimized C#" counterpart in certain instances. Also, please stop hyperfocusing on micro-optimizations in a community-maintained project that you yourself agreed to that it deserves to be rewritten from scratch as there are many more macro-optimizations and larger problems on the table. Sometimes code clarity and having code that's maintainable by as many people as possible is much more important than writing code that's supposedly faster (I'd like to see benchmarks) with a less clear intent behind why it's done that way and that might even end up misleading new programmers stumbling upon such arcane practices.

Lastly, why not use events with the delegate += / -= syntax?

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -13,5 +15,26 @@ class patch_Entity : Entity {
...
+            }
+        }
+
+        internal void UpdateFinalizer() {
+            if (UpdateFinalizerActions?.Count > 0) { 
+                foreach (Action action in UpdateFinalizerActions) {
+                    action?.Invoke();
+                }
+            }
+        }
```
0x0ade, line 38: Please make the first method's spacing after if / foreach consistent with the second method's spacing.

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2408,5 +2411,29 @@ bgt.un.s	ldloca.s
...
+            ILCursor clone = cursor.Clone();
+            clone.GotoNext(MoveType.After, instr => instr.MatchLdloc(1), instr => instr.MatchCallvirt("Monocle.Entity", "Update"));
+            clone.MarkLabel(activeBranchLabel);
+            cursor.GotoNext(MoveType.After, instr => instr.MatchStloc(1));
+            cursor.Emit(OpCodes.Ldloc_1);
+            cursor.Emit(OpCodes.Callvirt, entity_UpdatePreceder);
+            cursor.GotoNext(MoveType.After, instr => instr.MatchLdloc(1), instr => instr.MatchLdfld("Monocle.Entity", "Active"));
+            cursor.Remove();
+            cursor.Emit(OpCodes.Brfalse, activeBranchLabel);
```
0x0ade, line 2433: `cursor` and `clone` are confusing variable names without any context. Please rename, or document the purpose of the clone with a comment so that it can be identified better at a quick glance.

It's apparently used to mark a label, which can be done in one step without the pre-definition of `activeBranchLabel`:
image

Given how it's used, it makes me wonder if juggling the clone around is necessary to begin with.

0x0ade (changes requested):

Celeste.Mod.mm/MonoModRules.cs
vivianlonging: ah I see I forgot to remove the "temporary error", great job Viv

Celeste.Mod.mm/Patches/Monocle/Entity.cs
vivianlonging: I am uncertain whether or not an instance `event` will be slower than this methodology. I had asked about it in #modding-irl with no response so I ran with this originally.

Also my compiles were different in a few cases, some of which had no-ops littered in with no real context, I assume this is a MonoMod thing but my thought process is if this is happening per entity per update, I'd want to optimize this as much as possible.

vivianlonging:
Closing this pull request due to a major functionality rewrite that works better.

PR #407, opened 2022-01-08
Fix for scaling emoji added to fonts.

RedFlames:
So, emoji scaling. I tried turning game assets into Everest Emoji within its font, and noticed that the "xadvance" property does not get set correctly for the emoji characters in the font, when `MTexture emoji.ScaleFix != 1f`. 
This means if I grab a Texture and apply scaling via ScaleFix e.g. `= .5f` the emoji gets rendered at halfs its size, but the following characters in the text still got spaced out by the original size.

Initially my fix was simply:
```c#
xml.SetAttr("xadvance", (int) (emoji.Width * ((patch_MTexture) emoji).ScaleFix));
```

but that felt a bit too much like a hack based on a hack (fix). I'm not 100% what problem ScaleFix solves and if I may use it freely tbh!

So now I have altered:
```c#
public static void Register(string name, MTexture emoji, float? scale = null)
```
with that nullable float, where it internally replaces null with `((patch_MTexture) emoji).ScaleFix)` if `emoji != null`; and `1f` otherwise. (ScaleFix defaults to `1f` maybe probably most likely?) 

A new problem now is that `scale` only applies to "xadvance" for rendering the emoji.

So I have one overload for Emoji.Register:
```c#
public static void Register(string name, MTexture emoji, int targetHeight)
```

**But** this one modifies the MTexture in that it sets 
```c#
((patch_MTexture) emoji).ScaleFix = targetHeight / (float) emoji.Height;
```
i.e. you can specify the height you wish the emoji to render at and it alters the ScaleFix _for you_.
I'm not 100% sure if this is reasonable or not, since altering the MTexture feels wrong. But it would be a useful feature to set the scaling of the Texture correctly. :thinkeline:

Thoughts? :)

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Emoji.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Emoji.cs
@@ -57,11 +57,27 @@ namespace Celeste.Mod {
         /// </summary>
         /// <param name="name">The emoji name.</param>
         /// <param name="emoji">The emoji texture.</param>
-        public static void Register(string name, MTexture emoji) {
+        /// <param name="targetHeight">The height to render this emoji as. Adjusts the MTexture.ScaleFix as side-effect!</param>
+        public static void Register(string name, MTexture emoji, int targetHeight) {
+            if(emoji != null && emoji.Height != targetHeight)
+                ((patch_MTexture) emoji).ScaleFix = targetHeight / (float) emoji.Height;
+            Register(name, emoji);
+        }
+
+        /// <summary>
+        /// Register an emoji.
+        /// </summary>
+        /// <param name="name">The emoji name.</param>
+        /// <param name="emoji">The emoji texture.</param>
+        /// <param name="scale">Scaling factor for the emoji spacing. Defaults to emoji.ScaleFix.</param>
+        public static void Register(string name, MTexture emoji, float? scale = null) {
             if (!Initialized) {
                 Queue.Enqueue(new KeyValuePair<string, MTexture>(name, emoji));
                 return;
             }
+
+            if (scale == null)
+                scale = emoji != null ? ((patch_MTexture) emoji).ScaleFix : 1f;
 
             bool monochrome;
             if (monochrome = name.EndsWith(".m")) {
@@ -75,7 +91,7 @@ namespace Celeste.Mod {
             xml.SetAttr("height", emoji.Height);
             xml.SetAttr("xoffset", 0);
             xml.SetAttr("yoffset", 0);
-            xml.SetAttr("xadvance", emoji.Width);
+            xml.SetAttr("xadvance", (int) (emoji.Width * scale));
 
             int id = _Registered.IndexOf(name);
             if (id < 0) {
--- a/Celeste.Mod.mm/Mod/Everest/Emoji.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Emoji.cs
@@ -62,14 +62,28 @@ namespace Celeste.Mod {
         }
 
         /// <summary>
-        /// Register an emoji.
+        /// Register an emoji with scaling constraints.
         /// </summary>
         /// <param name="name">The emoji name.</param>
         /// <param name="emoji">The emoji texture.</param>
+        /// <param name="targetWidth">The width to render this emoji as. Adjusts the MTexture.ScaleFix as side-effect!</param>
         /// <param name="targetHeight">The height to render this emoji as. Adjusts the MTexture.ScaleFix as side-effect!</param>
-        public static void Register(string name, MTexture emoji, int targetHeight) {
-            if(emoji != null && emoji.Height != targetHeight)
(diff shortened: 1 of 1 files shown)


WEGFan:
There are already existing mods using this method (CelesteNet as far as I know), so changing the signature will cause a crash:

```
System.MissingMethodException: Method not found: 'Void Celeste.Mod.Emoji.Register(System.String, Monocle.MTexture)'.
   at Celeste.Mod.CelesteNet.Client.Components.CelesteNetEmojiComponent.<>c__DisplayClass4_0.<Handle>b__0()
   at Celeste.Mod.CelesteNet.Client.CelesteNetClientContext.Update(GameTime gameTime) in D:\Games\Steam\steamapps\common\Celeste\Mods\CelesteNet\CelesteNet.Client\CelesteNetClientContext.cs:line 98
```

So it's better to add back the original method, and call the new one inside it:

```cs
public static void Register(string name, MTexture emoji) {
    Register(name, emoji, null);
}
```

RedFlames:
Oh dang, I didn't realize that's what would happen, I had rebuilt my work-in-progress Celestenet at the same time which is why everything was fine for me 😅 

Well I scratched the idea of the nullable float then, it was a bit awkward anyway. Tested this with current public Celestenet and my work-in-progress stuff.

One extra consideration might be whether to generate more "default" Emoji within Everest instead of me building these into Celestenet only?

My current slightly unclean draft is something like:
```c#
 protected readonly string[] DefaultEmoji = 
  { 
      "strawberry=i:collectables/strawberry", 
      "heart=i:collectables/heartgem/0/spin", 
      "forsaken_city=i:areas/city",
      "old_site=i:areas/oldsite",
      "celestial_resort=i:areas/resort",
      "golden_ridge=i:areas/cliffside",
      "mirror_temple=i:areas/temple",
      "reflection=i:areas/reflection", 
      "the_summit=i:areas/Summit",
      "core=i:areas/core"
  };

  public override void Init() {
      base.Init();

      RunOnMainThread(() => {
          foreach (string emoji in DefaultEmoji) {

              string[] emoji_parts = emoji.Split('=');

              string emoji_name = emoji_parts.Length > 0 ? emoji_parts[0] : emoji;
              string emoji_icon = emoji_parts.Length > 1 ? emoji_parts[1] : emoji;

              MTexture icon = GhostEmote.GetIcon(emoji_icon, 0.0f);

              if (icon != null && icon.Texture?.Texture_Safe != null) {
                  icon = new(icon.Parent, icon.ClipRect);

                  Emoji.Register(emoji_name, icon, 64);
              }

              if (!Emoji.TryGet(emoji_name, out char _)) {
                  Logger.Log(LogLevel.VVV, "netemoji", $"Could not get {emoji_name} emoji.");
              }
          }
          Emoji.Fill(CelesteNetClientFont.Font);
      });
  }
```
although as you can see I'm also slightly leaning on the GhostEmote class to deal with Atlases for me 😉 

At least in theory really basic emotes could be generated from vanilla icons and portraits, e.g. replacing the current
https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Content/Graphics/Atlases/Gui/emoji/glad.png
by some code.

RedFlames:
I have extended the new function to take an `int targetWidth` along the `targetHeight` and they're both optional now, defaulting to `0` _(which basically means "unrestrained", don't do any scaling around this dimension)_...

My only question is, existing Mods are built against the Everest that only had
```c#
public static void Register(string name, MTexture emoji)
```
which still exists with this signature

but now I also have
```c#
public static void Register(string name, MTexture emoji, int targetWidth = 0, int targetHeight = 0)
```
which has essentially the same signature if it fills in the default values... except in that case as I understand it overloading resolution prefers the original method, and that might also be why I can safely call `Register(name, emoji)` from this new overload?...

My question would be whether this can break anything or is just a _plain bad idea_:tm: or what 😅 

Oh and uhh my testing if my reasoning about the scaling cases was correct was basically just-
image
Showing the values I put into the targetW/H and underlined which ones "won" to scale the icon. 

I also tried values bigger than the original size of the Texture, and it all looked good to me at a glance.
In any case all the default cases and original method signature should be unaffected. 

__Edit:__
The thing I originally needed this for (showing chapter icons in Celestenet player list) has been implemented and merged successfully without using scaled emoji after all. 
So I don't have an immediate need for this fix anymore, until I decide I want to implement proper emoji for Celestenet that get created from vanilla assets. 🤞

coloursofnoise:
> My question would be whether this can break anything

Luckily it should not, once a mod has been compiled it references the method signature in the IL code, which won't break as long as the original method still exists.
Methods with optional arguments don't really exist on an IL level, the compiler inserts the default arguments on the stack before the method is called.

0x0ade:
Looks good to me :) The only nitpick I have is that it would be a bit safer for the "target size" method to create a copy of the MTexture instance to apply the scale on *that*, but that can be handled by the caller as well, and applying the scale onto a shared instance might even be the desired behavior in some weird edge case.

PR #409, opened 2022-01-14
Use MonoModLinkTo for base.base. calls

coloursofnoise:
- part of #351 

Standardizes use of `[MonoModLinkTo]` to call `base.` methods in patch classes, allowing those classes to extend their vanilla counterparts.

`patch_BackgroundTiles` was not changed because doing so would require changing how its constructor is handled.

Other changes:
- Streamlines the `BounceBlock` constructor from `.ctor` -> `orig_ctor` -> `.ctor` to `.ctor` -> `.ctor`
- Removes the unused `RotateSpinner.orig_get_Angle` method

### DIFF
--- a/Celeste.Mod.mm/Patches/BounceBlock.cs
+++ b/Celeste.Mod.mm/Patches/BounceBlock.cs
@@ -5,7 +5,6 @@ using Monocle;
 using MonoMod;
 
 namespace Celeste {
-    // : Solid because base.Added
     class patch_BounceBlock : BounceBlock {
 
         // We're effectively in BounceBlock, but still need to "expose" private fields to our mod.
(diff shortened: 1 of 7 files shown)


Celeste.Mod.mm/Patches/RotateSpinner.cs
```diff
@@ -4,15 +4,18 @@
 using Microsoft.Xna.Framework;
 using Monocle;
 using MonoMod;
+using System;
 
 namespace Celeste {
     // : Entity because there's no original Awake method to hook, thus base.Awake must be Entity::Awake.
```
WEGFan:
```suggestion
```
This can be removed

Celeste.Mod.mm/Patches/BounceBlock.cs
```diff
@@ -6,27 +6,27 @@
 
 namespace Celeste {
     // : Solid because base.Added
```
WEGFan:
```suggestion
```

WEGFan (approved):

PR #412, opened 2022-01-24
Entity PreUpdate and PostUpdate implemented through Components.

vivianlonging:
This acts to replace https://github.com/EverestAPI/Everest/pull/406 by implementing PreUpdate and PostUpdate as interfaces for Components to inherit. This has also made the changes requested in that PR with respect to anything that remains between the two versions.
This has been tested with a test mod and does work. Unknown benchmarking difference compared to the event option.


Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2384,6 +2392,27 @@ bgt.un.s	ldloca.s
...
         }
 
+        public static void PatchEntityListUpdate(ILContext context, CustomAttribute attrib) {
+            TypeDefinition Entity = MonoModRule.Modder.FindType("Monocle.Entity").Resolve();
+            MethodDefinition entity_UpdatePreceder = Entity.FindMethod("UpdatePreceder");
+            MethodDefinition entity_UpdateFinalizer = Entity.FindMethod("UpdateFinalizer");
+
+            ILCursor cursor = new ILCursor(context);
+            ILLabel loopBranchLabel = null;
+            cursor.GotoNext(MoveType.Before, instr => instr.MatchBr(out loopBranchLabel));
```
coloursofnoise: Why is `loopBranchLabel` defined if it doesn't get used? Why not just jump straight to `Entity.Update`?

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2384,6 +2392,27 @@ bgt.un.s	ldloca.s
...
+            cursor.GotoNext(MoveType.After, instr => instr.MatchLdloc(1), instr => instr.MatchCallvirt("Monocle.Entity", "Update"));
+            ILLabel activeBranchLabel = cursor.MarkLabel();
+            cursor.Emit(OpCodes.Ldloc_1);
+            cursor.Emit(OpCodes.Callvirt, entity_UpdateFinalizer);
+            cursor.GotoPrev(MoveType.After, instr => instr.MatchStloc(1));
+            cursor.Emit(OpCodes.Ldloc_1);
+            cursor.Emit(OpCodes.Callvirt, entity_UpdatePreceder);
+            cursor.GotoNext(MoveType.After, instr => instr.MatchLdloc(1), instr => instr.MatchLdfld("Monocle.Entity", "Active"));
+            cursor.Remove();
+            cursor.Emit(OpCodes.Brfalse, activeBranchLabel);
```
coloursofnoise: If Pre/PostUpdate will be called regardless of the entity being `Active`, there should be no need to replace the branch instructions here since it does not change.

coloursofnoise (changes requested):
Rather than an interface, might it be better to just add a `PostUpdate` and `PreUpdate` Component that accepts a delegate? (Or just one component that accepts both, then you could retrieve the list once for PreUpdate and reuse it in PostUpdate?)

Also, should Pre/PostUpdate be called when the entity isn't `Active`? It seems odd to me that they would be called when `Update` isn't.

vivianlonging:
> Also, should Pre/PostUpdate be called when the entity isn't `Active`? It seems odd to me that they would be called when `Update` isn't.
This is to enable certain objects to toggle on and off the object even if Active is already false, see Flag-Toggled Containers from EeveeHelper as an example of the specific use case.

Celeste.Mod.mm/MonoModRules.cs
vivianlonging: I'll test removing it

vivianlonging:
Updated MonoModRules patch as well as replaced interfaces with abstract Component inherited classes
abstract version not tested but given that the method should clearly work I didn't think it prudent to do the test here, given that the same function effectively worked last time. ILSpy code reviewed and does patch as expected.

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -13,5 +14,17 @@ class patch_Entity : Entity {
         internal void DissociateFromScene() {
             Scene = null;
         }
+
+        internal void UpdatePreceder() {
+            foreach (PreUpdateComponent puc in Components) {
```
DemoJameson: Cause 'System.InvalidCastException' when casting from Monocle.Component to Celeste.Mod.Entities.PreUpdateComponent

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -13,5 +14,17 @@ class patch_Entity : Entity {
...
+                    puc.PreUpdate();
+            }
+        }
+
+        internal void UpdateFinalizer() {
+            foreach (PostUpdateComponent puc in Components) {
```
DemoJameson: Cause 'System.InvalidCastException' when casting from Monocle.Component to Celeste.Mod.Entities.PostUpdateComponent

DemoJameson (changes requested):

DemoJameson (changes requested):
`IPreUpdateComponent.cs` should be renamed to `PreUpdateComponent.cs`

DemoJameson:
> Updated MonoModRules patch as well as replaced interfaces with abstract Component inherited classes
> 
> abstract version not tested but given that the method should clearly work I didn't think it prudent to do the test here, given that the same function effectively worked last time. ILSpy code reviewed and does patch as expected.

Because you use two abstract classes, there is no way we can use two methods in the same component at the same time, which I think is unreasonable. You can merge them into one abstract class, but I think it's better to add methods directly in the component

vivianlonging:
> > Updated MonoModRules patch as well as replaced interfaces with abstract Component inherited classes
> > abstract version not tested but given that the method should clearly work I didn't think it prudent to do the test here, given that the same function effectively worked last time. ILSpy code reviewed and does patch as expected.
> 
> Because you use two abstract classes, there is no way we can use two methods in the same component at the same time, which I think is unreasonable. You can merge them into one abstract class, but I think it's better to add methods directly in the component

My point of concern there is more of "if every component is running two extra methods every entity every frame that equates to a lot of lag. Furthermore, if not a lot of Components *need* that behavior then I wouldn't want to slow down the game massively over something that isn't going to be used much compared to not being used, hence the abstract class

I will condense it into a subclass of Component though.

DemoJameson:
then I think two interfaces would be more appropriate than two abstract classes
since a class can implement multiple interfaces but not inherit multiple classes

vivianlonging:
> then I think two interfaces would be more appropriate than two abstract classes
> since a class can implement multiple interfaces but not inherit multiple classes

This waa my original idea but colours recommended otherwise. I can re-implement it though, if thatll move the ball forward on this

DemoJameson (approved):

coloursofnoise (approved):

vivianlonging:
Please dont merge this yet. I'm gonna make one small optimization, changing Action to Action\<Entity\> since storing it as an argument can reduce lag at high numbers of actions

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -13,5 +14,19 @@ class patch_Entity : Entity {
         internal void DissociateFromScene() {
             Scene = null;
         }
+
+        internal void UpdatePreceder() {
+            foreach (Component _component in Components) {
+                if(_component is UpdateWrappingComponent component)
```
0x0ade:
```suggestion
                if (_component is UpdateWrappingComponent component)
```

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2384,6 +2392,24 @@ bgt.un.s	ldloca.s
...
+            MethodDefinition entity_UpdatePreceder = Entity.FindMethod("UpdatePreceder");
+            MethodDefinition entity_UpdateFinalizer = Entity.FindMethod("UpdateFinalizer");
+
+            ILCursor cursor = new ILCursor(context);
+            ILLabel branch = null;
+            cursor.GotoNext(MoveType.After, instr => instr.MatchBr(out branch));
+            cursor.GotoNext(MoveType.After, instr => instr.MatchStloc(1));
+            cursor.Emit(OpCodes.Ldloc_1);
+            cursor.Emit(OpCodes.Callvirt, entity_UpdatePreceder);
+            cursor.GotoNext(MoveType.AfterLabel, instr=>instr.MatchLdloca(0), i2 => i2.MatchCall(out MethodReference method) && method.Name == "MoveNext");
```
0x0ade:
```suggestion
            cursor.GotoNext(MoveType.AfterLabel, instr => instr.MatchLdloca(0), i2 => i2.MatchCall(out MethodReference method) && method.Name == "MoveNext");
```

0x0ade (changes requested):
LGTM besides the small style nitpicks. Can merge as is if necessary and fix afterwards tho.

I'm not sure if I'm a fan of the "preceder" and "finalizer" terminology, but the component fields are perfectly understandable (pre and post), and that terminology is only used behind the scenes anyways.

vivianlonging:
> I'm not sure if I'm a fan of the "preceder" and "finalizer" terminology

I can change that to PreUpdate and PostUpdate since that makes sense in both contexts. EntityList.Update runs Entity.Update, Entity.PreUpdate can run PreUpdate

0x0ade (approved):

PR #415, opened 2022-01-28
Calculate height of subheader for multiline text

coloursofnoise:
Code:
```cs
// CreateModMenuSectionHeader
menu.Add(new TextMenuExt.SubHeaderExt(Dialog.Clean("communalhelper_failedloadingdeps")) {
   TextColor = Color.OrangeRed,
   HeightExtra = 0f, // TODO: fix SubMenu height calculation for multi-line text
});
---
communalhelper_failedloadingdeps=Failed loading optional dependencies.
Send your log.txt to a CommunalHelper maintainer.
```

Before: 
image

After:
image

Could potentially also be implemented for other TextMenu types (`Button`, `Option`, etc..), but I'd expect this would be the most common use-case.

### DIFF
--- a/Celeste.Mod.mm/Patches/TextMenu.cs
+++ b/Celeste.Mod.mm/Patches/TextMenu.cs
@@ -366,6 +366,10 @@ namespace Celeste {
                 ctor(label, true);
             }
 
+            [MonoModReplace]
+            public override float Height() {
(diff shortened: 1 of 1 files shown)


0x0ade:
I'm always a bit paranoid when it comes to text measuring, as it could be that a single-line string ends up having a height that doesn't match `ActiveFont.LineHeight`. Still looks good to me tho, thank you :)

PR #418, opened 2022-02-13
Implement parsing Backdrop Tags as a comma separated list

### LINKED ISSUE #417: [Suggestion] Allow multiple tags to be applied to a Backdrop
currently, `Backdrop` has a `HashSet<string>` of tags, but only one tag can actually be applied from the map bin, because the tags are loaded with
```cs
// 1.4.0.0-fna + 1.3205.0-azure-ec20c
// Celeste.MapData.ParseBackdrop, from IL_0442 to IL_046d
if (child.HasAttr("tag")) {
  backdrop.Tags.Add(child.Attr("tag"));
}
```
which adds at most one value to `backdrop.Tags`.
i'd recommend changing this to something along the lines of
```cs
if (child.HasAttr("tag")) {
  foreach (string tag in child.Attr("tag").Split(',')) {
    backdrop.Tags.Add(tag);
  }
}
```

JaThePlayer:
Closes #417 
Parses Backdrop Tags as a comma separated list. Since `Backdrop.Tags` is `HashSet<string>` already, this is really simple to implement. This is really beneficial for any sort of advanced backdrop manipulation, like what Windowpane Helper does (in fact, that mod manually implements this exact change)

This is the change in `MapData.ParseBackdrop`:
```cs
if (child.HasAttr("tag")) {
    //backdrop.Tags.Add(child.Attr("tag"));
    MapData.ParseTags(child, backdrop);
}
if (above != null && above.HasAttr("tag")) {
    //backdrop.Tags.Add(above.Attr("tag"));
    MapData.ParseTags(above, backdrop);
}
```

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -832,17 +832,21 @@ namespace MonoMod {
             cursor.Emit(OpCodes.Brtrue, branchCustomToSetup);
 
             // Allow multiple comma separated tags
-            OpCode argOpcode = OpCodes.Ldarg_1;
+            int matches = 0;
             while (cursor.TryGotoNext(MoveType.After, instr => instr.MatchLdstr("tag"), instr => instr.MatchCallvirt("Celeste.BinaryPacker/Element", "HasAttr"))) {
                 cursor.Index++; // move past the branch
                 cursor.RemoveRange(8); // remove old code
 
-                cursor.Emit(argOpcode); // child
+                cursor.Emit(matches switch {
+                    0 => OpCodes.Ldarg_1, // child
+                    1 => OpCodes.Ldarg_2, // above
+                    _ => throw new Exception($"Incorrect number of matches for HasAttr(\"tag\"): {matches}")
+                }); // child
                 cursor.Emit(OpCodes.Ldloc_0); // backdrop
 
                 cursor.Emit(OpCodes.Call, m_ParseTags);
 
-                argOpcode = OpCodes.Ldarg_2;
+                matches++;
             }
         }
 
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -840,7 +840,7 @@ namespace MonoMod {
                 cursor.Emit(matches switch {
                     0 => OpCodes.Ldarg_1, // child
                     1 => OpCodes.Ldarg_2, // above
-                    _ => throw new Exception($"Too many of matches for HasAttr(\"tag\"): {matches}")
+                    _ => throw new Exception($"Too many matches for HasAttr(\"tag\"): {matches}")
                 }); // child
                 cursor.Emit(OpCodes.Ldloc_0); // backdrop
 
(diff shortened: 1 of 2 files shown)


WEGFan:
Everest patches should fail hard wherever possible, the code should add checks for instruction matches like this.

```diff
    // Allow multiple comma separated tags
+   int matches = 0;
    OpCode argOpcode = OpCodes.Ldarg_1;
    while (cursor.TryGotoNext(MoveType.After, instr => instr.MatchLdstr("tag"), instr => instr.MatchCallvirt("Celeste.BinaryPacker/Element", "HasAttr"))) {
        cursor.Index++; // move past the branch
        cursor.RemoveRange(8); // remove old code
    
        cursor.Emit(argOpcode); // child
        cursor.Emit(OpCodes.Ldloc_0); // backdrop
    
        cursor.Emit(OpCodes.Call, m_ParseTags);
    
        argOpcode = OpCodes.Ldarg_2;
+       matches++;
    }
+   if (matches != 2) {
+       throw new Exception("Incorrect number of matches for HasAttr(\"tag\")");
+   }
```

WEGFan (changes requested):
Everest patches should fail hard wherever possible, the code should add checks for instruction matches like this.

```diff
    // Allow multiple comma separated tags
+   int matches = 0;
    OpCode argOpcode = OpCodes.Ldarg_1;
    while (cursor.TryGotoNext(MoveType.After, instr => instr.MatchLdstr("tag"), instr => instr.MatchCallvirt("Celeste.BinaryPacker/Element", "HasAttr"))) {
        cursor.Index++; // move past the branch
        cursor.RemoveRange(8); // remove old code
    
        cursor.Emit(argOpcode); // child
        cursor.Emit(OpCodes.Ldloc_0); // backdrop
    
        cursor.Emit(OpCodes.Call, m_ParseTags);
    
        argOpcode = OpCodes.Ldarg_2;
+       matches++;
    }
+   if (matches != 2) {
+       throw new Exception("Incorrect number of matches for HasAttr(\"tag\")");
+   }
```

WEGFan (approved):
Ideally `MapData.ParseTags` should be private but I guess that's not a big problem, and it will require us using reflections if we want to access that in mods 😅

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -829,6 +830,27 @@ we found
...
+            // Allow multiple comma separated tags
+            int matches = 0;
+            while (cursor.TryGotoNext(MoveType.After, instr => instr.MatchLdstr("tag"), instr => instr.MatchCallvirt("Celeste.BinaryPacker/Element", "HasAttr"))) {
+                cursor.Index++; // move past the branch
+                cursor.RemoveRange(8); // remove old code
+
+                cursor.Emit(matches switch {
+                    0 => OpCodes.Ldarg_1, // child
+                    1 => OpCodes.Ldarg_2, // above
+                    _ => throw new Exception($"Too many matches for HasAttr(\"tag\"): {matches}")
```
0x0ade, line 843: `matches` should always be 2 at that point. Having `matches` as part of the exception message feels unnecessary, but isn't critical.

Celeste.Mod.mm/Patches/MapData.cs
```diff
@@ -208,6 +208,11 @@ public patch_MapData(AreaKey area)
             return null;
         }
 
+        public static void ParseTags(BinaryPacker.Element child, Backdrop backdrop) {
```
0x0ade, line 211: I agree with WEGFan that this method should be private, and personally don't see why a mod would ever want to use this. But on the other hand, there should be no harm in keeping this public.

PR #419, opened 2022-02-16
Add contributor documentation

coloursofnoise:
## Contributor Documentation Preview
[CONTRIBUTING.md](https://github.com/coloursofnoise/Everest/blob/contributors/CONTRIBUTING.md#contributing)

## Issue Creation Screen Preview
Issue Creation Screen


WEGFan:
What about using [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms)? I tried it for bug reports template and you can see the template [here](https://github.com/WEGFan/Everest/blob/contributors/.github/ISSUE_TEMPLATE/bug_report.yml) and preview [here](https://github.com/WEGFan/Everest/issues/new?assignees=&labels=bug&template=bug_report.yml).

Also error log file is `error_log.txt` on XNA, and `errorLog.txt` on FNA. The location of it on FNA is also different based on the operating system (see `Monocle.ErrorLog.GetLogPath()` for more info), so I don't know whether we should write down how to find the file and tell users to upload it, or just let them copy the first error log.

WEGFan:
Also maybe add a "if you don't want to upload your map/mod publicly, you can send them in Everest team member DMs" note?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,83 @@
...
+        # Thanks for reporting a bug!
+  - type: markdown
+    attributes:
+      value: '## Please confirm:'
+  - type: input
+    attributes:
+      label: My mods and Everest install are up to date
+      placeholder: yes/no
+    validations:
+      required: true
```
WEGFan:
```suggestion
  - type: dropdown
    attributes:
      label: My mods and Everest install are up to date
      options:
        - 'Yes'
        - 'No'
    validations:
      required: true
```

These can be changed to dropdowns if you want a Y/N option but don't prefer using checkboxes

WEGFan (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
coloursofnoise: Good idea, thanks!

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+
+# Making Changes
+These are explanations and guidelines for how to make changes to the Everest source code.
+
+## Adding Content
+Content added *must not* change the functionality of the base game, or change any public facing API.  
+While many mods also make use of non-public code it is not as important to maintain backwards compatibility for it.
+
+New features should only be added to improve or un-hardcode vanilla features, especially for general use in code mods (f.e. [StrawberryRegistry](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/Registry/StrawberryRegistry.cs), [CustomNPC](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/Entities/CustomNPC.cs), [Custom Events](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Patches/EventTrigger.cs)).  
+Gameplay mechanics, entities, and other features not present in the base game are usually better suited for a mod.
```
0x0ade: What about quality of life improvements (f.e. the return to map hint that later inspired the vanilla implementation, or other QoL changes to Everest itself)? Or features which require a deep integration with the game's code, and thus would be better suited as part of the mod loader, but don't necessarily count as "improving vanilla features" (f.e. the map parenting system)? Maybe it's worth mentioning that exceptions will be made on a case by case basis.

CONTRIBUTING.md
```diff
@@ -0,0 +1,124 @@
...
+```cs
+public extern void orig_DoThing(int val);
+public void DoThing(int val) {
+    int newVal = Transform(val);
+    orig_DoThing(newVal);
+}
+```
+
+There are a few attributes that can be applied to members of the class with different effects:
+- `[MonoModIgnore]` ignore this member, do not patch it.
```
0x0ade: Custom attributes on `MonoModIgnore`d members still get handled appropriately though, making it necessary sometimes for MonoModRules. Maybe worth mentioning here or under the MonoModRules section? 😅

0x0ade (changes requested):

coloursofnoise:
@0x0ade added some qualifiers, I'm hoping that a more in-depth patching guide will be compiled in the future that can explain MonoMod patches more comprehensively.

0x0ade (approved):

PR #422, opened 2022-02-26
Adding KR translations

lDMDiamondl:
I just found that there were no KR translations of the Mod options, so I made a new one. But since I'm using a custom KR translations of the vanilla side, and I'm new to moddings, so I wasn't able to test these translations. I'm not sure if these fonts would be applied well, but I'll just send these files first.


Celeste.Mod.mm/Content/Dialog/Fonts/korean.fnt
```diff
@@ -0,0 +1,28 @@
+<font>
+	<info face="Noto Sans CJK KR Medium" size="64" bold="0" italic="0" charset="" unicode="1" stretchH="100" smooth="1" aa="4" padding="1,1,1,1" spacing="1,1"/>
+	<common lineHeight="65" base="50" scaleW="256" scaleH="256" pages="1" packed="0" alphaChnl="0" redChnl="0" greenChnl="0" blueChnl="0"/>
+	<pages>
+		<page id="0" file="korean.png"/>
```
WEGFan:
```suggestion
		<page id="0" file="korean_everest_0.png"/>
```
This needs to be changed after you renamed the png file.

WEGFan (changes requested):

Celeste.Mod.mm/Content/Dialog/Fonts/korean.fnt
lDMDiamondl: didnt noticed about that, ty

WEGFan (approved):

PR #423, opened 2022-03-08
Added extra mod blacklist commandline argument

Pioooooo:
Allow for setting an extra blacklist file with commandline argument `--blacklist <filename>.`
It works similarly to the current whitelist file as it is ignored when toggeling mods in-game. 
Maybe should change the dialog `MODOPTIONS_MODTOGGLE_WHITELISTWARN` since this dialog is now also shown when an extra blacklist is active.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -34,6 +34,16 @@ namespace Celeste.Mod {
             public static ReadOnlyCollection<string> Blacklist => _Blacklist?.AsReadOnly();
 
             /// <summary>
+            /// The path to the Everest /Mods/temperalblacklist.txt file.
+            /// </summary>
+            public static string PathTemperalBlacklist { get; internal set; }
+            internal static string NameTemperalBlacklist;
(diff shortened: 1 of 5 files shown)


Pioooooo:
just found that the collab maps are not looking good..

Pioooooo:
> just found that the collab maps are not looking good..

It was because the debug build configuration fails the il hook and i suppose it now works fine

0x0ade:
Typo: "Temperal" should probably be "Temporary"

Pioooooo:
temperal tower syndrome 🙃
also sorry for those branch stuff, I tried to rename it on github and it was like...

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -33,6 +33,16 @@ public static class Loader {
             /// </summary>
             public static ReadOnlyCollection<string> Blacklist => _Blacklist?.AsReadOnly();
 
+            /// <summary>
+            /// The path to the Everest /Mods/temperalblacklist.txt file.
```
0x0ade:
```suggestion
            /// The path to the Everest /Mods/temporaryblacklist.txt file.
```

PR #424, opened 2022-03-08
Drastically improve GenerateModAssemblyResolver performance

JaThePlayer:
I just looked into loading times, and realized that the Tracker takes 26 seconds to initialize for me with quite a few mods installed. Weirdly, `FakeAssembly.GetTypesSafe` spent tons of time reading zip files to resolve mod assemblies for no reason:
image

Furthermore, I had lag issues when opening Collab Utils 2 Chapter Panels, and it turns out it was caused by this exact same function.

[Explanation of changes](https://github.com/EverestAPI/Everest/pull/424#issuecomment-1062773514)

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -493,7 +493,7 @@ namespace Celeste.Mod {
                                 if (entryName == meta.DLL)
                                     using (MemoryStream stream = entry.ExtractStream())
                                         asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
-                                else if (entryName.StartsWith(dllDirectory)) // immediately load any other libraries to avoid reading zip files over and over again later
+                                else if (Path.GetDirectoryName(entryName) == dllDirectory) // immediately load any other libraries to avoid reading zip files over and over again later
                                     using (MemoryStream stream = entry.ExtractStream())
                                         Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(entryName), stream);
                             }
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -484,17 +484,18 @@ namespace Celeste.Mod {
                 Assembly asm = null;
                 if (!string.IsNullOrEmpty(meta.PathArchive)) {
                     bool returnEarly = false;
-                    string dllDirectory = Path.GetDirectoryName(meta.DLL);
+                    string dllDirectory = Path.GetDirectoryName(meta.DLL) ?? "";
                     using (ZipFile zip = new ZipFile(meta.PathArchive)) {
                         foreach (ZipEntry entry in zip.Entries) {
                             string entryName = entry.FileName.Replace('\\', '/');
 
                             if (Path.GetExtension(entryName) == ".dll") {
-                                using MemoryStream stream = entry.ExtractStream();
                                 if (entryName == meta.DLL)
-                                    asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
+                                    using (MemoryStream stream = entry.ExtractStream())
+                                        asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
                                 else if (entryName.StartsWith(dllDirectory)) // immediately load any other libraries to avoid reading zip files over and over again later
-                                    Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(entryName), stream);
+                                    using (MemoryStream stream = entry.ExtractStream())
+                                        Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(entryName), stream);
                             }
 
                             if (entryName == "main.lua") {
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -487,9 +487,13 @@ namespace Celeste.Mod {
                     using (ZipFile zip = new ZipFile(meta.PathArchive)) {
                         foreach (ZipEntry entry in zip.Entries) {
                             string entryName = entry.FileName.Replace('\\', '/');
-                            if (entryName == meta.DLL) {
-                                using (MemoryStream stream = entry.ExtractStream())
+
+                            if (Path.GetExtension(entryName) == ".dll") {
+                                using MemoryStream stream = entry.ExtractStream();
+                                if (entryName == meta.DLL)
                                     asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
+                                else // immediately load any other libraries to avoid reading zip files over and over again later
+                                    Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(entryName), stream);
                             }
 
                             if (entryName == "main.lua") {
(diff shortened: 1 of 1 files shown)


DemoJameson:
nice! but `SpeedrunTool` crash the game during launching, seems `DeepCloner.dll` is not loaded
here is the [log.txt](https://github.com/EverestAPI/Everest/files/8210871/log.txt)

DemoJameson:
think we should cache the `ResolveEventHandler` result and return cahced result if the meta and args.Name are the same

JaThePlayer:
Okay so I did more testing, turns out that the performance hit comes from trying to load external assemblies - each time anything tries to resolve that assembly, the `GenerateModAssemblyResolver` will read all zip files of all mods to try to find the assembly. This is especially bad for optional dependencies that aren't loaded (in my case, the majority of those 26 seconds comes from me not having CelesteNet installed and some other mod trying to load it).
The approach I tried in this PR is invalid, something else needs to be done

JaThePlayer:
The new approach is loading all .dll files from zips at the same time as scanning for the main .dll, meaning I could remove zip handling from `GenerateModAssemblyResolver` (which was the main performance bottleneck).
Performance wise, this is around ~48x faster in case of missing assemblies (like not having CelesteNet installed), initializing the Tracker takes ~0.5s now, which is still much better
`SpeedrunTool` now works correctly again, more testing is still needed though in case something else broke.

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -487,9 +487,13 @@ public static class Loader {
...
                             string entryName = entry.FileName.Replace('\\', '/');
-                            if (entryName == meta.DLL) {
-                                using (MemoryStream stream = entry.ExtractStream())
+
+                            if (Path.GetExtension(entryName) == ".dll") {
+                                using MemoryStream stream = entry.ExtractStream();
+                                if (entryName == meta.DLL)
                                     asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
+                                else // immediately load any other libraries to avoid reading zip files over and over again later
+                                    Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(entryName), stream);
```
DemoJameson: I think only dll files in the same directory as meta.DLL need to be loaded

JaThePlayer: Yeah that's more consistent with how it used to work, just implemented this

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -484,12 +484,17 @@ public static class Loader {
...
                         foreach (ZipEntry entry in zip.Entries) {
                             string entryName = entry.FileName.Replace('\\', '/');
-                            if (entryName == meta.DLL) {
-                                using (MemoryStream stream = entry.ExtractStream())
+
+                            if (Path.GetExtension(entryName) == ".dll") {
+                                using MemoryStream stream = entry.ExtractStream();
+                                if (entryName == meta.DLL)
                                     asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
+                                else if (entryName.StartsWith(dllDirectory)) // immediately load any other libraries to avoid reading zip files over and over again later
```
DemoJameson:
```suggestion
                                else if (!string.IsNullOrEmpty(dllDirectory) && entryName.StartsWith(dllDirectory)) // immediately load any other libraries to avoid reading zip files over and over again later
```

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -484,12 +484,17 @@ public static class Loader {
...
                     bool returnEarly = false;
+                    string dllDirectory = Path.GetDirectoryName(meta.DLL);
                     using (ZipFile zip = new ZipFile(meta.PathArchive)) {
                         foreach (ZipEntry entry in zip.Entries) {
                             string entryName = entry.FileName.Replace('\\', '/');
-                            if (entryName == meta.DLL) {
-                                using (MemoryStream stream = entry.ExtractStream())
+
+                            if (Path.GetExtension(entryName) == ".dll") {
+                                using MemoryStream stream = entry.ExtractStream();
```
DemoJameson: perhaps move `using MemoryStream stream = entry.ExtractStream();` into if and else if branch

DemoJameson (changes requested):

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
JaThePlayer: I'll just do `dllDirectory = Path.GetDirectoryName(meta.DLL) ?? "";`, accomplishes the same thing

DemoJameson: no, if `dllDirectory` is empty string, then `entryName.StartsWith(dllDirectory)` will always return true

DemoJameson: no, if `dllDirectory` is empty string, then `entryName.StartsWith(dllDirectory)` will always return true

JaThePlayer: Actually yeah, in that case my approach is better - if your mod dll is in the top folder of your mod (`dllDirectory = ""`, happens quite often), then libraries should also be loadable from the top folder, which that `!NullOrEmpty` check would prevent

DemoJameson: okay, but now there's another question, `entryName.StartsWith(dllDirectory)` would return true if  entryName in meta.DLL's subfolder
`!string.IsNullOrEmpty(meta.DLL) && Path.GetDirectoryName(meta.DLL) == Path.GetDirectoryName(entryName)` should work

WEGFan: Comparing `Folder/A.dll` and `Folder2/B.dll` will give wrong result, it should be false instead.

Also, shouldn't this be `Path.GetDirectoryName(meta.DLL) == Path.GetDirectoryName(entryName)`? I think assemblies in same folder can only be referenced.

JaThePlayer: I was using `StartsWith` to support subfolders, but library dlls are always in the same directory anyway so yeah that makes more sense.

DemoJameson (approved):

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -484,12 +484,18 @@ public static class Loader {
...
-                                using (MemoryStream stream = entry.ExtractStream())
-                                    asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
+
+                            if (Path.GetExtension(entryName) == ".dll") {
+                                if (entryName == meta.DLL)
+                                    using (MemoryStream stream = entry.ExtractStream())
+                                        asm = Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(meta.DLL), stream);
+                                else if (Path.GetDirectoryName(entryName) == dllDirectory) // immediately load any other libraries to avoid reading zip files over and over again later
+                                    using (MemoryStream stream = entry.ExtractStream())
+                                        Relinker.GetRelinkedAssembly(meta, Path.GetFileNameWithoutExtension(entryName), stream);
```
0x0ade, line 498: I'm a bit paranoid about this conflicting with mod load order, multiple copies of the same dependency now being differently preferred, and "accidental" dependencies being forcibly loaded. Yet on the other hand, the existing code is horrible, thus if this change doesn't break any mods, I'm willing to accept it.

0x0ade:
My 2ct: instead of loading all .zips and crawling them completely every time a new dependency needs to be loaded, this could probably benefit heavily from the virtual content "filesystem" by treating .dlls as another type of mod asset. Only one dependency resolver would need to be added for the entire content system.

bigkahuna443:
Will be implemented in #469.

PR #425, opened 2022-03-11
Viv pre and post update

vivianlonging:
We're trying this again lmao.
Apparently the UpdateWrappingComponent was hard lagging the game, while this solution doesn't. All of this works, it's just the old event method again.


DemoJameson:
LGTM

now
image
before
image

0x0ade:
Thank you for finally realizing that your "major functionality rewrite that works better" did not, in fact, work better than the simple 
and stupid solution I had already proposed back in January.

PR #426, opened 2022-03-13
Fix crash when exiting feather state while inside a solid

Kalobi:
When the player exits the StStarFly (feather) state while inside a solid, the vanilla game deliberately throws an exception in the Player.StarFlyReturnToNormalHitbox() method, crashing the game. This PR replaces that exception throw with killing the player, which is the expected behavior for ending up inside a solid.

To test: place a solid two tiles above the ground and a feather under it, then jump into the feather while crouched, causing the player to uncrouch into the ceiling. When the feather ends, the game will now kill the player instead of crashing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -252,3 +252,6 @@ result
 
 # Olympus temporary build files
 tmp-olympus/
+
+# JetBrains Rider files
+.idea/
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -421,6 +421,12 @@ namespace MonoMod {
 
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchEntityListUpdate))]
     class PatchEntityListUpdateAttribute : Attribute { }
+    
+    /// <summary>
+    /// Patches the method to kill the player instead of crashing when exiting a feather in a solid.
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchPlayerStarFlyReturnToNormalHitbox))]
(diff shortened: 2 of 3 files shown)


DemoJameson:
perhaps check invincible before invoke die()

WEGFan:
Reproduce video:

https://user-images.githubusercontent.com/7558201/158045510-c0777c49-8117-4bf5-8e65-673769c4b06b.mp4

Kalobi:
@DemoJameson What do you think the behavior when invincible should be? Because right now Die() is being called with the default option evenIfInvincible = false, so it doesn't kill the player if invincible.

Kalobi:
Actually while looking at Die() again, I just realized that the Everest event Everest.Events.Player.Die is being invoked even if the Die function doesn't actually kill the player, for example due to invincibility. Is that intended behavior?

DemoJameson:
> @DemoJameson What do you think the behavior when invincible should be? Because right now Die() is being called with the default option evenIfInvincible = false, so it doesn't kill the player if invincible.

I misremembered
I thought the die method did not handle invincibility

0x0ade:
> Actually while looking at Die() again, I just realized that the Everest event Everest.Events.Player.Die is being invoked even if the Die function doesn't actually kill the player, for example due to invincibility. Is that intended behavior?

Apparently this went unnoticed for over 4 years now. Luckily it's not the only way to listen for player deaths.

> Because right now Die() is being called with the default option evenIfInvincible = false, so it doesn't kill the player if invincible.

I don't think it makes much sense for invincible players to be stuck like that. IIRC in other situations where an invincible player would get squished to death by a block, it would temporarily become non-solid. Yet I can't think of a better way to handle this than "try to push the player out harder".

Issue #430, opened 2022-04-10
multi options selected in main menu


genufish:
### My mods and Everest install are up to date

Yes

### I have recreated the bug with only Everest OR a minimum number of mods enabled

Yes

### Describe the bug

Add multiple mods to the mod directory at once while celeste is running ,mainMenu cant work.

### Steps to reproduce

![Peek 2022-04-10 23-15](https://user-images.githubusercontent.com/67955801/162626651-5dd3a2ff-3ce1-477c-bd1c-56e67254cfe9.gif)

### Expected behavior

Menu can work and only one option is selected

### Operating System

ubuntu 20.04

### Everest Version

3282

genufish:
i fix it by mod code
>     On.Celeste.OuiMainMenu.RebuildMainAndTitle += menuPatch;
>     private static void menuPatch(On.Celeste.OuiMainMenu.orig_RebuildMainAndTitle orig, OuiMainMenu self){
>             string selected = null;
>             foreach (MenuButton button in self.Buttons) {
>                 if (button.Selected && !(button is MainMenuClimb))
>                     selected = (button as MainMenuSmallButton)?.GetLabelName();;
>                 button.SetSelected(false);
>             }
>             orig(self);
>             if (selected == null) {
>                 foreach (MenuButton button in self.Buttons) {
>                     button.SetSelected(button is MainMenuClimb);
>                 }
>             } else {
>                 foreach (MenuButton button in self.Buttons) {
>                     button.SetSelected((button as MainMenuSmallButton)?.GetLabelName() == selected);
>                 }
>             }
>         }

DashingCat:
This issue was resolved by https://github.com/EverestAPI/Everest/pull/845.

PR #431, opened 2022-04-10
Update OuiMainMenu.cs

genufish:
The GC cannot free objects that are referenced.
Add multiple mods to the mod directory at once ,mainMenu cant work.


genufish:
report bug https://github.com/EverestAPI/Everest/issues/430

PR #441, opened 2022-04-25
Fix badeline chaser stuttering after unpausing

DemoJameson:
https://user-images.githubusercontent.com/181192/165041221-1bac352a-3324-4928-bce3-1d7c67640605.mp4


Kalobi:
Should this be restricted to modded maps since it changes vanilla behavior?

DemoJameson:
> Should this be restricted to modded maps since it changes vanilla behavior?

Since it is a regression bug, I think it would be nice to fix it on the vanillla maps

coloursofnoise (approved):
Might be good to add a comment about it being a vanilla bug, or referencing this issue, but otherwise looks good to me :+1:

Issue #320, opened 2021-06-14
Rework Mod Toggling System


coloursofnoise:
Toggling mods is a bit of a mess of disconnected features that needs to be redone in a more cohesive way with a clear focus.

Features that should be included:
- An easy, GUI way to enable/disable individual mods (currently implemented as the Enable/Disable Mods Menu)
- A Search or Filter function, possibly also including dependencies & dependents of returned mods (below the actual search results)
- Toggling should respect dependencies, where enabling a mod enables all required dependencies. (could this be done recursively, and without restarting if possible?)
- Disabling could optionally also respect dependencies, and have a way to disable all dependencies *only* used by a particular mod
- There should be a way to "protect" some mods from being disabled by non-specific disabling functions (e.g. Disable All)
- Whitelists could be implemented through the addition of mod "presets" that will enable a set of mods when toggled

The format these are saved in should still be easy to edit by hand, as well as being usable by Olympus in the future.

DriesMoo:
It would be handy if the Whitelist Overrides the blacklist. This could help with disabling mods in ahorn (recently added function) but still being able to have them open in Celeste. Also making the Whitelist a notepad-only operation will prevent a mod being active in Ahorn being non active in Everst and not being able to load

genufish:
auto load the dependency mod in the blacklist instead of enable it by user

mayromr:
I started experming with implementing a search function to the Enable/Disable Mods Menu.
But I am missing some design guildlines for the feature, I need to know:

- Where will the search bar reside? (in the same menu as the mod list, overlay, diffrent menu page like the level search)
- Where should we display the results? 
- What keybindings should be used?
- Should we limit input size like elave search or allow overflow in some way?

PR #442, opened 2022-04-28
Alter whitelist behaviour to override blacklist (in some or all cases)

caeyo:
Previously, if a mod was disabled in the base blacklist (or temporary blacklist I guess), then even if a provided whitelist explicitly allowed it to load, `ShouldLoadFile()` still returned false because it had to be enabled in both the blacklist _and_ whitelist in order for it to return true. From a brief discussion (plus [this comment](https://github.com/EverestAPI/Everest/issues/320#issuecomment-890322061) and other mentions of it on Discord), it was agreed that this didn't align with expectations for general whitelist behaviour and for Everest in particular (since you need to go out of your way to create a whitelist txt file and enable it via launch args). So this PR alters the loading behaviour to always enable a mod if it is present in a given whitelist.

This also adds an "advanced" setting (i.e. only accessible via `modsettings-Everest.celeste`) to override the blacklist if it enables a particular mod, but that mod is implicitly disabled in a given whitelist (i.e. not present in the file). Basically if a whitelist is provided under this setting, ignore the blacklist(s) entirely, and only use the whitelist to determine what mods should load.
Consensus wasn't as clear on whether this extension of the whitelist behaviour from the first change was ideal, so implementing it via a setting seemed like the most reasonable option. I've defaulted it to false for now so the full override is opt-in, but that should(?) easily be changed if opinions change.

Also cleaned up comments/whitespace at the beginning of the file.

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
@@ -206,6 +206,10 @@ namespace Celeste.Mod.Core {
         [SettingIgnore] // TODO: Show as advanced setting.
         public bool? SaveDataFlush { get; set; } = null;
 
+        [SettingInGame(false)]
+        [SettingIgnore] // TODO: Show as advanced setting.
+        public bool? WhitelistFullOverride { get; set; } = null;
+
         public string InputGui { get; set; } = "";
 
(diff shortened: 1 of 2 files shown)


coloursofnoise (approved):
Looks good to me, and I agree with the reasoning for the behaviour :+1:

PR #445, opened 2022-05-11
Fix tile entity SoundIndex issues

### LINKED ISSUE #311: Add Fallback SoundSurfaceIndex for Tile Entities
`SurfaceIndex.TileToIndex[tile]` crashes with a KeyNotFound exception when given a tileset without a sound attribute.
This could be fixed by providing a default sound id (`0`) to all tilesets missing a sound attribute.
Relevant code: [Patches/Autotiler.cs](https://github.com/EverestAPI/Everest/blob/0d098aab868084d5b5fc649fd15a35edd6e21557/Celeste.Mod.mm/Patches/Autotiler.cs#L72-L74)

Additionally, I noticed that IntroCrushers do not respect tileset sounds properly, which could be added in [Patches/IntroCrusher.cs](https://github.com/EverestAPI/Everest/blob/0d098aab868084d5b5fc649fd15a35edd6e21557/Celeste.Mod.mm/Patches/IntroCrusher.cs#L35-L52)

caeyo:
Closes #311 

Went with `8` (in FMOD this is the `brick` sound) as the fallback SurfaceIndex instead of `0` (no sound) since that seems to be vanilla's default index, and colours and I agreed that it was appropriate to match that.

Needed the `SurfaceIndex.TileToIndex.ContainsKey` check to avoid overwriting the hardcoded vanilla values, as I think the vast majority of existing maps include the vanilla tileset definitions in their XMLs (since the recommended practice is to copy-paste the vanilla ForegroundTiles.xml into your mod) and vanilla tilesets don't define a `sound` attribute. 
Wasn't sure if there was a faster or cleaner way to do this - didn't have access to `TryAdd`, didn't want to use `Add` for fear of some epic exception being thrown, and couldn't find anything else that would easily slot in here. Let me know if there is.

### DIFF
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -69,8 +69,8 @@ namespace Celeste {
                 ReadIntoCustomTemplate(data, tileset, xml);
             }
 
-            if (!SurfaceIndex.TileToIndex.ContainsKey(xml.AttrChar("id"))) {
-                SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.HasAttr("sound") ? xml.AttrInt("sound") : 8; // '8' as fallback instead of '0' to match vanilla's default index
+            if (!SurfaceIndex.TileToIndex.ContainsKey(xml.AttrChar("id")) || xml.AttrChar("id") == 'o') { // Backwards-compat: some existing mods (e.g. Into the Jungle) use 'o' and overwrite its sound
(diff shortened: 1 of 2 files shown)


caeyo:
Minor update: figured out through some testing that some existing mods (e.g. Into the Jungle) were overwriting the index assigned to 'o' in [LevelLoader.cs](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Patches/LevelLoader.cs#L66) - apparently this id goes unused in the vanilla ForegroundTiles.xml so it seemed like it was free to use, despite the [wiki](https://github.com/EverestAPI/Resources/wiki/Custom-Tilesets#step-3---adding-a-tileset-to-the-xml) saying otherwise. Not allowing them to do this causes more issues than it fixes, so added a special-case workaround to maintain backwards-compat

PR #446, opened 2022-05-15
Add lightOcclude property to DecalRegistry

### LINKED ISSUE #291: Addition of a LightOccluder property to the Decal Registry
Hello! I was wondering if it would be possible to add Everest support for a LightOccluder component to the decal registry. 
It's currently possible to add a solid hitbox to a decal, but the lack of an option for light occlusion makes it not as viable for a lot of decals.
alright, thanks for your time!

caeyo:
Closes #291 

LightOcclude already accounts for the decal/entity's position when setting its bounds if you give it a Rectangle, so didn't need to do it when initially setting the Rectangle's dimensions. Can confirm that the bounds of the occluder will line up if the same values for e.g. a `solid` property's dimensions are used, which should satisfy the primary use-case expressed in the request.

I will update the [documentation](https://github.com/EverestAPI/Resources/wiki/Decal-Registry) with this new property and its attributes when this gets merged (feel free to yell at me if I forget)

### DIFF
--- a/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
+++ b/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
@@ -140,6 +140,14 @@ namespace Celeste.Mod {
                 int endFade = attrs["endFade"] != null ? int.Parse(attrs["endFade"].Value) : 24;
                 decal.Add(new VertexLight(offset, color, alpha, startFade, endFade));
             }},
+            { "lightOcclude", delegate(Decal decal, XmlAttributeCollection attrs) {
+                int x = attrs["x"] != null ? int.Parse(attrs["x"].Value) : 0;
+                int y = attrs["y"] != null ? int.Parse(attrs["y"].Value) : 0;
+                int width = attrs["width"] != null ? int.Parse(attrs["width"].Value) : 16;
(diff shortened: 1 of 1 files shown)


coloursofnoise:
Should this be implemented as an option for the existing `solid` property instead?  
Is there a use-case for a decal that is a lightoccluder but not solid?

caeyo:
> Should this be implemented as an option for the existing `solid` property instead?
> Is there a use-case for a decal that is a lightoccluder but not solid?

I thought about that, and while I couldn’t think of any concrete examples, my conclusion was that the flexibility gained in having it separate is better overall - not just for having a lightoccluder without solid properties, but also for specifying different dimensions for the lightoccluder and solid. There are some vanilla objects (`Door` and `Trapdoor`) that don’t have solid collision but do add lightoccluders to feign being solid before passing through, though there are other things that go into those entities to make them work that don’t directly translate to decals. 
 
We can always add a `lightOcclude=“(float)alpha”` option to the solid attribute as well (using the solid’s dimensions) if you think there’s enough benefit in that.

coloursofnoise:
Seems fine to me then, I don't think there's a need for both currently but if someone decides they want to have an attribute as well then it can be added at a later date.

PR #448, opened 2022-06-05
Make "solid" and "staticMover" Decal Registry attributes compatible

bigkahuna443:
- Updates staticMover commands to update the added solid as well
- Patches MoveBlock to disable static movers before resetting their position. Non-solid movers behave the same either way, but moving a solid static mover while it's still collidable teleports any actors on it

I checked other solids as well and didn't find any major compatibility issues besides the usual funkiness from combing two solids (e.g. Kevins getting stuck).

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -428,6 +428,11 @@ namespace MonoMod {
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchPlayerStarFlyReturnToNormalHitbox))]
     class PatchPlayerStarFlyReturnToNormalHitboxAttribute : Attribute { }
 
+    /// <summary>
+    /// Patches MoveBlock.Controller to disable static movers before resetting their position when breaking.
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchMoveBlockController))]
+    class PatchMoveBlockControllerAttribute : Attribute { }
 
     static class MonoModRules {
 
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -428,12 +428,16 @@ class PatchEntityListUpdateAttribute : Attribute { }
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchPlayerStarFlyReturnToNormalHitbox))]
     class PatchPlayerStarFlyReturnToNormalHitboxAttribute : Attribute { }
 
-    /// <summary>
```
coloursofnoise: ?

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -428,12 +428,16 @@ class PatchEntityListUpdateAttribute : Attribute { }
...
     /// Patches the method to update the Name and the TheoSisterName, if the file has been renamed in-game, in the file's SaveData.
     /// </summary>
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchOuiFileSelectSlotOnContinueSelected))]
     class PatchOuiFileSelectSlotOnContinueSelectedAttribute : Attribute { }
-
+
```
coloursofnoise: Whitespace :despair:

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2541,6 +2545,29 @@ bgt.un.s	ldloca.s
             // Target for if renamed is false
             cursor.MarkLabel(renamedTarget);
         }
+        
+        public static void PatchMoveBlockController(MethodDefinition method, CustomAttribute attrib) {
+            FieldDefinition f_this = null;
```
coloursofnoise: `f_this` is not used

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2541,6 +2545,29 @@ bgt.un.s	ldloca.s
...
+                    continue;
+                method = nest.FindMethod("System.Boolean MoveNext()") ?? method;
+                break;
+            }
+
+            // Goal is to flip these two commands. We move to before the Move call, emit the Disable call, then remove the original Disable call.
+            // this.MoveStaticMovers(this.startPosition - this.Position);
+            // this.DisableStaticMovers();
+            ILCursor cursor = new ILCursor(new ILContext(method));
+            cursor.GotoNext(MoveType.After, instr => instr.OpCode == OpCodes.Blt);
```
coloursofnoise: It would be preferable to target a more specific instruction, such as the `MoveStaticMovers`, instead of a `blt`. The `DisableStaticMovers` call could even be inserted immediately before the call to `MoveStaticMovers`, which still decompiles fairly well in ILSpy:
```cs
Vector2 amount = moveBlock.startPosition - moveBlock.Position;
moveBlock.DisableStaticMovers();
moveBlock.MoveStaticMovers(amount);
```
Alternatively jumping to `MoveStaticMovers` and then moving backwards to the `blt` (or two `ldloc_1`s) would be an equally viable solution

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -82,23 +86,51 @@ private class DecalImage : Component {
...
         public void MakeStaticMover(int x, int y, int w, int h, bool jumpThrus = false) {
-            StaticMover sm = new StaticMover {
+            staticMover = new StaticMover {
                 SolidChecker = s => s.CollideRect(new Rectangle((int) X + x, (int) Y + y, w, h)),
-                OnMove = v => { X += v.X; Y += v.Y; },
-                OnShake = v => { X += v.X; Y += v.Y; },
+                OnDestroy = () => {
+                    RemoveSelf();
+                    if (solid != null) 
+                        solid.RemoveSelf();
```
coloursofnoise: nitpick, but this could use the null-chaining operator (`solid?.RemoveSelf();`) (if only that could also be used for assignment, `solid?.Collidable = Active = Visible = Collidable = false;`)

coloursofnoise (changes requested):
Has the MoveBlock change been tested with mods? I don't know if there'd be any that specifically target that section of instructions so just checking sc2020 or another large collab should suffice.

Celeste.Mod.mm/MonoModRules.cs
bigkahuna443: GitHub web browser merge conflict resolver :despair: (fixed)

coloursofnoise (approved):

PR #449, opened 2022-06-06
Remove staticMover decals if their Platform is removed

### LINKED ISSUE #329: change attached decals to disappear when their solid is broken
it'd be quite helpful if, say, you have a decal with the "attached" parameter on top of a dash block or crumble on rumble block, then when that block is broken the decal also goes away.
Perhaps this could be an option in the decal registry? I'm not sure if any maps make use of this being not the case, but who knows

caeyo:
Closes #329 

If a decal with a staticMover attribute successfully attaches to a Platform, remove the decal whenever that Platform is removed. Also immediately remove the decal if it was attached to a Platform at some point in time (tracked through Session flags based on the decal's texture (i.e. the path to the image from the `decals` Atlas) and position) and is now not attached to any Platform.

Thanks to @bigkahuna443 for input and help on this - this is tangentially related to his open PR #448 but was submitted separately since it serves a different purpose and should probably be assessed outside of his. 

I opted to put together the EntityRemovedListener to make this as non-invasive as possible (hardly any vanilla entities actually call DestroyStaticMovers and this avoids having to patch them), to make it easily extend to modded entities, and to avoid janky issues with other approaches like having the decal removal lag behind by a frame or the removal of large amounts of decals being visually staggered. I also tried to make it more generalised instead of a "DecalAttacherComponent" in case any future Everest features/other mods want to make use of it. Hopefully an addition like this is okay, this was really the best option 😅 
The session flags are probably the ugliest part of this but we couldn't think of a better way to globally track whether a decal should be immediately removed - happy to hear/implement any cleaner alternatives to this.
If desired, I can make this behaviour optional via a parameter in the staticMover attribute.

### DIFF
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -29,8 +29,6 @@ namespace Celeste {
         private float showRange;
 
         private StaticMover staticMover;
-        private Vector2 position;
-        private string texture;
 
         public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
             : base(texture, position, scale, depth) {
@@ -60,8 +58,6 @@ namespace Celeste {
             }
             hideRange = 32f;
             showRange = 48f;
-            this.texture = texture;
-            this.position = position;
 
             orig_ctor(texture, position, scale, depth);
         }
@@ -101,7 +97,7 @@ namespace Celeste {
                 OnShake = v => { X += v.X; Y += v.Y; },
                 OnAttach = p => {
                     p.Add(new EntityRemovedListener(() => RemoveSelf()));
-                    (Scene as Level).Session.SetFlag($"Everest_Decal_texture:{texture}_X:{position.X}_Y:{position.Y}_WasAttached");
+                    (Scene as Level).Session.SetFlag($"Everest_Decal_texture:{Name}_X:{Position.X}_Y:{Position.Y}_WasAttached");
                 }
             };
             if (jumpThrus)
@@ -131,7 +127,7 @@ namespace Celeste {
 
         public override void Awake(Scene scene) {
             base.Awake(scene);
-            if (staticMover?.Platform == null && (scene as Level).Session.GetFlag($"Everest_Decal_texture:{texture}_X:{position.X}_Y:{position.Y}_WasAttached"))
+            if (staticMover?.Platform == null && (scene as Level).Session.GetFlag($"Everest_Decal_texture:{Name}_X:{Position.X}_Y:{Position.Y}_WasAttached"))
                 RemoveSelf();
         }
 
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -2,6 +2,7 @@
 #pragma warning disable CS0414 // The field is assigned but its value is never used
 
 using Celeste.Mod;
+using Celeste.Mod.Entities;
 using Microsoft.Xna.Framework;
 using Monocle;
 using MonoMod;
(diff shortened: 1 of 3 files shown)


bigkahuna443:
Yeah the thought process with this one is:
1. Mappers who use the "staticMover" attribute to expect the decals to attach to solids and remove themselves when the entity is removed at minimum. It's not intuitive that some solids don't work with it due to internal implementation and from the outside seems arbitrary
2. Patching solids to add missing StaticMover support would be a potentially breaking change
3. Patching decals to handle unsupported solids would limit the scope and it's highly unlikely it would break any behavior, since if a mapper did _not_ want the decal to attach they would have no reason to use the tag in the first place, which made it make sense to add this as the default behavior

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -97,7 +98,7 @@ private class DecalImage : Component {
                 OnShake = v => { X += v.X; Y += v.Y; },
                 OnAttach = p => {
                     p.Add(new EntityRemovedListener(() => RemoveSelf()));
-                    (Scene as Level).Session.SetFlag($"Everest_Decal_texture:{Name}_X:{Position.X}_Y:{Position.Y}_WasAttached");
+                    CoreModule.Session.AttachedDecals.Add($"{Name}||{Position.X}||{Position.Y}");
```
microlith57, line 133: does this break existing sessions on everest version upgrade?

caeyo: How would it? Would be pretty alarming if adding properties to a module’s session broke existing sessions on updating the mod, don’t see how Everest’s mod session would function any different. Not like Everest’s session was being used for anything else prior to this anyway, plus I ran around some levels with and without attached decals and nothing exploded in my face so not sure what else I’m supposed to be looking for.

If you’re worried about changing the approach from session flags to mod session, neither of these changes have ever been on main… only files at risk of breaking are my own lol

bigkahuna443 (approved):

Celeste.Mod.mm/Patches/Decal.cs
microlith57: i meant that existing sessions' attached flags would be ignored and some decals would reappear:
- player loads a file before this change, loads a room with a persistent dash block (w/ attached decal) and breaks it
- player updates everest
- player reenters map, and sees that the block is broken but the decal is back
is this a concern?

caeyo: Old maps that used staticMover decals never had those decals removed in the first place - the main focus of this pr is rectifying that via the EntityRemovedListener - so we're not "breaking" any existing behaviour there. The decal will just remain present in the room until the session is reset and the block comes back, and then the decals will start using the attachment behaviour being added here.

I guess it'd be nice if we could account for this and apply these fixes after-the-fact, but needing to add this functionality to track attached decals should be evidence enough that we don't have anything at our disposal to determine if these existing decals were ever attached to something, so I'm not really seeing a way forward on this. If you come up with something then I'd appreciate if you could put it together yourself, I touched on it in celestecord but not in much of a position to keep chipping away at this, the time for me to do that was 4 weeks ago when I opened the pr

coloursofnoise (approved):
There was a significant discussion on Discord regarding the implementation details for this PR which I'll try to summarize:
### This is going to be a change to the default behaviour
This should be treated as a bugfix to align with the existing expectations for StaticMover entities.

### Decals will behave as normal if they never attach to a solid
It was decided that, again, aligning with existing expectations for StaticMovers is important for users. This also removes the additional effort for the end user to create separate `staticMover` and non-`staticMover` decals if they want to use them in both contexts, with minimal overhead in the implementation (needing to keep track of attached decals in the Session to determine which should not reappear on reload).

PR #450, opened 2022-06-07
Add Decal Registry "overlay" tag

bigkahuna443:
Implements a more approachable version of objtiles. Basic functionality is:
1. Add an "overlay" tag to a decal in the DecalRegistry.xml (excludes all other tags).
2. Layer the decal over a supported tile entity (will include list in docs).
3. For every tile the decal overlaps, the tile texture of the entity will be replaced by that of the "tile" of the decal.

Other implementations were considered, e.g. supporting a custom tilesets/scenery.png image, but this implementation is pretty flexible, easy to work with, and won't require any extra editor support. It also allows users to work with segments larger than a single tile and not have to artificially break them up and piece them together.

The overlay happens in a BeforeRender hook since some tile entities don't set up their TileInterceptors until Awake.


coloursofnoise (approved):
Looks good to me, can be merged once the conflict is resolved :+1:

PR #455, opened 2022-06-13
Disable unnecessary build warnings in patch classes

Kalobi:
Building Everest had two warnings caused by fields in patch classes whose usage the compiler couldn't recognize properly. This PR adds pragmas to disable those warnings.


DemoJameson:
I think it's better to just delete these unused fields

Kalobi:
Oh yeah, you're right, they aren't actually used. The one in `Settings` wasn't even used in the commit that introduced it, so I asked coloursofnoise, who authored that commit, whether there's a reason for it being there, and will update accordingly once I get a response.

coloursofnoise (approved):
> The one in Settings wasn't even used in the commit that introduced it

I expect it's just something I copied over from the V1/V2 classes without checking to see if it was actually used. 

Seems fine to remove both.

PR #456, opened 2022-06-13
Add support for non-looping complete screen animations

Kalobi:
Currently, any animation on complete screens except for the parallax must loop indefinitely. This PR implements support for non-looping layers by adding a new meta attribute Loop to complete screen layers, true by default, that causes the animation to stay on the last frame if set to false. This is implemented both for .meta.yaml and .xml so that it can be used both for custom maps and global skinmods.

### DIFF
--- a/Celeste.Mod.mm/MonoModRules.cs
+++ b/Celeste.Mod.mm/MonoModRules.cs
@@ -427,6 +427,12 @@ namespace MonoMod {
     /// </summary>
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchPlayerStarFlyReturnToNormalHitbox))]
     class PatchPlayerStarFlyReturnToNormalHitboxAttribute : Attribute { }
+    
+    /// <summary>
+    /// Patches the method to support non-looping complete screen layers.
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchCompleteRendererImageLayerRender))]
+    class PatchCompleteRendererImageLayerRenderAttribute : Attribute { }
 
 
     static class MonoModRules {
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/Patches/CompleteRenderer.cs
```diff
@@ -97,7 +98,47 @@ public UILayerNoXML(CompleteRenderer renderer, MapMetaCompleteScreenLayer meta)
 
         }
 
-        public class ImageLayerNoXML : ImageLayer {
+        public class patch_ImageLayer : ImageLayer {
+
+            public bool Loop;
+            public patch_ImageLayer(Vector2 offset, Atlas atlas, XmlElement xml) : base(offset, atlas, xml) {
+                //no-op
+            }
```
coloursofnoise:
```suggestion
            public patch_ImageLayer(Vector2 offset, Atlas atlas, XmlElement xml)
                : base(offset, atlas, xml) {
                //no-op
            }
```

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2473,6 +2479,17 @@ bgt.un.s	ldloca.s
             cursor.RemoveRange(3);
         }
 
+        public static void PatchCompleteRendererImageLayerRender(ILContext context, CustomAttribute attrib) {
+            MethodReference m_ImageLayer_Render = MonoModRule.Modder.FindType("Celeste.CompleteRenderer/ImageLayer")
+                .Resolve().FindProperty("ImageIndex").GetMethod;
```
coloursofnoise:
```suggestion
            MethodReference m_ImageLayer_Render = context.Method.DeclaringType.FindProperty("ImageIndex").GetMethod;
```

Celeste.Mod.mm/MonoModRules.cs
```diff
@@ -2473,6 +2479,17 @@ bgt.un.s	ldloca.s
...
+            
+            ILCursor cursor = new(context);
```
coloursofnoise:
```suggestion
            ILCursor cursor = new ILCursor(context);
```

coloursofnoise (commented):
Some style nitpicks, fields should all be defined at the top of the class, with an empty line before the constructor.

Some of these should probably be added to the `.editorconfig`, such as target-typed new expressions :sweat_smile:

Celeste.Mod.mm/MonoModRules.cs
Kalobi: What's the issue with target-typed new expressions?

Kalobi: Fixed

coloursofnoise (approved):
LGTM 👍

PR #457, opened 2022-06-13
Implement DebugRC endpoint for console commands

### LINKED ISSUE #390: Console command DebugRC endpoint
Small feature request, I'd like to send commands to the game plz (URL encoded cause spaces and such)

Kalobi:
Closes #390 
Implement a new DebugRC endpoint `/console` that executes console commands and shows the output (ignoring color information). The command parsing should be identical to that of the ingame console.


DemoJameson:
It would be nice if there is a way to list all commands name and help text, as well as support for click-to-execute command

Kalobi:
> It would be nice if there is a way to list all commands name and help text, as well as support for click-to-execute command

Would this be a separate endpoint (`commandlist` or something)? And what should click-to-execute look like for commands with parameters?

DemoJameson:
> Would this be a separate endpoint (commandlist or something)?

I don't think we need another endpoint, just add them to the end of the console endpoint page

> And what should click-to-execute look like for commands with parameters?

we can refer to the help command
```
// http://localhost:32270/console?command=help%20load
:: load [idOrSID:string="0" level:string]
test a level
```

Kalobi:
Alright, I implemented it so that you get the command list if you call the endpoint without a command. So if you call it with a command, you still only get the command output.

Celeste.Mod.mm/Patches/Monocle/Commands.cs
```diff
@@ -42,6 +43,9 @@ class patch_Commands : Commands {
 
         private static readonly Lazy<bool> celesteTASInstalled = new Lazy<bool>(() => Everest.Modules.Any(module => module.Metadata?.Name == "CelesteTAS"));
 
+        // redirects command logs to the StringBuilder when not null, only set this from main thread
+        public StringBuilder debugRClog;
```
coloursofnoise: Is there a reason to make this available to mods or should it be made `internal` given the restrictions? If there is, this comment should probably be a proper XMLDoc.

Kalobi: Probably makes more sense to make this internal, yeah. At least I can't think of a good reason why modders would want to redirect debug command output. The only mod that really does things that technical is CelesteTAS, and that uses its own DebugRC endpoints anyway.

bigkahuna443 (approved):

PR #460, opened 2022-06-28
Fix TypeLoadException during module instantiation

Popax21:
When a mod assembly contains a type which is a subclass of a missing optional dependency's type, they sometimes (likely influenced by what runtime is used) still show up in the type list, and as such proceed to crash during the module instantiation loop and prevent the entire mod from loading. This PR fixes this by handling the exception and only logging a warning instead.


coloursofnoise (approved):
`mod.Register()` moved outside try/catch to ensure only exceptions from `.IsAssignableFrom` are caught.

PR #466, opened 2022-07-04
add a LevelData cache to `MapData.Get` to increase its performance

### LINKED ISSUE #465: poor performance getting the current `LevelData`

Yes

### Describe your request

currently, `LevelData Session.LevelData` is implemented as:
```cs
public LevelData LevelData => MapData.Get(Level);
```
which has O(n) performance as it loops over all `LevelData` objects to check their names.
because the names of `LevelData`s should not change at runtime, the `Session` is safe to cache its `LevelData` property, invalidating when its `string Session.Level` changes, which would have O(1) performance on all but the first call.

this has performance implications everywhere the current `LevelData` is used, notably everywhere the room bounds are needed.

### Additional context

_No response_

microlith57:
re. #465

### DIFF
--- a/Celeste.Mod.mm/Patches/MapData.cs
+++ b/Celeste.Mod.mm/Patches/MapData.cs
@@ -7,7 +7,6 @@ using MonoMod;
 using MonoMod.Utils;
 using System;
 using System.Collections.Generic;
-using System.Xml.Serialization;
 using System.Linq;
 
 namespace Celeste {
--- a/Celeste.Mod.mm/Patches/MapData.cs
+++ b/Celeste.Mod.mm/Patches/MapData.cs
@@ -19,9 +19,6 @@ namespace Celeste {
 
         private Dictionary<string, LevelData> levelsByName = new Dictionary<string, LevelData>();
 
-        [XmlIgnore]
-        internal uint session_leveldata_cache_validity = 0;
-
         public MapMetaModeProperties Meta {
             get {
                 MapMeta metaAll = AreaData.Get(Area).GetMeta();
@@ -42,9 +39,6 @@ namespace Celeste {
 
         [PatchMapDataLoader] // Manually manipulate the method via MonoModRules
         private void Load() {
-            // invalidate the LevelData cache
-            session_leveldata_cache_validity += 1;
-
             // reset those fields to prevent them from stacking up when reloading the map.
             DetectedStrawberries = 0;
             DetectedHeartGem = false;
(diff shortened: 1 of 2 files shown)


Kalobi:
Is there a reason patch_Session couldn't inherit from Session? That would make it unnecessary to reintroduce the Level field.

microlith57:
good question. i will look into it

DemoJameson:
perhaps replace the getter directly like this
https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Patches/Level.cs#L43-L51

DemoJameson:
since you have improved mapData.Get, I think there's no need to cache session.LevelData
do we really need to squeeze the little performance

microlith57:
@DemoJameson done!

Celeste.Mod.mm/Patches/MapData.cs
```diff
@@ -45,10 +47,18 @@ public patch_MapData(AreaKey area)
             DetectedCassette = false;
             DetectedStrawberriesIncludingUntracked = 0;
 
+            if (levelsByName is null) {
```
bigkahuna443, line 50: don't think this should ever be null since you're initializing it with an empty dictionary further up. don't think there is a significant difference between creating a new list vs clearing the old one either

Celeste.Mod.mm/Patches/MapData.cs
```diff
@@ -115,6 +125,13 @@ public patch_MapData(AreaKey area)
             return orig_StartLevel() ?? Levels[0];
         }
 
+        public new LevelData Get(string levelName) {
```
bigkahuna443, line 129: shouldn't this be tagged with MonoModReplace?

bigkahuna443:
Can you edit the PR title to more accurately describe the changes since it got revised?

Celeste.Mod.mm/Patches/MapData.cs
DemoJameson: > don't think this should ever be null since you're initializing it with an empty dictionary further up. don't think there is a significant difference between creating a new list vs clearing the old one either

It seems that fields added in patch_MapData are not initialized, even if you assign values to them when you define them

microlith57: correct, when i didn't have this null check it wasn't initialised
personally i prefer clearing the old dictionary over recreating it as it's more obvious what the intent is and also possibly more efficient (no new allocation/garbage collection needed)

bigkahuna443: Oh huh, yeah apparently field initialization doesn't work in patch classes in general due to how C# handles it. And that's fair, I generally try to keep things concise/consistent so I was comparing to the other lists that get initialized

Celeste.Mod.mm/Patches/MapData.cs
microlith57: done!

DemoJameson (approved):

PR #468, opened 2022-07-05
Implement Custom Backdrops through custom attribute

### LINKED ISSUE #461: CustomBackdrop attribute

Yes

### Describe your request

Exactly like you decorate a custom entity class with a `[CustomEntity]` attribute, and ensure that it has a constructor with a valid signature, as Everest requires, there could be a `[CustomBackdrop]` attribute, destined to be placed on classes that derives from `Celeste.Backdrop`, and which takes in an ID string (or more).
A custom backdrop class with this attribute would also require a valid constructor. Such a constructor would need to take as parameter the data of the backdrop, which can be retrieved through a `Celeste.BinaryPacker.Element`, which itself works very similarly to how `EntityData` works. Maybe a `BackdropData` class could be created to load data that is common to all backdrops, like flags, looping, etc... however the problem with this is that Celeste already loads those values when loading any backdrop, and fields in `Celeste.Backdrop` are set automatically.

### Additional context

This is how I would define a custom backdrop with such an attribute:
```cs
namespace Celeste.Mod.MyMod;

[CustomBackdrop("MyMod/MyCustomBackdrop")]
public class MyCustomBackdrop : Backdrop {
    public MyCustomBackdrop(BinaryPacker.Element data)
        : base() {
        bool someFlag = data.AttrBool("someFlag");
        // do whatever you want...
    }
}
```

Right now though, you need to subscribe to the `Everest.Events.Level.OnLoadBackdrop` event and instantiate your backdrop manually, if the element's name happens to match your backdrop ID:
```cs
// Load and Unload are called from the module class,
// or those could be the Load and Unload methods that you must implement in the module class.
public static void Load() {
    Everest.Events.Level.OnLoadBackdrop += Mod_Level_OnLoadBackdrop;
}

public static void Unload() {
    Everest.Events.Level.OnLoadBackdrop -= Mod_Level_OnLoadBackdrop;
}

private static Backdrop Mod_Level_OnLoadBackdrop(MapData map, BinaryPacker.Element child, BinaryPacker.Element above) {
    string name = child.Name;

    if (name.Equals("MyMod/MyCustomBackdrop", StringComparison.OrdinalIgnoreCase))
        return new MyCustomBackdrop(child);

    return null;
}
```

catapillie:
This adds a `CustomBackdropAttribute` attribute class under the `Celeste.Mod.Backdrops` namespace. You can use it instead of subscribing to an everest provides when custom backdrops are loaded. This makes registering custom backdrops much easier, and could have its own section in the wiki, for those who are looking for implementing their own backdrops.

When an Everest module is registered, after looking for custom entities and events, it will, in the same manner, look for custom backdrops. IDs are associated with a loader delegate, and stored in a dictionary in `patch_MapData`.

For loading custom backdrops, the `patch_MapData.LoadCustomBackdrop` will now:
* attempt to load any custom backdrop through the `Everest.Events.Level.OnLoadBackdrop` event, which thus have priority over any other custom backdrop
* search for a matching ID in the dictionary that contains the registered custom backdrops, and invoke the associated loader delegate if found
* do the rest, notably, perform the custom loading of the `rain` backdrop, or return `null`.

Example:
```cs
[CustomBackdrop("CustomBackdrops/Test")]
public class CustomBackdrop : Backdrop {
    public CustomBackdrop (BinaryPacker.Element data)
        : base() {
        int valueA = data.AttrInt("a", defaultValue: 3);
        bool valueB = data.AttrBool("b");
        // etc...
    }

    public override void Update(Scene scene) { ... }
    public override void Render(Scene scene) { ... }
}
```

Similarly to custom entities, instead of having a constructor which takes in a `BinaryPacker.Element`, you can have a static method which takes it in, and returns a `Backdrop`. By default, the method's name must be `Load`, but it can be changed by appending `=Name` at the end of the custom backdrop ID to set it to `Name`, for instance.

```cs
[CustomBackdrop("CustomBackdrops/Test=CreateBackdrop")]
public class CustomBackdrop : Backdrop {
    // ...
    public static Backdrop CreateBackdrop(BinaryPacker.Element data) => new();
}
```

Closes #461.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.cs
@@ -1,4 +1,5 @@
-﻿using Celeste.Mod.Core;
+﻿using Celeste.Mod.Backdrops;
+using Celeste.Mod.Core;
 using Celeste.Mod.Entities;
 using Celeste.Mod.Helpers;
 using Celeste.Mod.UI;
@@ -652,7 +653,7 @@ namespace Celeste.Mod {
                         genName = genName.Trim();
 
                         patch_EventTrigger.CutsceneLoader loader = null;
-
+                        
                         ConstructorInfo ctor;
                         MethodInfo gen;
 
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.cs
```diff
@@ -652,7 +653,7 @@ public static partial class Everest {
                         genName = genName.Trim();
 
                         patch_EventTrigger.CutsceneLoader loader = null;
-
+
```
microlith57: whitespace :despair:

microlith57 (changes requested):

Celeste.Mod.mm/Mod/Everest/Everest.cs
catapillie: my bad, fixed.

coloursofnoise (commented):
Should the `CustomBackdropAttribute` file be lumped into the `Entities` folder instead of in its own? Currently that folder is a bit of a catch-all, including renderers and cutscenes.

catapillie:
I wouldn't mind doing so. The reason I didn't is that everything in the `Entities` folder is related to entities in some way, most of the files being entities themselves, attributes for custom entity stuff, and there's like one component.

coloursofnoise:
Leaving it in `Backdrops` should be fine, I would just make a note to specify the namespace in the documentation, since it differs from the other `Custom___Attribute`s

PR #469, opened 2022-07-06
Improve GenerateModAssemblyResolver performance

bigkahuna443:
Alternate implementation of Ja's assembly resolver fix in #424 that checks the content map for the assembly instead of re-crawling already loaded mods. Should maintain current mod loading order, while still removing the performance hit from unzipping archives.

EDIT: one part of the discussion that didn't get posted in Ja's PR is that anything that uses `FakeAssembly.GetTypes[Safe]` also triggers the assembly resolver for each missing type. So anything that re-initializes the tracker like hot reloading, unblacklisting installed dependencies, etc. can cause a massive load spike. Can also occur with CU2 chapter panels and Eevee Helper on room transitions.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -498,9 +498,6 @@ namespace Celeste.Mod {
 
                 using var _ = new ScopeFinalizer(() => Events.Everest.LoadMod(meta));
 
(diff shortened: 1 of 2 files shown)


DemoJameson:
LGTM

DemoJameson (approved):

bigkahuna443:
Went over it with Jade in DMs and she gave a thumbs up.

PR #471, opened 2022-07-08
Improve asset ingestion

bigkahuna443:
1. Add `AssetTypeNonConflict` for mods to subclass from. ModAssets with this base type won't log conflicts.
2. Extend blacklist to avoid ingesting Ahorn and Loenn files as well as common code/documentation files.
3. Refactor `GuessType` to allow custom mod types to take precedence over generic types.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
@@ -14,26 +14,34 @@ using System.Reflection;
 using System.Threading;
 
 namespace Celeste.Mod {
-    // Special meta types.
-    public sealed class AssetTypeDirectory { private AssetTypeDirectory() { } }
-    public sealed class AssetTypeAssembly { private AssetTypeAssembly() { } }
-    public sealed class AssetTypeYaml { private AssetTypeYaml() { } }
-    public sealed class AssetTypeXml { private AssetTypeXml() { } }
-    public sealed class AssetTypeText { private AssetTypeText() { } }
-    public sealed class AssetTypeLua { private AssetTypeLua() { } }
-    public sealed class AssetTypeMetadataYaml { private AssetTypeMetadataYaml() { } }
+    /// <summary>
+    /// Special meta type for assets. 
+    /// A ModAsset with a Type field that subclasses from this will log path conflicts.
+    /// </summary>
+    public abstract class AssetTypeConflict { }
+
+    // Asset types for which conflicts are not important
+    public sealed class AssetTypeDecalRegistry { private AssetTypeDecalRegistry() { } }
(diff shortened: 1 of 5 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
```diff
@@ -798,9 +801,12 @@ public static void Add(string path, ModAsset metadata)
                     }
 
                 } else if (file.StartsWith("Ahorn/")) {
-                    // Special case: Fallback type for anything inside of the Ahorn folder.
+                    // Fallback types for anything inside of the Ahorn or Loenn folders.
                     // Will be ignored during collision checks.
```
microlith57: nitpick: this comment applies to both `else if` branches but is inside of one of them

microlith57 (approved):

Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
bigkahuna443: Yeah auto-formatter wasn't liking the comment position but I do like it better outside lol, fixed

coloursofnoise:
Another one I see regularly is the backup `everest.yaml` created by the Ahorn Dependency Generator.

It may also be beneficial to expose an endpoint for mods to register their own no-conflict paths, such as the CollabUtils or SkinModHelper config files.

bigkahuna443:
> It may also be beneficial to expose an endpoint for mods to register their own no-conflict paths, such as the CollabUtils or SkinModHelper config files.

Yeah that's a good point, I noticed that you could register an OnGuessType delegate but not label it as "non-conflict". And the delegate has lower priority than all the other types, so e.g. a custom YAML asset type wouldn't even get registered. I have an idea on how to improve both of those I think

Kalobi:
I looked into this a while ago, but got distracted by school. My idea was a .everestignore file that works like a .gitignore and lets people exclude files from being scanned by Everest, probably using [DotIgnore](https://github.com/markashleybell/MAB.DotIgnore). I guess that would only work for things like documentation though, not for things that you want to be scanned, but prevent from conflicting.

bigkahuna443:
Updated:

This change makes assets *not* log conflicts by default, which seems more reasonable than filtering everything that gets ingested.

Added `AssetTypeConflict` for existing conflict types + specifying user-specified types they want to log conflicts for.

Removed `AssetTypeAhorn` and `NonConflictTypes` as they aren't relevant with the new change. Couldn't find either in use anywhere.

Moved custom asset type resolution before generic asset types. Improved formatting in the section.

Brokemia:
I noticed you're removing the `.git` blacklist. Is there a reason behind that?

Brokemia:
Ignore me, it's because you're blacklisting everything starting with `.` anyway

PR #473, opened 2022-07-10
Allow suppressing error log opening on crash

coloursofnoise:
Changing `OpenErrorLogOnCrash` to `false` in `modsettings-Everest.celeste` or setting the environment variable `EVEREST_NO_ERRORLOG_ON_CRASH` to `1` will prevent the error log from being automatically opened upon crashing.

The env var is more reliable than the setting, because if the game crashes before settings are read, `OpenErrorLogOnCrash` defaults to `true`.

Open to suggestions for better/other names.


WEGFan:
maybe `EVEREST_NO_OPEN_ERRORLOG_ON_CRASH`, the current one makes me think it's "completely disable error log"

bigkahuna443 (approved):

PR #474, opened 2022-07-11
Make logging less "verbose"

### LINKED ISSUE #335: Default log level should be Info
Currently, the default log level is Verbose, and the default verbosity the game runs at it also Verbose. This is a bad idea; the default log level being the lowest possible means that a modder cannot have message at a lower log level without using `SetLogLevel`. Everest should follow the common practice of having Info be both the default verbosity and default log level. That way, stuff explicitly marked Verbose won't show up unless explicitly desired, as it should be. It would also allow a user to change the default log level to Verbose to show such messages anyway. This has the potential to reduce log spam, if used correctly, while skipping the friction of having to manually set the log level for every prefix.

With it, I'd like to request a mod option for the verbosity (ideally, it would override any SetLogLevel).

### LINKED ISSUE #336: There is too much logging
This is somewhat adjacent to #335, but more... _radical_. Logging is one of the slowest actions to occur during load, be it game or map loading. Currently, there is just too much going on, with everyone logging their IL hooks for example, which is entirely irrelevant during normal operation. I propose changing the default log verbosity to Info as specified in #335, but keep the default log level at Verbose, so that by default, log messages will *not* show up unless explicitly specified in `Logger.Log` or using `SetLogLevel`. Along with the proposed mod option, this would allow letting a user increase the verbosity for debugging purposes while keeping load times low during normal operation. It would also force modders to learn about log levels and how to use them.

bigkahuna443:
1. Changes the default tag minimum level to `Info`.
2. Changes detour log level to Debug. If the `logdetours` debug command is used, print regardless of log level.
3. Adds `setloglevel` debug command

This makes logs much easier to read. Log levels for individual tags can still be set in mod settings or from command line, but the ones for Everest that are useful for debugging crashes should all be Info or higher already. Detour logs can be printed on demand with the `logdetours` command, so we're not really losing functionality there.

I'll probably also work on a logging page in the wiki so we can define some good practices.
EDIT: new wiki section [here](https://github.com/EverestAPI/Resources/wiki/Recommended-Practices-and-Pitfalls#logging).

Resolves #335 and resolves #336.


bigkahuna443:
I'm going to look at adding a mod setting to override the minimum log level for all tags -- this seems better than having to edit tags individually in the settings file.

WEGFan:
You can use this to enable all logs in settings without editing tags individually, since the setting checks prefix, not full match:

```yaml
LogLevels:
  '': Verbose
```

coloursofnoise (approved):
Looks good to me, as discussed on Discord this is a change that should be announced to the community alongside having a good wiki page.

PR #478, opened 2022-07-17
Apply offset correctly to YAML complete screens

bigkahuna443:
This has gone unnoticed for four years :)


Kalobi (approved):
Sensible change, but should figure out whether old maps accidentally used the non-functional property and thus will have broken endscreens.

bigkahuna443:
Maps were reviewed, only a couple were affected negatively and the creator was informed about the change.

PR #480, opened 2022-07-26
Add page up/down functionality to save file select screen

Brokemia:
Pretty straightforward. I have 33 saves, so this will save me quite a bit of annoyance


bigkahuna443 (approved):
Tested, looks good to me!

PR #481, opened 2022-07-26
Recognize files from Loenn and ignore them for the purpose of conflicts

Brokemia:
Seeing the warnings in the log (mostly for en_gb.lang) was bothering me, so now Everest should ignore conflicts for Loenn files

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
@@ -502,7 +502,8 @@ namespace Celeste.Mod {
                 typeof(AssetTypeDialogExport),
                 typeof(AssetTypeAhorn),
                 typeof(AssetTypeLoenn),
-                typeof(AssetTypeSpriteBank)
+                typeof(AssetTypeSpriteBank),
+                typeof(AssetTypeDecalRegistry)
             };
 
             internal readonly static List<string> LoadedAssetPaths = new List<string>();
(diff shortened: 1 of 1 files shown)


Brokemia:
I still get a warning for OutbackHelper and Pandora's box for some reason.
`CONFLICT for asset path Loenn/lang/en_gb.lang (OutbackHelper vs PandorasBox)`
This has dealt with all the other conflicts though

bigkahuna443:
I opened #471 to address this, but ended up marking it as a draft since there were some bigger changes I wanted to look into in the content system. Fine with this going in and leaving that one open for discussion, I think it would make sense to pull in the change for changing the Decal Registry + Sprite Bank types to non-conflict into this PR though 
EDIT: I looked into the conflict issue you're having, looks like Outback Helper has a `en_gb.lang.txt` file, so it gets tagged as a text-type (higher-priority + conflict) and gets the file extension stripped, making it conflict with normal `.lang` files

Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
```diff
@@ -801,6 +803,10 @@ public static void Add(string path, ModAsset metadata)
                     // Special case: Fallback type for anything inside of the Ahorn folder.
                     // Will be ignored during collision checks.
                     type = typeof(AssetTypeAhorn);
+                } else if (file.StartsWith("Loenn/")) {
```
bigkahuna443, line 807: the Loenn directory can also be `Lönn`, so would check for that as well

Brokemia:
Whoops, I missed your earlier PR on this. If you think it makes sense to merge in some of those changes I can do that.
Regarding the Outback Helper issue, should I do the same thing for text files that I did for lua files, make the Loenn folder check a higher priority, or just blame Outback Helper and ignore it?

Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
bigkahuna443: re: cruor this is no longer used, and couldn't find mods using it in the database so not needed

Kalobi:
Implemented by #471

PR #482, opened 2022-07-27
Add .everestignore file

Brokemia:
There is a global .everestignore in the Mods folder and mods can also ship with one in their root. Specifies files inside mods to ignore when loading mod content. These files follow the same format as .gitignore
The global .everestignore runs the risk of stopping a mod from loading essential files if misused, so I'm not totally sure if it's a great idea.


Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -390,6 +425,11 @@ public static class Loader {
                 FileSystemModContent contentMeta = new FileSystemModContent(dir);
                 EverestModuleMetadata contentMetaParent = null;
 
+                string ignorePath = Path.Combine(dir, ".everestignore");
+                if(File.Exists(ignorePath)) {
```
Kalobi:
```suggestion
                if (File.Exists(ignorePath)) {
```

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -295,11 +319,22 @@ public static class Loader {
                             continue;
                         }
                     }
+                    if(zip.ContainsEntry(".everestignore")) {
```
Kalobi: Why not do this in the preceding loop over the zip entries? `ContainsEntry` presumably does another loop, which seems wasteful. Also very cursed thought, should we support people .everestignoring any of the yaml files that are checked for above?

Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
```diff
@@ -606,6 +611,11 @@ public static class Content {
                     Path.GetFileName(path).StartsWith("._"))
                     return false;
 
+                if(Loader.GlobalEverestIgnore.IsIgnored(path, metadata.Type == typeof(AssetTypeDirectory)) ||
+                    (metadata?.Source?.Ignore?.IsIgnored(path, metadata.Type == typeof(AssetTypeDirectory)) ?? false)) {
```
Kalobi:
```suggestion
                    (metadata.Source?.Ignore?.IsIgnored(path, metadata.Type == typeof(AssetTypeDirectory)) ?? false)) {
```
If `metadata` is `null` here, the first argument of the `||` has already crashed.

Celeste.Mod.mm/Celeste.Mod.mm.csproj
```diff
@@ -51,6 +51,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="MAB.DotIgnore" Version="3.0.2" />
```
Kalobi: Not sure there's any consistent convention for order of dependencies, but right at the top doesn't seem appropriate for this one.

Kalobi:
Also yeah the global .everestignore will probably create more problems than it solves, so unless there's some really compelling reason to have it, it can probably go.

Celeste.Mod.mm/Celeste.Mod.mm.csproj
Brokemia: I just let Visual Studio do its thing. I'll move it to the bottom though

Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
```diff
@@ -22,6 +23,7 @@ public sealed class AssetTypeXml { private AssetTypeXml() { } }
     public sealed class AssetTypeText { private AssetTypeText() { } }
     public sealed class AssetTypeLua { private AssetTypeLua() { } }
     public sealed class AssetTypeMetadataYaml { private AssetTypeMetadataYaml() { } }
+    public sealed class AssetTypeEverestIgnore { private AssetTypeEverestIgnore() { } }
```
bigkahuna443: Like Kalobi mentioned, with #471 `.everestignore` would be blacklisted from ingest by default. We _could_ whitelist it, but as we never need it again after loading it in I think it'd be cleaner to just remove it from the pipeline

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -114,6 +115,16 @@ public static class Loader {
                 Tuple.Create("Nameguy's D-Sides", _VersionMax, "Monika's D-Sides", _VersionMax),
             };
 
+            /// <summary>
+            /// The path to the Everest /Mods/.everestignore file.
+            /// </summary>
+            public static string PathGlobalEverestIgnore { get; internal set; }
```
bigkahuna443: I agree with Kalobi, I don't see a good use case for a global `.everestignore` and think it could cause more harm than good. Plus running recursive rule checks + regex on tens of thousands of assets at launch sounds like it could be a performace hit vs. the occasional mod-local one (hard to say w/o testing though)

Brokemia: The point of the global `.everestignore` is to allow users to ignore stuff from mod creators who aren't providing their own for things they really should be. For example, if users want to hide demo/test maps. But yes, I agree it's too easy to cause issues that are difficult to track down, so I'll remove it.

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -294,12 +297,25 @@ public static class Loader {
                             }
                             continue;
                         }
+                        if (entry.FileName == ".everestignore") {
+                            List<string> lines = new();
```
Kalobi:
```suggestion
                            List<string> lines = new List<string>();
```
Everest code style only uses target-typed new for horrendously long types.

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -294,12 +297,25 @@ public static class Loader {
...
+                            using (MemoryStream stream = entry.ExtractStream())
+                            using (var reader = new StreamReader(stream)) {
```
Kalobi:
```suggestion
                            using (StreamReader reader = new StreamReader(stream)) {
```
More code style consistency, especially here where there's two other similar paragraphs.

Kalobi (changes requested):

Kalobi (approved):

PR #483, opened 2022-07-28
Make debug map key rebindable

Kalobi:
This PR allows rebinding the key to open the debug map instead of forcibly rebinding it from Tab to F6. The default is kept at F6 to match the behavior in existing Everest versions.


coloursofnoise (approved):
Looks good to me :+1:

PR #486, opened 2022-07-30
Nearly double Logger.Log performance

JaThePlayer:
Logging is one of the slowest actions during startup, in my case it takes 20% of startup! Of course, this is related to #336, but the `Logger.Log` function implementation itself is really sub-optimal, as it does tons of `Console.Write` calls, which all flush immediately.

This PR improves the performance of `Logger.Log`, by merging all the `Console.Write` calls into one call with string interpolation. While this does cause extra allocations, making use of allocation-free methods such as using a shared `char[]` buffer didn't seem to provide a significant increase in performance, and IMO it's not worth it, as IO is still ~99% of the cost.

Other minor optimisations:
- Avoiding boxing of the `DateTime` struct by explicitly calling `.ToString()`, despite IDE suggestions.
- Using a custom `.FastToString()` method for converting the `LogLevel` enum to a `string`, a well-known enum performance trick.

```
old:
DONE LOADING (in 13169ms), 2724 spent in Logger.Log (20,6%)
new:
DONE LOADING (in 12544ms), 1521 spent in Logger.Log (12.1%)
```

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Logger.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Logger.cs
@@ -97,8 +97,8 @@ namespace Celeste.Mod {
         /// <param name="str">The string / message to log.</param>
         public static void Log(LogLevel level, string tag, string str) {
             if (shouldLog(tag, level)) {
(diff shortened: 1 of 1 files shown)


bigkahuna443 (approved):
Looks good!

PR #487, opened 2022-07-30
Allow localizing the Granny ch9 easter egg

maddie480:
I have got a request from someone wanting to translate the "I see you have discovered Debug Mode" line from Granny in the chapter 9 easter egg... except it is added to the game's dialogue through code, rather than through a dialog txt file like the rest of dialogue, probably to hide it away.

**If a mod defines the `CH10_GRANNY_EASTEREGG` key, it is overwritten and therefore doesn't show up in-game.** This PR changes this in order to only use the hardcoded value if no value is already defined for this dialog key.

### DIFF
--- a/Celeste.Mod.mm/Patches/NPC07X_Granny_Ending.cs
+++ b/Celeste.Mod.mm/Patches/NPC07X_Granny_Ending.cs
@@ -44,10 +44,7 @@ namespace MonoMod {
             new ILContext(routine).Invoke(ctx => {
                 ILCursor cursor = new ILCursor(ctx);
 
-                if (!cursor.TryGotoNext(MoveType.After, instr => instr.MatchLdstr("{portrait GRANNY right mock} I see you have discovered Debug Mode."))) {
-                    return;
-                }
-
+                cursor.GotoNext(MoveType.After, instr => instr.MatchLdstr("{portrait GRANNY right mock} I see you have discovered Debug Mode."));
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/NPC07X_Granny_Ending.cs
```diff
@@ -0,0 +1,55 @@
...
+    class PatchCh9EasterEggText : Attribute { }
+
+    static partial class MonoModRules {
+        public static void PatchCh9EasterEggText(MethodDefinition method, CustomAttribute attrib) {
+            MethodDefinition routine = method.GetEnumeratorMoveNext();
+
+            new ILContext(routine).Invoke(ctx => {
+                ILCursor cursor = new ILCursor(ctx);
+
+                if (!cursor.TryGotoNext(MoveType.After, instr => instr.MatchLdstr("{portrait GRANNY right mock} I see you have discovered Debug Mode."))) {
```
Kalobi: Patches in Everest should generally fail hard, so `GotoNext` is preferred.

Kalobi:
Do you think Everest should provide its own localization for this secret in its supported languages or does that fall under changing vanilla?

maddie480:
I think that would fall under changing vanilla 🤔

bigkahuna443 (approved):
Tested and changes look good, approved!

PR #488, opened 2022-07-30
Add attribute to use subheaders in mod menus

Kalobi:
This PR adds a `SettingSubHeaderAttribute` that can be used on a setting to add a subheader before that setting. When used for settings that appear in the main mod options menu, it will omit the padding before the subheader to prevent confusion with separate mod sections. For controls, it maintains the padding, since controls form their own submenu.
Sample images:
image
image


coloursofnoise (approved):
Looks good to me, although it makes me realize how many different ways we have to make "SubMenus" now :sweat_smile: 

This will also need to be added to the [wiki page](https://github.com/EverestAPI/Resources/wiki/Mod-Settings) once merged, which is also still missing #325.

bigkahuna443 (approved):
Tested it out, looks good to me too! Surprised this wasn't added yet, pretty natural way to break up some of the bigger mod menus

Issue #484, opened 2022-07-28
Missing localization for key bindings


Kalobi:
The Debug Console keybind and all mountain camera related keybinds are currently not localized. I can add localization in German, but other people will have to take care of the remaining languages.

PR #489, opened 2022-07-30
German and better English localization for keybinds

Kalobi:
This clarifies what the mountain camera controls are for and localizes keybindings in German. Feedback from other German speakers is appreciated. This partially addresses #484 . Once #488 is merged, I plan to also add subheaders to the keybindings menu that will be able to further clarify.


Kalobi:
Subheaders have been added and localized in German. The mountain camera keybinds have been reverted to only contain "Camera" since the subheader now explains the context.

Issue #314, opened 2021-06-08
Issues with the Debug Console


coloursofnoise:
- [x] Using the `q` command ([added by Everest](https://github.com/EverestAPI/Everest/blob/0d098aab868084d5b5fc649fd15a35edd6e21557/Celeste.Mod.mm/Mod/Helpers/Commands.cs#L6-L9)) while in-game to close the console will open the pause menu, since it seems that the `Enter` input is not consumed and `Enter` is now bound to pause by default in 1.4.0.0
  - This can also be seen by rebinding `Enter` to any other action and then using the `q` command when that action could otherwise be used. 
- [ ] The <code>\`</code>  character, when *typed* on a non-US keyboard layout, opens and closes the debug console as expected, however *pressing the key* in the location of the <code>\`</code> key on the US layout will close the console but not open it
- [ ] Opening the console with the `.` key, as [added by Everest](https://github.com/EverestAPI/Everest/blob/0d098aab868084d5b5fc649fd15a35edd6e21557/Celeste.Mod.mm/Patches/Monocle/Commands.cs#L50) seems to put the console into a state of being "extra-open", and processes key inputs differently than when opening it with <code>\`</code>, such that on a non-US layout, the key corresponding to the *position of the US* <code>\`</code> key no longer closes the console, and instead types it's corresponding character. When the <code>\`</code> character is *typed*, it causes the console to return to "normal" behaviour without closing it, allowing either the <code>\`</code> character or corresponding <code>\`</code> key to close it.

coloursofnoise:
The above was found using a `US` standard keyboard remapped through the Windows 7 language settings to the `Canadian French` layout

Spirialis:
If this code gets overhauled, the addition of an option to remove or rebind the `.` key would be greatly appreciated. The `.` key works fine as a normal keybind in vanilla celeste with debug enabled, but causes issues with Everest. 

I've spoken to a couple speedrunners who can't use their normal control schemes with Everest enabled, and personally I've had issues not only with being unable to incorporate the key into my control scheme but accidentally pressing it (`.` is prime keyboard real estate in Dvorak!)

VerdalJG:
Problem: when I open the console (using `) and type something, the first letter is either always inputting an asterisk (if I press E, and then typing in the console works normally), or it just closes the console (inputting Q,W, and other binds). This happens in ENG-INTL keyboard and ENG-ES keyboard; ENG-US works fine.

PR #490, opened 2022-07-31
Consume enter press for "q" command

Kalobi:
This prevents the Enter press when executing the `q` command from being recognized as a new keypress by `MInput`. This fixes the first point of #314 .
Explanation: While the console is open, `MInput` is made to believe that no buttons are pressed. By marking Enter as pressed before closing the console, `MInput` thinks that Enter was already pressed on the previous frame and therefore does not register a new button press.


bigkahuna443 (approved):
Looks good, clean way to handle it.

PR #491, opened 2022-07-31
Fix tileset sounds

Kalobi:
Fix the way tileset sounds are determined. This fixes two related issues introduced by #445 : Foreground tilesets using the same ID as a background tileset would have their sound ID set to the default (reported by tobyaaa on Discord), and tilesets using a vanilla tileset ID would be forced to use the vanilla sound. Now the order of preference goes: explicit sound attribute > hardcoded sound index for vanilla tileset ID > fallback 8 for non-vanilla tileset ID.

### DIFF
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -75,7 +75,7 @@ namespace Celeste {
             } else if (xml.HasAttr("sound")) {
                 SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
             } else if (!SurfaceIndex.TileToIndex.ContainsKey(xml.AttrChar("id"))) {
-                SurfaceIndex.TileToIndex[xml.AttrChar("id")] = SurfaceIndex.Brick; // 8 as fallback instead of 0 to match vanilla's default index
+                SurfaceIndex.TileToIndex[xml.AttrChar("id")] = 0; // fall back to no sound
             }
 
             if (xml.HasAttr("debris"))
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -75,7 +75,7 @@ namespace Celeste {
             } else if (xml.HasAttr("sound")) {
                 SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
             } else if (!SurfaceIndex.TileToIndex.ContainsKey(xml.AttrChar("id"))) {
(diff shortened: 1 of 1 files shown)


Kalobi:
A possible bit of further cleanup would be preventing bg tiles from influencing the sound mapping in the first place, but now that only causes issues if someone uses the `sound` attribute in a BackgroundTiles.xml, which I think we can just assume people shouldn't do in the first place, and fixing this would require introducing a new `Autotiler` constructor with an extra parameter.

Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -72,8 +72,10 @@ public patch_Autotiler(string filename)
             if (xml.HasAttr("soundPath") && xml.HasAttr("sound")) { // Could accommodate for no sound attr, but requiring it should improve clarity on user's end 
                 SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
                 patch_SurfaceIndex.IndexToCustomPath[xml.AttrInt("sound")] = (xml.Attr("soundPath").StartsWith("event:/") ? "" : "event:/") + xml.Attr("soundPath");
-            } else if (!SurfaceIndex.TileToIndex.ContainsKey(xml.AttrChar("id")) || xml.AttrChar("id") == 'o') { // Backwards-compat: some existing mods use 'o' and overwrite its sound
-                SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.HasAttr("sound") ? xml.AttrInt("sound") : 8; // 8 as fallback instead of 0 to match vanilla's default index
+            } else if (xml.HasAttr("sound")) {
+                SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
+            } else if (!SurfaceIndex.TileToIndex.ContainsKey(xml.AttrChar("id"))) {
+                SurfaceIndex.TileToIndex[xml.AttrChar("id")] = SurfaceIndex.Brick; // 8 as fallback instead of 0 to match vanilla's default index
```
bigkahuna443: Re: discord discussion, 0 or -1 (no sound) would be a better default, since that makes it consistent with current solid tile behavior. Probably 0 (assuming it doesn't crash) since that's what's on the wiki + in the FMOD project lol. Rest of it looks good though!

Kalobi: Falling back to 0. -1 is used in contexts where there are no tiles below the player, as far as I can tell.

Kalobi:
Drafting for a bit while I look into `k`, the tileset ID for core tiles, which seems to have some special handling.

Kalobi:
Gonna address core tiles separately (see #494 ) and leave this as a fix to return to earlier behavior before feature creep delays it.

bigkahuna443 (approved):

PR #492, opened 2022-08-01
Draw parallax stylegrounds using single Draw call

### LINKED ISSUE #464: render looped parallaxes as one draw call rather than 𝑛×𝑚

no

### Describe your request

currently, looping parallaxes are drawn with nested `for` loops, and one `Texture.Draw` per tile.
this has very poor performance for smaller backdrops, with up to 230×180=41400 calls per backdrop per frame; using [an appropriate TextureAddressMode](https://docs.microsoft.com/en-us/previous-versions/windows/xna/bb975153(v=xnagamestudio.40)) this can be reduced to 1 call per backdrop per frame.
apart from the obvious performance improvement, this would allow for solid-colour backdrops to be reduced to 1 pixel, saving on memory usage and file size.

### Additional context

_No response_

Kalobi:
Previously, looping Parallax stylegrounds were drawn using a nested loop of Draw calls, leading to poor performance especially for small stylegrounds. This PR instead creates a `PointWrap` spritebatch for parallaxes so they can be drawn using a single Draw call. This fixes #464 .
Profiling results for a 1x1 pixel styleground looping on both axes:
Before:
image
After:
image
The time spent in `Parallax.Draw()` has been reduced from 4% to <0.01%.

Submitted as draft because I have done very limited testing so far. Needs to be tested in actual maps and with scroll, flip etc. attributes to make sure it always produces the same result as previously. Other necessary test: `HeatWave.Render` directly calls `Parallax.Render`, but currently does not have the looping spritebatch. It needs to be checked whether that still works or requires additional changes.
Also the IL patches need to be commented more.

### DIFF
--- a/Celeste.Mod.mm/Patches/BackdropRenderer.cs
+++ b/Celeste.Mod.mm/Patches/BackdropRenderer.cs
@@ -26,7 +26,7 @@ namespace Celeste {
 
 namespace MonoMod {
     /// <summary>
-    /// Patches the method to begin a wrapping spritebatch for parallaxes.
+    /// Patches the method to render parallaxes using a wrapping SamplerState.
     /// </summary>
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchBackdropRendererRender))]
     class PatchBackdropRendererRenderAttribute : Attribute { }
@@ -36,6 +36,7 @@ namespace MonoMod {
         public static void PatchBackdropRendererRender(ILContext context, CustomAttribute attrib) {
             MethodReference m_BackDropRenderer_StartSpritebatchLooping = context.Method.DeclaringType.FindMethod("StartSpritebatchLooping");
             TypeReference t_Parallax = context.Module.GetType("Celeste.Parallax");
+            MethodReference n_Parallax_ImprovedRender = t_Parallax.Resolve().FindMethod("ImprovedRender");
             
             ILCursor cursor = new ILCursor(context);
 
@@ -53,6 +54,22 @@ namespace MonoMod {
             cursor.MarkLabel(nextIf);
             cursor.Index--;
             cursor.Emit(OpCodes.Br, nextIf);
+
+            // call ImprovedRender instead of Render for Parallax
+            cursor.GotoNext(instr => instr.MatchLdarg(1), instr => instr.MatchCallvirt("Celeste.Backdrop", "Render"));
+            cursor.Emit(OpCodes.Isinst, t_Parallax);
+            cursor.Emit(OpCodes.Dup);
+            ILLabel parallaxRender = cursor.DefineLabel();
+            cursor.Emit(OpCodes.Brtrue_S, parallaxRender);
+            cursor.Emit(OpCodes.Pop);
+            cursor.Emit(OpCodes.Ldloc_2);
+            cursor.Index += 2;
+            ILLabel continueLoop = cursor.DefineLabel();
+            cursor.Emit(OpCodes.Br, continueLoop);
+            cursor.MarkLabel(parallaxRender);
+            cursor.Emit(OpCodes.Ldarg_1);
+            cursor.Emit(OpCodes.Callvirt, n_Parallax_ImprovedRender);
+            cursor.MarkLabel(continueLoop);
         }
     }
 }
--- a/Celeste.Mod.mm/Patches/BackdropRenderer.cs
+++ b/Celeste.Mod.mm/Patches/BackdropRenderer.cs
@@ -1,5 +1,4 @@
-﻿using Celeste;
-using Microsoft.Xna.Framework.Graphics;
+﻿using Microsoft.Xna.Framework.Graphics;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Monocle;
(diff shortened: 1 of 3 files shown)


Kalobi:
Example of issues that need to be fixed:
image
:widegladeline2:

Kalobi:
Another issue: seems to cut off at integer multiple of texture width for looping parallax.

Kalobi:
After a series of changes, this fix is now only applied to modded stylegrounds, neatly sidestepping any atlas issues. Also, regular `Parallax.Render()` is left unchanged from vanilla, the improved rendering happens in a separate method.

Celeste.Mod.mm/Patches/BackdropRenderer.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchBackdropRendererRender))]
+    class PatchBackdropRendererRenderAttribute : Attribute { }
+
+    static partial class MonoModRules {
+
+        public static void PatchBackdropRendererRender(ILContext context, CustomAttribute attrib) {
+            MethodReference m_BackDropRenderer_StartSpritebatchLooping = context.Method.DeclaringType.FindMethod("StartSpritebatchLooping");
+            TypeReference t_Parallax = context.Module.GetType("Celeste.Parallax");
+            MethodReference n_Parallax_ImprovedRender = t_Parallax.Resolve().FindMethod("ImprovedRender");
```
microlith57: typo?
```suggestion
            MethodReference m_Parallax_ImprovedRender = t_Parallax.Resolve().FindMethod("ImprovedRender");
```

Celeste.Mod.mm/Patches/BackdropRenderer.cs
```diff
@@ -0,0 +1,75 @@
...
+using MonoMod;
+using MonoMod.Cil;
+using MonoMod.Utils;
+using System;
+
+namespace Celeste {
+    class patch_BackdropRenderer : BackdropRenderer {
+        private bool usingSpritebatch;
+        
+        public void StartSpritebatchLooping(BlendState blendState) {
```
microlith57, line 14: this might cause problems if an `Effect` that wants a spritebatch renders after a `Backdrop`, and doesn't expect the spritebatch to be looping; this could be fixed by adding a `private bool spritebatchLooping` and checking that too, but that requires more patching

Celeste.Mod.mm/Patches/Parallax.cs
```diff
@@ -0,0 +1,59 @@
...
+                position.Y = (position.Y % Texture.Height - Texture.Height) % Texture.Height;
+            }
+            SpriteEffects flip = SpriteEffects.None;
+            if (FlipX) {
+                flip |= SpriteEffects.FlipHorizontally;
+            }
+            if (FlipY) {
+                flip |= SpriteEffects.FlipVertically;
+            }
+            Rectangle rect = new Rectangle(0, 0, LoopX ? (int) Math.Ceiling(Celeste.GameWidth - position.X) : Texture.Width, LoopY ? (int) Math.Ceiling(Celeste.GameHeight - position.Y) : Texture.Height);
```
microlith57: more readable with:
```suggestion
            Rectangle rect = new Rectangle(0, 0,
                                           LoopX ? (int) Math.Ceiling(Celeste.GameWidth - position.X) : Texture.Width,
                                           LoopY ? (int) Math.Ceiling(Celeste.GameHeight - position.Y) : Texture.Height);
```

microlith57 (changes requested):

Celeste.Mod.mm/Patches/BackdropRenderer.cs
Kalobi: Maybe I should just MonoModReplace the method at this point

microlith57: yea might as well
would be a good place to fix the bug with additive spritebatches persisting into the next backdrop too

Kalobi: Well it sounded like maps have been made that use that behavior, so I'm not sure how to handle that one.

Kalobi: Also MonoModReplacing definitely increases the probability of max having to rewrite their hooks.

Celeste.Mod.mm/Patches/BackdropRenderer.cs
Kalobi: destareline

Celeste.Mod.mm/Patches/BackdropRenderer.cs
Kalobi: BackdropRenderer.Render logic has now been refactored to account for this.

Kalobi:
Hopefully final state: BackdropRenderer.Render has been replaced to start a PointWrap spritebatch for parallaxes. Parallax.Render has been patched to take advantage of this if possible and fall back to vanilla rendering for vanilla stylegrounds (to avoid atlas wrapping issues) or if the sampler state is incorrect (because Render has been called from somewhere other than the BackdropRenderer).

PR #495, opened 2022-08-04
Add flagSwap decal registry attribute

Kalobi:
This adds a decal registry attribute that allows swapping decal textures based on a flag. It mostly works analogously to the `coreSwap` attribute, except it uses a flag instead of core mode and it can handle animations as well as omitting one of the paths to make a decal only visible on one flag state. It's worth considering making those changes to the `coreSwap` attribute as well, which would probably just mean calling `MakeFlagSwap` with the `cold` flag and retiring `MakeCoreSwap`.


bigkahuna443 (approved):
Make sure to update wiki with the new property (and maybe remove `coreSwap` with a note about how to replicate with `flagSwap`?)

PR #498, opened 2022-08-09
Rewrite CI pipeline (again)

coloursofnoise:
Rewrites the Everest build and release pipelines to make better use of available resources.

- [x] ~~Move all CI-related files to a `.azure-pipelines` subfolder~~ 620b7d2d0d02698c171bb708670ee464b1f8a1a0
- [x] Separate build pipeline into stages for build, verify, and publish
- [x] Use pipeline artifacts instead of build artifacts (deprecated)
- [x] Remove postbuild and verify script files in favour of appropriate tasks and short in-line scripts
- [x] Download vanilla files from azure secure files for security
- [x] Use deployment job for release pipeline
- [x] Set up scheduled releases for beta branch
- [x] Separate stable and beta builds from dev pipeline
- [x] Release beta builds on GitHub as pre-releases
- [ ] Deploy online documentation (online documentation not currently set up)

The following changes will need to be applied through the Azure Devops interface:
- Update pipeline yml path
- Upload vanilla files to secure files
- ~~Set up checks for pull requests to access vanilla files~~
- Add `Common Variables` variable group containing build offset
- Create `Release-Stable` and `Release-Beta` environments

Other possible changes
- Run code style and format checks
- Investigate running patched Celeste in-pipeline to run automated tests

PR #500, opened 2022-08-17
Update Everest updater to align with build pipeline rework

coloursofnoise:
Each branch is now a different source, and is displayed separately in the version list.
This allows for more flexibility in what branches are included, opening the possibility of adding feature branches in the future.

Stable and beta builds are taken from GitHub Releases to align with the changes in #498.

I believe this can be safely merged before merging the pipeline changes, although it might be a good idea to manually create a few beta releases on GitHub to ease the transition.

This PR should also be tested on Windows (and probably Mac just in case) to make sure the in-game Updater still functions as intended.


Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
```diff
@@ -232,27 +229,47 @@ public class Source {
...
 
-                        int id = build["id"].ToObject<int>();
-                        string branch = build["sourceBranch"].ToObject<string>().Replace("refs/heads/", "");
-                        string url = string.Format(artifactFormat, id);
-                        entries.Add(new Entry((id + offset).ToString(), branch, url, id + offset));
+                        string build = Regex.Match(release["name"].ToString(), @"\d+$").Value;
+                        string url = null;
+                        foreach (JObject asset in release["assets"] as JArray) {
+                            if (asset["name"].ToString() == "main.zip") {
+                                url = asset["url"].ToString();
```
DemoJameson: should use `browser_download_url` instead of `url`

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
```diff
@@ -159,14 +136,34 @@ public class Source {
...
+                    Index = "https://api.github.com/repos/EverestAPI/Everest/releases",
+                    ParseData = GitHubReleasesParser(offset: 700)
+                },
+                new Source {
+                    Name = "updater_src_beta",
+                    Description = "updater_src_release_github",
 
-                    ParseData = AzureDataParser("https://dev.azure.com/EverestAPI/Everest/_apis/build/builds/{0}/artifacts?artifactName=main&api-version=5.0&%24format=zip", 700)
-                }
+                    Index = "https://api.github.com/repos/EverestAPI/Everest/releases",
```
DemoJameson: Is there a way to avoid triggering the API request cap?
Get this error when I try accessing this url via browser.

```json
{"message":"API rate limit exceeded for 104.134.74.233. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

DemoJameson (changes requested):

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
WEGFan: > For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the person making requests.

🤔 Maybe we can ask max4805 to fetch releases with access token and cache the results?

maddie480: I think API rate limits is also why Olympus shows a limited amount of commit descriptions in the versions list 🤔 I could definitely fetch and cache the list on my end, the only issue would be that there would be a small delay before a new release appears

WEGFan: > There would be a small delay before a new release appears.

We can use webhooks, also the limit is 5000 requests per hour with access tokens so I think we can safely increase fetching frequency.

maddie480: We can either have an API that gives [a preformatted list of versions](https://cdn.discordapp.com/attachments/429775439423209472/1010647544541499512/a.json) that could be used by Everest and Olympus to avoid maintaining the logic twice, or just have an API that returns the releases as-is to keep all logic client-side, what do we prefer? 🤔

maddie480: https://max480-random-stuff.appspot.com/celeste/everest-versions ^^

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
coloursofnoise: This has been changed, although the GitHub Release Parser is no longer used for the main Everest branches.

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
coloursofnoise: Has been implemented using https://everestapi.github.io/everestupdater.txt to link to the update list source

Celeste.Mod.mm/Mod/Everest/Everest.Updater.cs
```diff
@@ -277,6 +314,7 @@ public class Source {
                     return;
                 }
 
+                CoreModule.Settings.CurrentBranch = version.Source.Name;
```
DemoJameson, line 319: This setting is not actually written to the file
perhaps move it until after the installation succeeds and call `CoreModule.Instance.SaveSettings()` manually

DemoJameson (changes requested):

DemoJameson (approved):

PR #503, opened 2022-08-21
Make metadata loading consistent; closes #438

### LINKED ISSUE #438: Metadata in map file overrides .meta.yaml/general need for refactoring in AreaData.Load()

Yes

Yes

### Describe the bug

When the same metadata option is defined in both the map file via Ahorn and in the .meta.yaml file, the one in the map file takes precedence.

### Steps to reproduce

Example: set Title Text Color to 00ff00 in Ahorn, add `TitleTextColor: ff0000` to the .meta.yaml. The map title will be green.

### Expected behavior

The .meta.yaml should take precedence over the map file. Cruor confirmed this to be the expected behavior, and it makes more sense because default values set by a map editor can be hard to remove from the map file, while the .meta.yaml is trivial to edit.

### Operating System

Windows 10

### Everest Version

3324

_No response_

### Additional context

The .meta.yaml is parsed in the big loop over the content map in `AreaData.Load()`. The parsing of the in-map file metadata happens later when the `MapData(AreaKey)` constructor is called, via `MapData.Load()`->`MapData.orig_Load()`->`MapData._Process()`->`MapData.Process()`->`MapData.ProcessMeta()`.

While tracking this down, I also found several other things in `AreaData.Load()` that seem like they should be refactored:

- The loop over the content map parses each map's .meta.yaml twice, apparently because of some kind of shadow structure assets that are not being ignored? Duplicates are eliminated afterwards, but this feels like an inelegant solution.
- The comment `// Remove AreaDatas which are now a mode of another AreaData.
            // This can happen late as the map data (.bin) can contain additional metadata.`
   appears to suggest that the section after it should run after the map data has been parsed so that it knows about said additional metadata, but it happens before any .bins are parsed.
- The first line of the `area.OnLevelBegin` callback at the end of the content map loop looks like a no-op to me. If if is not, it could use a comment explaining why it's necessary

Unfortunately, I do not understand the area loading process well enough yet to fix this without breaking everything, although I will probably continue to look into it. Help would be much appreciated.

rhelmot:
This smooths over the rough edges with respect to how metadata is loaded between yaml and bin files.

The order of precedence is (lowest to highest):

- you can specify both Mode and Modes in a yaml file - the Mode is a single set of mode mode properties, and the Modes is a list of one-propset-per-mode. The former will override the latter.
- a yaml file will override a bin file.
- Side-specific config will override the a-side config.


Kalobi:
Re: the third point, how does this interact with the existing Override A-Side Meta option?

rhelmot:
It still functions the same way as before - the thing I wrote only applies to how the MapMeta which is applied to a given map via SetMapMeta is generated. For a mode other than the a-side, the only field that will be used is the appropriate index of the Modes array, unless override a-side meta is set, in which case it will copy the relevant fields of the config over to the Area on start.

Noting the third point is important because you can specify conflicting information related to just one mode - in the a-side meta  you can set the second entry of the Modes array, and then in the b-side meta you can set the same thing. This is obviously very stupid, but the latter should override the former in this case.

rhelmot:
Any updates? What kind of testing or discussion is needed before this can be merged?

Kalobi:
It should probably be tested on mods with and without multiple sides and with different combinations of .meta.yaml and embedded metadata to check that it always produces the desired results. I'm assuming you've done at least some part of that.
I don't really understand map loading well enough to comprehensively review the code changes. I'd be down to just merge this and wait for people to scream if it breaks something, but that's probably not a great approach for one of the most important methods in Everest.

Celeste.Mod.mm/Patches/MapEditor.cs
```diff
@@ -85,7 +85,7 @@ public patch_MapEditor(AreaKey area, bool reloadMapData = true)
             }
 
             session.Level = level.Name;
-            session.RespawnPoint = customRespawnPoint ? at : null;
+            session.RespawnPoint = customRespawnPoint ? at : new Vector2?();
```
rhelmot: I wish I knew why this was necessary, but dotnet refused to compile everest until I made this change.

Celeste.Mod.mm/Patches/MapData.cs
```diff
@@ -175,8 +180,35 @@ public patch_MapData(AreaKey area)
             AreaData area = AreaData.Get(Area);
             AreaMode mode = Area.Mode;
 
+            MapMeta metaParsedFromFile = null;
```
rhelmot, line 194: This is the core of the PR's logic. This should, ideally, be the _only_ place we pull a MapMeta out of the everest asset store: in the middle of MapData parsing.

rhelmot: In service of this, the method that calls this method was patched to call this method even if the map has no meta node in its packed data, so meta may be null here.

Celeste.Mod.mm/Patches/MapData.cs
```diff
@@ -189,20 +221,10 @@ public patch_MapData(AreaKey area)
...
-
-            // Metadata for B-Side and C-Side are parsed and stored.
-            if (mode != AreaMode.Normal) {
-                MapMeta mapMeta = new MapMeta(meta) {
-                    Modes = area.GetMeta().Modes
-                };
-                area.Mode[(int) mode].SetMapMeta(mapMeta);
+            } else {
+                MapMeta combinedMeta = MapMeta.Add(metaParsed, area.GetMeta());
+                area.Mode[(int)mode].SetMapMeta(combinedMeta);
```
rhelmot: Hmmm.... I wonder if there should be a SetMapMeta call in the Normal branch as well. I can't figure out where the map mode properties meta is used.

rhelmot: It looks like there's two places mode meta is stored - on the maps themselves and in the modes property of the area meta. Should we try to stick to one of them, e.g. by following the same contract that I made up for the new mode property where MapMeta.modes will be null outside of the loading context?

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -132,9 +133,83 @@ public class MapMeta : IMeta {
                     Modes[i] = null;
                 }
             }
+            BinaryPacker.Element modeMeta = meta.Children?.FirstOrDefault(el => el.Name == "mode");
+            if (modeMeta != null) {
+                Mode = new MapMetaModeProperties(modeMeta);
+            }
+        }
+
+        public void AddTo(MapMeta other) {
```
rhelmot, line 150: AddTo is the main utility method I add - it does the merging between two MapMetas. It mutates other, overriding any properties which are set by `this`.

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -132,9 +133,83 @@ public class MapMeta : IMeta {
...
+                }
+            }
+            if (Mountain != null) { other.Mountain = Mountain; }
+            if (CompleteScreen != null) { other.CompleteScreen = CompleteScreen; }
+            if (LoadingVignetteScreen != null) { other.LoadingVignetteScreen = LoadingVignetteScreen; }
+            if (LoadingVignetteText != null) { other.LoadingVignetteText = LoadingVignetteText; }
+            if (CassetteModifier != null) { other.CassetteModifier = CassetteModifier; }
+        }
+
+        public static MapMeta Add(MapMeta self, MapMeta other) {
```
rhelmot, line 196: Add is like AddTo but it is static and works correctly if either of the metas are null.

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -192,59 +267,7 @@ public class MapMeta : IMeta {
...
             area.MountainState = Mountain?.State ?? area.MountainState;
 
-            ModeProperties[] modes = area.Mode;
-            area.Mode = Convert(Modes) ?? modes;
-            if (modes != null)
-                for (int i = 0; i < area.Mode.Length && i < modes.Length; i++)
-                    if (area.Mode[i] == null)
-                        area.Mode[i] = modes[i];
-
-            MapMeta meta = area.GetMeta();
```
rhelmot: This is what previously did the merging that AddTo does.

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -382,6 +405,38 @@ public ModeProperties Convert()
             }
             area.Mode[(int) mode] = props;
         }
+
+        public void AddTo(MapMetaModeProperties other) {
```
rhelmot, line 409: And and AddTo are also implemented for MapMetaModeProperties and MapMetaAudioState. It could theoretically also be implemented for the further-down meta classes, such as checkpoint meta, but those are never serialized into bins by any known level editor so it's not a priority.

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -411,13 +471,13 @@ public class MapMetaCheckpointData {
 
         public string Level { get; set; }
         public string Name { get; set; }
-        public bool Dreaming { get; set; }
+        public bool? Dreaming { get; set; }
```
rhelmot, line 477: Making this a bool lets it be overridden

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -443,6 +503,16 @@ public CheckpointData Convert()
                 }
             }
         }
+
+        public void AddTo(MapMetaCheckpointData other) {
```
rhelmot, line 510: Oh, right. I implemented AddTo for checkpoint data before I realized it was useless. Maybe in the future it will not be useless!

Celeste.Mod.mm/Patches/AreaData.cs
```diff
@@ -246,28 +246,11 @@ class patch_AreaData : AreaData {
                     area.CassseteNoteColor = Calc.HexToColor("33a9ee");
                     area.CassetteSong = SFX.cas_01_forsaken_city;
 
-                    // Custom values can be set via the MapMeta.
-                    MapMeta meta = new MapMeta();
-                    meta.ApplyTo(area);
```
rhelmot, line 259: This is the change I'm the most worried about. By changing the location of meta.ApplyTo from here to whenever maps are actually parsed, I'm worried some accesses could get lost. I cannot find where maps are actually loaded in order to check, though.

Celeste.Mod.mm/Patches/MapEditor.cs
Kalobi: You might be able to make it `default`.

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -132,9 +133,83 @@ public class MapMeta : IMeta {
...
+                return result;
+            }
         }
 
         public void ApplyTo(AreaData area) {
+            // Some of the game's code checks for [1] / [2] explicitly.
+            // Let's just provide null modes to fill any gaps.
+            Modes = Modes ?? new MapMetaModeProperties[3];
+            if (Modes.Length < 3) {
+                MapMetaModeProperties[] larger = new MapMetaModeProperties[3];
```
l-Luna: why not use `Array.Resize(ref Modes, 3)`?

l-Luna:
will test this sometime

maddie480:
needs updating from master :sweat_smile:

PR #506, opened 2022-08-23
Allow 'log.txt' filename to be changed via EVEREST_LOG_FILENAME= in env

RedFlames:
Default behavior is unchanged and logs to `log.txt` with the existing log rotator.

I want to be able to change where logs are written to so that e.g. I can run multiple Itch.io Celestes for Celestenet testing and have them write individual logs.

When putting this in `everest-env.txt`
```ini
EVEREST_LOG_FILENAME=my_custom_log.txt
```

or launching with `EVEREST_LOG_FILENAME=mylog.txt ./Celeste.exe` etc. the specified filename will be used as a log file instead. 
The log rotation will not kick in, since that was only built to turn `log.txt` into `log_[date+time].txt` log backups, no custom names there unless someone else wants to implement that.

The filename is checked against `Path.GetInvalidFileNameChars()` which hopefully returns something sensible on every OS we run on, and unlike `Path.GetInvalidPathChars()` it will not allow any paths (`/` and `\` not allowed).

If the name doesn't already end in `.txt`, that will be appended just to be safe.

Using an invalid name, the game will fail to launch:
```ini
EVEREST_LOG_FILENAME=../a"t?est.txt ./Celeste.exe
```
```
08/25/2022 21:37:59
System.ArgumentException: Custom log filename set in EVEREST_LOG_FILENAME="../a"t?est.txt" contains invalid character(s): / " ?
   at Celeste.Celeste.Main(String[] args) in C:\Users\rf\Documents\code\Everest\Celeste.Mod.mm\Patches\Celeste.cs:line 32
   at Celeste.Mod.BOOT.Main(String[] args) in C:\Users\rf\Documents\code\Everest\Celeste.Mod.mm\Mod\Everest\BOOT.cs:line 0
```

---
**Edit: Old message below, behavior has been changed.**

---

Should this test if the specified file is actually a writeable file path and if it isn't, should it fall back to `log.txt`? 

~~Currently it will crash, like so:~~
```
Monocle Engine Error Log
==========================================

08/23/2022 23:05:22
System.UnauthorizedAccessException: Access to the path 'C:\mylog.txt' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at Celeste.Celeste.Main(String[] args) in C:\Users\rf\Documents\code\Everest\Celeste.Mod.mm\Patches\Celeste.cs:line 30
   at Celeste.Mod.BOOT.Main(String[] args) in C:\Users\rf\Documents\code\Everest\Celeste.Mod.mm\Mod\Everest\BOOT.cs:line 0
```

~~I just realized this does allow writing to arbitrary files with absolute paths or `../../`, is this any concern of ours? Should this be restricted to filenames with no paths?~~

### DIFF
--- a/Celeste.Mod.mm/Patches/Celeste.cs
+++ b/Celeste.Mod.mm/Patches/Celeste.cs
@@ -17,6 +17,8 @@ using System.Threading;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using MonoMod.Cil;
+using System.Text.RegularExpressions;
+using System.Text;
 
 namespace Celeste {
     class patch_Celeste : Celeste {
@@ -124,7 +126,7 @@ namespace Celeste {
                 return;
             }
 
-            string logfile = Environment.GetEnvironmentVariable("EVEREST_LOGFILE") ?? "log.txt";
+            string logfile = Environment.GetEnvironmentVariable("EVEREST_LOG_FILENAME") ?? "log.txt";
 
             // Only applying log rotation on default name, feel free to improve LogRotationHelper to deal with custom log file names...
             if (logfile == "log.txt" && File.Exists("log.txt")) {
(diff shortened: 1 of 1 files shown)


Kalobi:
I haven't looked at the code in detail, but is there any reason this is an environment variable and not just a command line argument?

RedFlames:
I kind of just... checked what options there were for doing this, and decided to go with Env because it's easy to throw an env var at something launched from bash. Should it be a command line argument instead?

coloursofnoise:
For posterity, this will be kept as an environment variable since it has more use than a command line option (can be set "globally").
Another PR can be created if adding a command line option is deemed necessary.

Issue #479, opened 2022-07-25
Improve level error handling


bigkahuna443:
### I am planning on implementing this myself

Yes

### Describe your request

Currently maps hard crash if you encounter an exception within them, _except_ when entering from `LevelEnter`, e.g. loading a map from a new savefile. In that case, it opens a postcard informing you that an exception has occurred and returns you to map. I want to look into:
1. Moving this "soft-crash" functionality so it gets called in more situations. Looking first into exceptions that occur during level load, ideally others as well.
2. Improving logs in places that hard-crash with non-descriptive exceptions and stack traces, e.g. autotiler-related crashes.
3. Potentially manually invoking the soft-crash features with specific messages for common crashes.

bigkahuna443:
- [x] Add postcard error handling to LevelLoader
- [x] Make postcard errors more usable by mods
- [x] Add descriptive logs and/or postcard error messages for common crashes
- [ ] Improve error handling for crashes not covered by the new exception handlers

PR #508, opened 2022-08-27
Improve error postcards

bigkahuna443:
Partially addresses #479.

- Simplifies error postcard logic in `LevelEnter`
  - Cleans up usage in `Level` and `AssetReloadHelper` to match changes
- Adds error handling to `LevelLoader`

Mods now only have to set `LevelEnter.ErrorMessage` to activate an error postcard if they're inside a Level, although currently full exception handling is only being done for level loading. Due to how the game loop works, we would have to wrap each scene function (Update, Render, etc.) with try-catches if we wanted to handle everything.

I'll be working on adding unique postcard messages and improving logs for common errors after this is merged.


microlith57 (commented):
:+1:

PR #513, opened 2022-08-30
Fix automatic strawberry order assignment

Kalobi:
Previously, berries with strawberry order -1 were blindly assigned to the current number of encountered strawberries, leading to possible collisions. Now, they will always be assigned unused indices, making it viable to mix -1 with explicitly assigned indices. This does not address assigning duplicate indices manually (may still cause crashes) or leaving gaps in the indices (breaks `GameplayStats`).

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreMapDataProcessor.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreMapDataProcessor.cs
@@ -7,7 +7,8 @@ namespace Celeste.Mod.Core {
     public class CoreMapDataProcessor : EverestMapDataProcessor {
 
         public int Checkpoint;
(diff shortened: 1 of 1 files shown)


Kalobi:
Duplicates and gaps are now addressed. Berry order within each checkpoint is determined the following way:
First all berries with explicitly specified position will be placed there, while berries with order -1 will be saved for later. If a berry with a duplicate index is found, it is treated as if it had -1 and a warning will be logged.
Then all of the -1 berries are used, in processing order, to fill the gaps in the order.
If any gaps remain in the order after that, berries will be sifted down to fill the gaps, and a warning will be logged.

GreyMario (approved):

PR #514, opened 2022-09-01
Restructure Decal Registry

bigkahuna443:
1. Use content map instead of gameplay atlas texture list. This fixes a bug where we can't register directory/wildcard paths if the registry is ingested before the textures are (see Discord discussion [here](https://discord.com/channels/403698615446536203/1014018168047538196/1014052229755441242)).
2. Limit registered decals to those within the source mod. This reduces conflicts and lets us reload individual registries instead of having to regenerate the global registry.
3. Rework interfaces and add documentation. Restricting access to the internal implementation will make it easier to make changes in the future.

To-Do: reloading a mod's registry should unload its registered paths before re-applying the registry. Not in this PR since this one was mainly to fix the ingest bug + get parity with the old version (which didn't unload old registrations either).

### DIFF
--- a/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
+++ b/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
 using Monocle;
 using System;
 using System.Collections.Generic;
@@ -12,9 +13,15 @@ namespace Celeste.Mod {
     /// Allows custom decals to have properties that are otherwise hardcoded, such as reflections, parallax, etc.
     /// </summary>
     public static class DecalRegistry {
+        /// <summary>
+        /// Mapping of decal paths to decal registry properties.
+        /// </summary>
+        public static readonly Dictionary<string, DecalInfo> RegisteredDecals = new Dictionary<string, DecalInfo>();
 
-        // string is propertyName
-        public static Dictionary<string, Action<Decal, XmlAttributeCollection>> PropertyHandlers = new Dictionary<string, Action<Decal, XmlAttributeCollection>>() {
+        /// <summary>
+        /// Everest-defined decal registry properties. See <see cref="AddPropertyHandler"/> to add a custom property.
+        /// </summary>
+        internal static Dictionary<string, Action<Decal, XmlAttributeCollection>> PropertyHandlers = new Dictionary<string, Action<Decal, XmlAttributeCollection>>() {
             { "parallax", delegate(Decal decal, XmlAttributeCollection attrs) {
                 if (attrs["amount"] != null)
                     ((patch_Decal)decal).MakeParallax(float.Parse(attrs["amount"].Value));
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
```
microlith57: this (and `System.IO`) are unused:
```suggestion
```
(github won't let me suggest the other deletion)

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -12,9 +13,15 @@ namespace Celeste.Mod {
...
+        /// Mapping of decal paths to decal registry properties.
+        /// </summary>
+        public static readonly Dictionary<string, DecalInfo> RegisteredDecals = new Dictionary<string, DecalInfo>();
 
-        // string is propertyName
-        public static Dictionary<string, Action<Decal, XmlAttributeCollection>> PropertyHandlers = new Dictionary<string, Action<Decal, XmlAttributeCollection>>() {
+        /// <summary>
+        /// Everest-defined decal registry properties. See <see cref="AddPropertyHandler"/> to add a custom property.
+        /// </summary>
+        internal static Dictionary<string, Action<Decal, XmlAttributeCollection>> PropertyHandlers = new Dictionary<string, Action<Decal, XmlAttributeCollection>>() {
```
microlith57, line 22: can we check if any mods relying on this being `public`?

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -162,15 +169,40 @@ public static class DecalRegistry {
             }},
         };
 
-        // Helper functions for scaling Decal Registry fields
+        /// <summary>
+        /// Adds a custom property to the decal registry. See <see cref="PropertyHandlers"/> for the list of Everest-defined properties.
+        /// </summary>
+        public static void AddPropertyHandler(string propertyName, Action<Decal, XmlAttributeCollection> action) {
+            if (PropertyHandlers.ContainsKey(propertyName)) {
```
microlith57, line 174: what happens if a mod wants to patch an existing property handler, eg. to turn off all parallax decals? could this warning be optionally silenced?

microlith57 (changes requested):
:+1:

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
bigkahuna443: that's exactly why we want this warning in place. someone overriding Everest/another mod's property affects any mods that use it so it's not recommended and we don't want to do that silently

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
bigkahuna443: I did a GitHub wide search for all of the interface changes + tested it with SJ which is about as thorough as I can get haha

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -162,15 +167,40 @@ public static class DecalRegistry {
...
+                string asmName = Assembly.GetCallingAssembly().GetName().Name;
+                string modName = Everest.Content.Mods
+                                 .Where(mod => mod is AssemblyModContent && mod.DefaultName == asmName)
+                                 .FirstOrDefault()?.Name;
```
Kalobi:
```suggestion
               string modName = Everest.Content.Mods
                                 .FirstOrDefault(mod => mod is AssemblyModContent && mod.DefaultName == asmName)?.Name;
```
Courtesy of my IDE

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -193,47 +226,80 @@ public static class DecalRegistry {
...
-                PropertyHandlers[propertyName] = action;
-            } else {
-                PropertyHandlers.Add(propertyName, action);
+        /// <summary>
+        /// Loads a mod's decal registry file.
+        /// </summary>
+        internal static void LoadModDecalRegistry(ModAsset decalRegistry) {
+            Logger.Log(LogLevel.Debug, "Decal Registry", $"Loading registry for {decalRegistry.Source.Name}");
+
+            char[] digits = new[] { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' };
```
Kalobi:
```suggestion
            char[] digits = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' };
```

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -193,47 +226,80 @@ public static class DecalRegistry {
...
 
             List<KeyValuePair<string, DecalInfo>> elements = new();
-            foreach (XmlNode node in file) {
+
+            foreach (XmlNode node in doc["decals"]) {
                 if (node is XmlElement decal) {
                     string decalPath = decal.Attr("path", null)?.ToLower();
                     if (decalPath == null) {
-                        Logger.Log(LogLevel.Warn, "Decal Registry", "Decal didn't have a path attribute!");
+                        Logger.Log(LogLevel.Warn, "Decal Registry", $"Decal in {decalRegistry.Source.Name} registry didn't have a path attribute!");
```
Kalobi: While we're changing the error message, this sounds slightly strange. Maybe "is missing a path attribute" instead of "didn't have"?

Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
```diff
@@ -193,47 +226,80 @@ public static class DecalRegistry {
...
+        private static List<KeyValuePair<string, DecalInfo>> ReadDecalRegistryXml(ModAsset decalRegistry) {
             XmlDocument doc = new XmlDocument();
-            doc.LoadXml(fileContents);
-            XmlElement file = doc["decals"];
+            doc.Load(decalRegistry.Stream);
 
             List<KeyValuePair<string, DecalInfo>> elements = new();
-            foreach (XmlNode node in file) {
+
+            foreach (XmlNode node in doc["decals"]) {
```
Kalobi: Should we check for the absence of this key and print a more informative error message?

Kalobi (approved):

PR #515, opened 2022-09-04
allow decals to be rotated

### LINKED ISSUE #467: decal rotation

Yes

### Describe your request

currently, decals cannot be rotated, only flipped; allowing them to be rotated to arbitrary angles could be very helpful for mappers.
this would require ahorn/lönn support for the rotation attribute and for keybinds to rotate by 90° increments.

### Additional context

i am implementing this currently

microlith57:
allows a `rotation` float (in radians) to be added to decals in the map data, and rotates decals accordingly.

closes #467

### DIFF
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -56,8 +56,8 @@ namespace Celeste {
                 // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
             }
 
-            [MonoModIgnore] // We don't want to change anything about the method...
-            [PatchDecalImageRotation] // ... except for manually manipulating the method via MonoModRules
+            [MonoModIgnore]
+            [PatchDecalImageRotation]
             public extern override void Render();
 
         }
@@ -69,8 +69,8 @@ namespace Celeste {
                 // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
             }
 
-            [MonoModIgnore] // We don't want to change anything about the method...
-            [PatchDecalImageRotation] // ... except for manually manipulating the method via MonoModRules
+            [MonoModIgnore]
+            [PatchDecalImageRotation]
             public extern override void Render();
 
         }
@@ -81,8 +81,8 @@ namespace Celeste {
                 // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
             }
 
-            [MonoModIgnore] // We don't want to change anything about the method...
-            [PatchDecalImageRotation] // ... except for manually manipulating the method via MonoModRules
+            [MonoModIgnore]
+            [PatchDecalImageRotation]
             public extern override void Render();
 
         }
@@ -140,70 +140,18 @@ namespace Celeste {
         [MonoModPublic]
         public extern void CreateSmoke(Vector2 offset, bool inbg);
 
-        [MonoModReplace]
+        [MonoModIgnore]
         [MonoModPublic]
-        public void MakeMirror(string path, bool keepOffsetsClose) {
-            if (keepOffsetsClose) {
-                MakeMirror(path, GetMirrorOffset());
-                return;
-            }
-
-            Depth = 9500;
-            foreach (MTexture mask in GFX.Game.GetAtlasSubtextures("mirrormasks/" + path)) {
-                MirrorSurface surface = new MirrorSurface();
-                surface.ReflectionOffset = GetMirrorOffset();
-                surface.OnRender = delegate {
-                    mask.DrawCentered(Position, surface.ReflectionColor, scale, Rotation);
-                };
-                Add(surface);
-            }
-        }
-
-        [MonoModReplace]
-        [MonoModPublic]
-        public void MakeMirror(string path, Vector2 offset) {
-            Depth = 9500;
-            foreach (MTexture mask in GFX.Game.GetAtlasSubtextures("mirrormasks/" + path)) {
-                MirrorSurface surface = new MirrorSurface();
-                surface.ReflectionOffset = offset + new Vector2(-2f + Calc.Random.NextFloat(4f), -2f + Calc.Random.NextFloat(4f));
-                surface.OnRender = delegate {
-                    mask.DrawCentered(Position, surface.ReflectionColor, scale, Rotation);
-                };
-                Add(surface);
-            }
-        }
-
-        [MonoModReplace]
-        [MonoModPublic]
-        // only used for that one big crystal in reflection
-        public void MakeMirrorSpecialCase(string path, Vector2 offset) {
-            Depth = 9500;
-            List<MTexture> masks = GFX.Game.GetAtlasSubtextures("mirrormasks/" + path);
-            for (int mask_num = 0; mask_num < masks.Count; mask_num++) {
-                Vector2 mini_offset;
-                switch (mask_num) {
-                    case 2:
-                        mini_offset = new Vector2(4f, 2f);
-                        break;
-                    case 6:
-                        mini_offset = new Vector2(-2f, 0f);
-                        break;
-                    default:
-                        mini_offset = new Vector2(-2f + Calc.Random.NextFloat(4f), -2f + Calc.Random.NextFloat(4f));
-                        break;
-                }
-                MTexture mask = masks[mask_num];
-                MirrorSurface surface = new MirrorSurface();
-                surface.ReflectionOffset = offset + mini_offset;
-                surface.OnRender = delegate {
-                    mask.DrawCentered(Position, surface.ReflectionColor, scale, Rotation);
-                };
-                Add(surface);
-            }
-        }
+        [PatchMirrorMaskRotation]
+        public extern void MakeMirror(string path, bool keepOffsetsClose);
 
         [MonoModIgnore]
-        private extern Vector2 GetMirrorOffset();
+        [PatchMirrorMaskRotation]
+        private extern void MakeMirror(string path, Vector2 offset);
+
+        [MonoModIgnore]
+        [PatchMirrorMaskRotation]
+        private extern void MakeMirrorSpecialCase(string path, Vector2 offset);
 
         [MonoModIgnore]
         [MonoModPublic]
@@ -362,7 +310,6 @@ namespace Celeste {
             => ((patch_Decal) self).Scale;
         public static void SetScale(this Decal self, Vector2 value)
             => ((patch_Decal) self).Scale = value;
-
     }
 }
 
@@ -378,6 +325,12 @@ namespace MonoMod {
     /// </summary>
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchDecalImageRotation))]
     class PatchDecalImageRotationAttribute : Attribute { }
+
+    /// <summary>
+    /// Allow mirror masks to be rotated.
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchMirrorMaskRotation))]
+    class PatchMirrorMaskRotationAttribute : Attribute { }
 
     static partial class MonoModRules {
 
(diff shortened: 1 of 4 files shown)


Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -45,22 +48,43 @@ public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
...
+            public override void Render() {
+                (((base.Scene as Level).CoreMode == Session.CoreModes.Cold) ? cold : hot).DrawCentered(
+                    ((patch_Decal) Entity).Position,
+                    Color.White,
+                    ((patch_Decal) Entity).scale,
+                    ((patch_Decal) Entity).Rotation);
             }
         }
 
-        [MonoModIgnore]
```
microlith57, line 55: maybe `patch_DecalImage` should be above `patch_CoreSwapImage`?

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -518,6 +519,12 @@ namespace MonoMod {
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchLevelLoader))]
     class PatchLevelLoaderAttribute : Attribute { }
 
+    /// <summary>
+    /// Patch the Godzilla-sized level loading method instead of reimplementing it in Everest.
```
microlith57, line 523: should have a better comment

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -45,22 +48,79 @@ public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
...
+                MTexture tex = decal.textures[(int) decal.frame];
+                List<MTexture> segments = Segments[(int) decal.frame];
+                for (int i = 0; i < segments.Count; i++) {
+                    float num = (EaseDown ? ((float) i / (float) segments.Count) : (1f - (float) i / (float) segments.Count)) * WindMultiplier;
+                    float x = (float) (Math.Sin(sineTimer * WaveSpeed + (float) i * SliceSinIncrement) * (double) num * (double) WaveAmplitude + (double) (num * Offset));
+                    segments[i].Draw(position: decal.Position + new Vector2(x, 0f),
+                                 origin: new Vector2(tex.Width / 2, tex.Height / 2 - i * SliceSize),
+                                 Color.White,
+                                 decal.scale,
+                                 decal.Rotation);
```
bigkahuna443: I think this syntax of one line per item makes sense with something like LINQ where each line is a different operation, but it looks a bit odd here. Might try condensing it to e.g. have the "origin" field on line 78 and the rest of the fields on line 79. applies elsewhere too

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -45,22 +48,79 @@ public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
...
+            public int SliceSize;
+            public float SliceSinIncrement;
+            public bool EaseDown;
+            public float Offset;
+            public float WindMultiplier;
+            private float sineTimer;
+            public List<List<MTexture>> Segments;
+#pragma warning restore CS0649
+
+            private patch_Decal decal => (patch_Decal) base.Entity;
```
bigkahuna443: This property is public in the original class, so this patch causes a new property to be created

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -45,22 +48,79 @@ public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
...
-            public DecalImage() : base(active: false, visible: true) {
-                // no-op
+        private class patch_DecalImage : Component {
+            public patch_DecalImage()
+                : base(active: false, visible: true) {
+                // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
+            }
+
+            public extern void orig_Render();
+            public override void Render() {
```
bigkahuna443: you're not using the orig method for anything so you can just `[MonoModReplace]` this instead

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -45,22 +48,79 @@ public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
...
+                ((patch_Decal) Entity).textures[(int) ((patch_Decal) Entity).frame].DrawCentered(
```
bigkahuna443: the component already has a `Decal` field to get the decal which you can use instead of casting `Entity`, makes IL a bit cleaner

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -82,7 +142,11 @@ private class FlagSwapImage : Component {
 
             public override void Render() {
                 if (activeTextures.Count > 0)
-                    activeTextures[(int) frame % activeTextures.Count].DrawCentered(Decal.Position, Color.White, Decal.scale);
+                    activeTextures[(int) frame % activeTextures.Count].DrawCentered(
```
bigkahuna443: yeah there are longer lines than this already, you could just have it be on one lol

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -45,22 +48,79 @@ public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
             // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
         }
 
-        [MonoModIgnore]
-        private class CoreSwapImage : Component {
-            public CoreSwapImage(MTexture hot, MTexture cold) : base(active: false, visible: true) {
-                // no-op
+        private class patch_Banner : Component {
```
bigkahuna443, line 52: Overall comment: it's a little unfortunate to have to add all these replaced functions. Looking at the IL it looks like you could get away with some shared IL patches. Just move before the `Draw` or `DrawCentered` call -> load the `Rotation` field -> call the function that takes in the rotation -> pop the old one. Would be a lot cleaner if it works.

(left some other comments in case you keep the original implementation though)

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -45,22 +48,79 @@ public patch_Decal(string texture, Vector2 position, Vector2 scale, int depth)
...
+                : base(active: false, visible: true) {
+                // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
+            }
+
+            [MonoModReplace]
+            public override void Render() {
+                MTexture tex = decal.textures[(int) decal.frame];
+                List<MTexture> segments = Segments[(int) decal.frame];
+                for (int i = 0; i < segments.Count; i++) {
+                    float num = (EaseDown ? ((float) i / (float) segments.Count) : (1f - (float) i / (float) segments.Count)) * WindMultiplier;
```
bigkahuna443: there are some unnecessary casts here from the decompilation that my IDE catches

Celeste.Mod.mm/Patches/DecalData.cs
```diff
@@ -0,0 +1,9 @@
+namespace Celeste {
+    class patch_DecalData : DecalData {
+
+#pragma warning disable CS0649 // The field is never assigned and will always be null
```
bigkahuna443: Not a huge deal but I think these are cleaner as a global disable at the top of the file

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -37,7 +37,7 @@ class patch_Level : Level {
 
         public delegate Entity EntityLoader(Level level, LevelData levelData, Vector2 offset, EntityData entityData);
         public static readonly Dictionary<string, EntityLoader> EntityLoaders = new Dictionary<string, EntityLoader>();
-
```
bigkahuna443, line 40: whitespace change 😱

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -262,7 +388,6 @@ private class FlagSwapImage : Component {
     }
 
     public static class DecalExt {
-
```
bigkahuna443: I think it's actually supposed to be convention to have a blank line at the beginning of a class lol

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -186,7 +186,8 @@ class patch_Level : Level {
             }
         }
 
-        [PatchLevelLoader] // Manually manipulate the method via MonoModRules
+        [PatchLevelLoaderDecalCreation] // Manually manipulate the method via MonoModRules
+        [PatchLevelLoader]
```
bigkahuna443: I took a look and only found one other usage like this in Everest... It's definitely easier to read patches when they're split up, but can also create issues if they're order-dependent or one changes IL the other depends on. Doesn't seem like there's a conflict here though and I agree that it's out of scope, we should probably try to combine them if/when we refactor the old patches though

EDIT: nitpicky but if you put the new attribute below the old one it'd be a 1-line patch instead of 3 smh

Celeste.Mod.mm/Patches/Level.cs
bigkahuna443: would prefer a more useful comment here

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -639,6 +646,45 @@ static partial class MonoModRules {
...
+            FieldDefinition f_DecalRotation = Decal.FindField("Rotation");
+            MethodReference m_Decal_ctor = Decal.FindMethod(".ctor");
+
+            ILCursor cursor = new ILCursor(context);
+
+            int matches = 0;
+            while (cursor.TryGotoNext(MoveType.Before,
+                                      // we are looking for one of the possible decal depths (fg or bg)...
+                                      instr => instr.MatchLdcI4(-10500)
+                                            || instr.MatchLdcI4(9000),
```
bigkahuna443: you can use `Celeste.Depths.FGDecals` and `Celeste.Depths.BGDecals` for these. could make more sense to add a new Decal constructor though and swap it out instead of adding it to the existing reference?

Celeste.Mod.mm/Patches/LevelData.cs
```diff
@@ -63,5 +70,61 @@ we found
...
+                IL_0a76: stfld string Celeste.DecalData::Texture
+                     ->  dup
+                         ldloc.s 11
+                         ldstr "rotation"
+                         ldc_r4 0.0
+                         callvirt instance float32 Celeste.BinaryPacker/Element::AttrFloat(string, float32)
+                         stfld float32 Celeste.DecalData::Rotation
+                IL_0a7b: callvirt instance void class [mscorlib]System.Collections.Generic.List`1<class Celeste.DecalData>::Add(!0)
+
+                in both places where a DecalData instance is added to a List
```
bigkahuna443: I definitely think a C# comment would be more helpful here than IL, we can already see the IL below

Celeste.Mod.mm/Patches/LevelData.cs
```diff
@@ -63,5 +70,61 @@ we found
...
+                // ask for the rotation field from the packer element, or a default of 0.0f
+                cursor.Emit(OpCodes.Ldloc, packer_element_loc);
+                cursor.Emit(OpCodes.Ldstr, "rotation");
+                cursor.Emit(OpCodes.Ldc_R4, 0.0f);
+                cursor.Emit(OpCodes.Callvirt, m_BinaryPackerElementAttrFloat);
+                // store the rotation in the decaldata
+                cursor.Emit(OpCodes.Stfld, f_DecalDataRotation);
+
+                // make sure we won't find this location the next time we TryGotoNext
+                cursor.GotoNext(MoveType.After, instr => instr.MatchCallvirt("System.Collections.Generic.List`1<Celeste.DecalData>", "Add"));
```
bigkahuna443: this is a bit confusing in combo with the while loop, it's the very next instruction so you can probably just increment the cursor

Celeste.Mod.mm/Patches/Level.cs
microlith57: my ide does it automatically

Celeste.Mod.mm/Patches/Level.cs
microlith57: i tried that but it's a massive pain to get a reference to one of several constructors (they're all called `.ctor`); if you know a way to do that i'll implement it

Kalobi: `GetConstructor(Type[])`?

microlith57:
huh, apparently i can only request one review at a time

bigkahuna443 (approved):

PR #517, opened 2022-09-08
Track the Sprite component

bigkahuna443:
Allows for easier sprite manipulation/replacement. Number of Sprites per scene is typically not that large so should be low-impact


microlith57:
is there anywhere in vanilla/everest where using this tracking could be helpful? if so, that could be in scope for this pr

bigkahuna443:
Nope, doesn't look like it

Kalobi:
Unlike `EntityList.FindFirst`/`FindAll` for entities, I don't think there even is an easy non-Tracker ways to get all components of a type.

PR #520, opened 2022-09-13
Remove various instances of spurious logging

Kalobi:
- Don't log missing entity if the entity name is `theoCrystalHoldingBarrier`. This is an unused entity that appears in the level data for chapter 5, but does not have a code equivalent.
- Remove dev debug print from `BadelineBoost`.
- Don't log missing FMOD events if the name is `null`, `event:/none`, or `event:/null`. These are generally used as intentional placeholders if no audio is desired.
- Replace missing audio event `event:/char/badeline/boss_idle_ground` in `NPC06_Badeline_Crying` NPC with `event:/none`. A perfectly functional version of this event exists at `event:/env/local/06_reflection/boss_idle_ground`, but using this is quite a noticable change from vanilla and therefore probably out of scope for Everest.

### DIFF
--- a/Celeste.Mod.mm/Patches/Audio.cs
+++ b/Celeste.Mod.mm/Patches/Audio.cs
@@ -263,7 +263,7 @@ namespace Celeste {
                 Audio.cachedEventDescriptions.Add(path, desc);
 
             } else if (status == RESULT.ERR_EVENT_NOTFOUND) {
-                if (path is not ("null" or "event:/none" or "event:/null")) {
+                if (path is not ("null" or "event:/none")) {
                     Logger.Log(LogLevel.Warn, "Audio", $"Event not found: {path}");
                 }
 
--- a/Celeste.Mod.mm/Patches/Audio.cs
+++ b/Celeste.Mod.mm/Patches/Audio.cs
@@ -263,7 +263,9 @@ namespace Celeste {
                 Audio.cachedEventDescriptions.Add(path, desc);
 
             } else if (status == RESULT.ERR_EVENT_NOTFOUND) {
-                Logger.Log(LogLevel.Warn, "Audio", $"Event not found: {path}");
+                if (path is not ("null" or "event:/none" or "event:/null")) {
+                    Logger.Log(LogLevel.Warn, "Audio", $"Event not found: {path}");
+                }
 
             } else {
                 throw new Exception("FMOD getEvent failed: " + status);
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -635,6 +635,10 @@ namespace MonoMod {
                 }
             }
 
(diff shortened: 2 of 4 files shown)


Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -263,7 +263,9 @@ static class patch_Audio {
                 Audio.cachedEventDescriptions.Add(path, desc);
 
             } else if (status == RESULT.ERR_EVENT_NOTFOUND) {
-                Logger.Log(LogLevel.Warn, "Audio", $"Event not found: {path}");
+                if (path is not ("null" or "event:/none" or "event:/null")) {
```
bigkahuna443: Are `null` and `event:/null` really used/recommended anywhere? I've only seen `event:/none` and I think we should just stick with that usage 
EDIT: saw your comment about `null` being used by vanilla + I guess it's not a big deal since this only triggers after an FMOD error anyway

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -635,6 +635,10 @@ static partial class MonoModRules {
                 }
             }
 
+            cctor_il.Emit(OpCodes.Dup);
+            cctor_il.Emit(OpCodes.Ldstr, "theoCrystalHoldingBarrier"); // theoCrystalHoldingBarrier is an unused entity appearing in chapter 5
+            cctor_il.Emit(OpCodes.Callvirt, m_LoadStrings_Add);
+            cctor_il.Emit(OpCodes.Pop);
```
bigkahuna443, line 641: Feels weird to have this jammed into another patch, but this is one of those patches that needs to be refactored anyway so maybe it can just be a future problem™️ lol

Celeste.Mod.mm/Patches/Audio.cs
Kalobi: microlith said it was using event:/null somewhere in an unreleased map, I don't know if anyone else does though, might be worth removing. null is explicitly checked for in some places in vanilla (for example, using it as room music already doesn't show an error message because vanilla handles that), so it should be included here for consistency.

Celeste.Mod.mm/Patches/Level.cs
Kalobi: I considered just adding a separate check for it, but adding it to `_LoadStrings` seemed like the cleanest way, and I don't see a good place to do that other than the .cctor patch.

Celeste.Mod.mm/Patches/Audio.cs
bigkahuna443: oh gotcha. and yeah for `event:/null` it feels a bit odd to add a new convention for one unofficial usage. although I'm pretty sure `event:/none` was completely arbitrary too and never officially recommended lol

bigkahuna443 (approved):

PR #524, opened 2022-09-21
Drastically reduce memory allocations of SaveData.get_TotalCassettes()

JaThePlayer:
From `Celeste.SaveData.get_TotalCassettes()` taking ~60% of all per-frame allocations even in maps with entities that generate tons of memory allocations, this PR makes this function take ~8%. This definitely will have positive impact on GC times.

Of course, there's still a ways to go (this really should be a free operation), however the way save data is handled is so fundamentally horrible for performance that a full rewrite of this file is in order, but that's too daunting for me at the moment.

### DIFF
--- a/Celeste.Mod.mm/Patches/SaveData.cs
+++ b/Celeste.Mod.mm/Patches/SaveData.cs
@@ -88,7 +88,8 @@ namespace Celeste {
             get {
                 if (LevelSet == "Celeste")
                     return UnlockedAreas_Unsafe;
-                return LevelSetStats.AreaOffset + LevelSetStats.UnlockedAreas;
+                var stats = LevelSetStats;
+                return stats.AreaOffset + stats.UnlockedAreas;
             }
             set {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/SaveData.cs
```diff
@@ -88,7 +88,8 @@ public class patch_SaveData : SaveData {
             get {
                 if (LevelSet == "Celeste")
                     return UnlockedAreas_Unsafe;
-                return LevelSetStats.AreaOffset + LevelSetStats.UnlockedAreas;
+                var stats = LevelSetStats;
```
Kalobi: I think in Everest's code style we generally don't use `var` except for very long type names.

Kalobi:
8% of all memory allocations for a property that's only used for autosplitter info is still bad, but a marked improvement.

Kalobi (approved):

bigkahuna443 (approved):

PR #527, opened 2022-09-22
Add spawnpoint attribute to mark default spawn

Kalobi:
This allows marking a specific spawnpoint as default in a starting room with multiple spawnpoints by setting the boolean attribute `isDefaultSpawn`. If no spawnpoint has this option set, the old fallback is used, so this will not break any existing maps.
The `isDefaultSpawn` attribute will need to be added to the Lönn plugin to make this feature usable. Cruor and Vexatos both agreed that this is the most sensible way to handle default spawnpoints.

### DIFF
--- a/Celeste.Mod.mm/Patches/LevelData.cs
+++ b/Celeste.Mod.mm/Patches/LevelData.cs
@@ -1,5 +1,6 @@
 #pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
 
+using Microsoft.Xna.Framework;
 using System;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/LevelData.cs
```diff
@@ -108,5 +125,25 @@ we found
...
+            context.Body.Variables.Add(v_spawnCoords);
+
+            ILCursor cursor = new ILCursor(context);
+
+            // call CheckForDefaultSpawn with checkpoint element and coordinates
+            cursor.GotoNext(instr => instr.MatchLdstr("player"));
+            cursor.GotoNext(MoveType.After, instr => instr.MatchNewobj("Microsoft.Xna.Framework.Vector2"));
+            cursor.Emit(OpCodes.Stloc, v_spawnCoords);
+            cursor.Emit(OpCodes.Ldarg_0);
+            cursor.Emit(OpCodes.Ldloc, 9);
```
bigkahuna443: I think that the local variable number can change if the method gets more variables added to it? Would probably be more stable if you grab the one you need earlier using e.g. `instr.MatchLdloc(out int num)`

Kalobi: I would hope people only add new variables to the end of the list, but I can try grabbing it another way.

bigkahuna443: I mean you're adding a new variable into the middle lol. I guess I'd have to check the IL to see how it changes
EDIT: yeah colours recommended grabbing the variable itself (grabbing the index from the correct instruction in my example would accomplish the same thing)

bigkahuna443 (approved):

PR #528, opened 2022-09-23
Add Holdable SetSpeed

bigkahuna443:
Interface is a bit funky but matches the vanilla SpeedGetter format.


Kalobi (approved):

coloursofnoise:
Should `Holdable.GetSpeed` and `Holdable.SetSpeed` be merged into a `Holdable.Speed` property? I believe that `MonoModLinkFrom` could be used to provide compatibility with `GetSpeed`.

PR #531, opened 2022-09-26
Improvements to CustomTextVignette

Kalobi:
Improve custom vignettes in a few ways:

- Allow customizing the audio (currently still defaults to the prologue audio for backwards compatibility).
- No longer ends after a hardcoded wait duration, uses the `CoreVignette` approach instead.
- Allow customizing wait times at the start and end (vanilla has a shorter initial wait time in `CoreVignette` than `IntroVignette`, the wait time at the end can be used e.g. to wait for a sound to finish as `IntroVignette` does).
- Make sure the snow direction always gets applied, not just when reusing an existing `HiresSnow` (probably never matters).
- Center all text lines for text with line breaks.
- Make snow in vignette blow to the left by default to match vanilla.

The new wait time attributes, as well as the wind direction one, should also be documented on the wiki.

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/CustomTextVignette.cs
+++ b/Celeste.Mod.mm/Mod/Entities/CustomTextVignette.cs
@@ -1,4 +1,5 @@
-﻿using FMOD.Studio;
+﻿using Celeste.Mod.Meta;
+using FMOD.Studio;
 using Microsoft.Xna.Framework;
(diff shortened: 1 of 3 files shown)


coloursofnoise (approved):
Thanks for cleaning this up :sweat_smile: I'll admit I didn't put a lot of thought into it when I first implemented it

PR #540, opened 2022-10-18
Start music early if CassetteSong metadata exists and LeadBeats is 0

Vexatos:
Using the `CassetteSong` map metadata field sets `LeadBeats` to 16 by default, which ticks down over time and starts the cassette song once the value reaches 0. This code never runs if the `LeadBeats` metadata field is set to 0 to begin with, meaning the cassette song never starts. My solution here simply adds three lines to the `Update` method of `CassetteBlockManager`
```cs
if (leadBeats == 0) {
    beatIndex = 0;
    sfx?.start();
}
```
Exactly where the cassette song is fetched so that it is started immediately. If you have any suggestion for a cleaner solution than replacing the whole method, I'd like to know.


microlith57:
looks like this would be pretty easy to do with an il patch

Vexatos:
I don't think that having a 6+ line IL patch is really any cleaner than copypasting 8 lines of source code...

microlith57:
yea, but it would avoid replacing the whole method if that's a concern. didn't mean to suggest it would necessarily be the better option

PR #543, opened 2022-10-25
Switch to Discord Game SDK + rework the presence itself

### LINKED ISSUE #307: Replace DiscordRPC with GameSDK

maddie480:
- Closes #307 
- Brings more options on what to show or not in-game. By default, everything is turned on except "show room name": 

image

- Gives an icon to the presence: 

image

This supports all vanilla icons, the "?" icon and all icons of maps published on GameBanana that are stored in `Graphics/Atlases/Gui/areas` in their mod zip. For other icons (and when in menus), the Everest logo is displayed instead.


maddie480:
Tested to work on macOS (Intel) and Linux

maddie480:
Bugs related to not having Discord open:
- if Discord is not open when the game starts, it fails initializing with an "InternalError", but the error is caught so the game still starts up
- if you toggle Rich Presence off and back on, the game crashes with a NullReferenceException though
- if you close Discord while the game is running, it instantly dies with a "NotRunning" error

maddie480:
fixed 😅

maddie480:
There are still quite a few segfaults happening that should be figured out before this can be merged.

maddie480:
This segfault issue is incredibly cursed, as it seems to be Windows-only and to affect all methods based on callbacks in the SDK (getting a user by ID occasionally crashes as well). We did a bunch of gdb (... on Windows) with Popax21 and it seems to die in C# territory when trying to jump to the callback... But we don't know _where_ it dies in C# territory.

I found out an easy way to replicate the crash is just to turn Discord Rich Presence off and then back on in Mod Options. I tried that on Linux and it doesn't crash. 🤔

maddie480:
Fixed! This was caused by an issue in the Discord Game SDK that got reported here: https://github.com/discord/gamesdk-and-dispatch/issues/36

PR #545, opened 2022-10-29
Fix relinker not resolving cross-mod dependencies

Popax21:
Previously, the relinker didn't resolve cross-mod assembly references when patching the mod's assemblies using `MonoModder`. It also didn't throw any warnings or errors either, causing the relinking process to silently derail. This caused certain actions to cause bugs which were very hard to debug / pin down, like overriding a class from another mod and proceeding to call a method using `base.` causing an infinite recursive loop and subsequently a stack overflow crash. This fix resolves this issue by introducing a dictionary of currently relinked modules (`_RelinkedModules`), and consulting it in addition to other sources when trying to resolve assembly references. Additionally, a warning was added to notify modders when a dependent assembly failed to be resolved properly.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
@@ -231,7 +231,7 @@ namespace Celeste.Mod {
                             _RelinkedModules.Add(mod.Assembly.Name.Name, mod);
                             mod = null;
                         } else
-                            Logger.Log(LogLevel.Warn, "relinker", $"Encountered module name conflict loading assembly {meta} - {asmname} - {mod.Assembly.Name}");
+                            Logger.Log(LogLevel.Warn, "relinker", $"Encountered module name conflict loading cached assembly {meta} - {asmname} - {mod.Assembly.Name}");
 
                         return asm;
                     } catch (Exception e) {
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
@@ -221,6 +221,15 @@ namespace Celeste.Mod {
                     try {
                         Assembly asm = Assembly.LoadFrom(cachedPath);
                         _RelinkedAssemblies.Add(asm);
+
+                        ModuleDefinition mod = ModuleDefinition.ReadModule(cachedPath);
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
```diff
@@ -219,6 +221,15 @@ public static class Relinker {
                     try {
                         Assembly asm = Assembly.LoadFrom(cachedPath);
                         _RelinkedAssemblies.Add(asm);
+
+                        ModuleDefinition mod = ModuleDefinition.ReadModule(cachedPath);
+                        if (!_RelinkedModules.ContainsKey(mod.Assembly.Name.Name))
+                            _RelinkedModules.Add(mod.Assembly.Name.Name, mod);
+                        else {
+                            Logger.Log(LogLevel.Warn, "relinker", $"Encountered module name conflict loading assembly {meta} - {asmname} - {mod.Assembly.Name}");
```
coloursofnoise: This warning should probably specify that it's a cached assembly, since the verbose message above may not be included in the log

Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
```diff
@@ -334,46 +347,72 @@ public static class Relinker {
...
+
+                    try {
+                        Assembly asm = Assembly.LoadFrom(cachedPath);
+                        _RelinkedAssemblies.Add(asm);
+
+                        if (!_RelinkedModules.ContainsKey(module.Assembly.Name.Name)) {
+                            _RelinkedModules.Add(module.Assembly.Name.Name, module);
+                            module = null;
+                        } else
+                            Logger.Log(LogLevel.Warn, "relinker", $"Encountered module name conflict loading assembly {meta} - {asmname} - {module.Assembly.Name}");
```
coloursofnoise, line 413: Would it be possible to log the name of the conflicting mod?

Popax21: The `{meta}`-part of the log statement should already contain the name

Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
Popax21: Fixed in latest commit ac4a1b4cc6d5a34395073950544dfd5e8abe5ab3

coloursofnoise (approved):

PR #546, opened 2022-11-01
Hotload newly-enabled mods in `OuiModToggler` when none are disabled

mirrorcult:
I noticed that, when the dependency downloader installs updates/dependencies to a mod, it's able to hotload these at runtime. However, in the enable / disable mods menu, strictly enabling mods still leads to a full restart. This PR modifies `OuiModToggler` to be more intelligent and attempt to load unblacklisted mods at runtime. The old behavior still exists if any mods are blacklisted, or in the case of failure (e.g. out-of-memory). This is far more convenient as a player when I want to load up certain mods without restarting the entire game.

There is currently no way to disable this functionality (outside of runtime loading not being supported), as I didn't see a reason it would need to be disabled. There is also no explicit feedback that mods are being loaded, and I'm not really sure what the best way to do that would be if it's desirable (I guess just making it clearer in the string at the top of the menu?).

https://user-images.githubusercontent.com/19853115/203141051-d1d33e5e-2ab2-4bca-b6d6-26b71616862b.mp4

(this video was taken at 84506a6611b952bb26709a5de6a67e2b714f0d8a, the subsequent 2 commits are not included)


Celeste.Mod.mm/Mod/UI/OuiModToggler.cs
```diff
@@ -436,7 +436,29 @@ class OuiModToggler : OuiGenericMenu, OuiModOptions.ISubmenu {
                         }
                     }
 
-                    // restart the game
+                    // does the modified blacklist contain strictly un-blacklisted mods?
+                    if (Everest.Flags.SupportRuntimeMods && blacklistedMods.IsSubsetOf(blacklistedModsOriginal)) {
+                        foreach (string mod in blacklistedModsOriginal.Except(blacklistedMods)) {
+                            try {
+                                Logger.Log("OuiModToggler", $"Attempting to load {mod} after enabling");
```
coloursofnoise: LogLevels should be explicitly set

coloursofnoise:
There may be some considerable lag time when enabling mods, as seems to be shown in the video. Could a loading message be displayed to show the user that things are still happening?

Kalobi:
> Could a loading message be displayed to show the user that things are still happening?

I think it could make sense to replicate the log screen shown when installing dependencies, people are already used to that UI.

Kalobi (changes requested):
This does not properly consider dependencies, especially optional dependencies. The following situations don't work properly (assume mod A comes before mod B in alphabetical order):

- Mod A has mod B as an optional dependency:
  - Both mods are disabled on startup and are then enabled simultaneously.
    - Expected behavior: both mods are loaded in dependency order (B -> A).
    - Actual behavior: both mods are loaded in alphabetical order.
  - Both mods are disabled on startup. Mod A is enabled on its own. Then the menu is reentered and mod B is enabled.
    - Expected behavior: The game restarts because satisfying the load order guaranteed by the dependency would require unloading mod A to reload after mod B.
    - Actual behavior: Mod B is loaded, disregarding the load order guaranteed by the dependency.
- Mod A has mod B as a normal dependency.
  - Both mods are disabled on startup. Mod A is then enabled without enabling mod B (only possible if the Toggle Dependencies Automatically option is disabled).
    - This mostly works (A fails to load as expected), but the "1 mod failed to load" prompt under the Mod Options entry in the main menu does not appear. The prompt at the top of the Mod Options menu when opening it does appear.

It's possible that there are other edge cases with more complicated dependency webs, I only tested these simple cases.

mirrorcult:
> > Could a loading message be displayed to show the user that things are still happening?
> 
> I think it could make sense to replicate the log screen shown when installing dependencies, people are already used to that UI.

i'll try to make something like this work

> This does not properly consider dependencies, especially optional dependencies. The following situations don't work properly (assume mod A comes before mod B in alphabetical order):
> 
>     <snip>

for the latter it seems reasonable to just always restart if "Toggle Dependencies Automatically" is not active, for the former I suppose I'll just have to make the code a little more intelligent about handling dependencies; all cases I tested with regards to dependencies worked fine (though that was non-optional deps)

Celeste.Mod.mm/Mod/UI/OuiModToggler.cs
mirrorcult: 162574b6c739f5696b21fbda5606671fa7d10d61

Kalobi:
> for the latter it seems reasonable to just always restart if "Toggle Dependencies Automatically" is not active

What if I disable "Toggle Dependencies Automatically", toggle some mods, then reenable "Toggle Dependencies Automatically"?
Also, I don't think forcing people to restart even if they enabled all necessary dependencies is a great option.

Kalobi:
I think the case of enabling mods simultaneously with their optional dependencies can be fixed by replicating the sequence of events from `Everest.Loader.LoadAuto()` (enable `enforceOptionalDependencies` (would have to be made internal instead of private), load mods, then disable `enforceOptionalDependencies` and call `Everest.CheckDependenciesOfDelayedMods()`. Enabling optional dependencies after the fact is gonna require some additional logic to check for.
Also, any reason you're excluding non-zipped mods?

mirrorcult:
> Also, any reason you're excluding non-zipped mods?

i think i was just under the assumption that the mod toggler didn't show non-zipped mods but i never actually tested that assumption, it probably does

mirrorcult:
reviews addressed, ft. loading screen + dependency fixes

Regular mod loading
<details>

https://user-images.githubusercontent.com/19853115/203141051-d1d33e5e-2ab2-4bca-b6d6-26b71616862b.mp4

</details>

Loading optional dependency + dependent both at runtime (doesn't restart, loads in correct order)
<details>

https://user-images.githubusercontent.com/19853115/203141291-4ec1513a-6bfd-4b40-9c1e-dd36b074ea8f.mp4

</details>

Satisfying new optional dependency at runtime (force restarts)
<details>

https://user-images.githubusercontent.com/19853115/203141460-c234df5f-9ba6-4527-9513-d3a3e5c9989d.mp4

</details>

Loading a mod without its dependencies (error + UI properly updates)
<details>

https://user-images.githubusercontent.com/19853115/203141630-228ad448-04ed-4013-ac6d-e0f03492782e.mp4

</details>

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -92,6 +93,11 @@ public static class Loader {
 
             public static bool AutoLoadNewMods { get; internal set; }
 
+            /// <summary>
+            /// Used by <see cref="OuiModTogglerProgress"/>
+            /// </summary>
+            internal static List<string> NewlyUnblacklistedMods;
```
mirrorcult: this is here because I essentially need to pass data from one oui to another (the list of mods to enable from `OuiModToggler` to `OuiModTogglerProgress`). i don't like this solution but i couldn't see a better option than just a static list somewhere

Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private void toggleMods() {
+            // give it a second to transition
+            Thread.Sleep(1000);
+            int oldDelayedMods = Everest.Loader.Delayed.Count;
+            Everest.Loader.EnforceOptionalDependencies = true;
+            foreach (string mod in Everest.Loader.NewlyUnblacklistedMods) {
+                try {
+                    // remove the mod from the loaded blacklist & attempt to load mod
+                    LogLine(string.Format(Dialog.Get("DEPENDENCYDOWNLOADER_MOD_UNBLACKLIST"), mod));
```
mirrorcult, line 41: is reusing loc strings like this alright; should I duplicate it, rename the ID to something more generic?

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
mirrorcult: other option i considered was keeping it as members of `OuiModToggler` and just doing a pattern match cast for `from` in `OuiModTogglerProgress.Enter` (while also not erasing the data in `OuiModToggler.Leave`) but that seems equally bad

mirrorcult: oh NVM had realization

Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs
coloursofnoise: That should be fine, I don't expect that there'd be much room for confusion in other translations

Celeste.Mod.mm/Mod/UI/OuiModToggler.cs
```diff
@@ -436,21 +436,47 @@ class OuiModToggler : OuiGenericMenu, OuiModOptions.ISubmenu {
...
                     // we want to check if a non-blacklisted mod has this mod as a dependency (by name).
-                    return modYamls.Any(mod => !blacklistedMods.Contains(mod.Key) && modFilename != mod.Key
-                        && mod.Value.Any(yaml => yaml.Dependencies.Any(dependency => dependency.Name == modName)));
+                    return modYamls.Any(mod => 
+                            (onlyPreviouslyEnabledMods ? !blacklistedModsOriginal.Contains(mod.Key)
+                                                      : !blacklistedMods.Contains(mod.Key))
+                        && modFilename != mod.Key
+                        && mod.Value.Any(yaml => 
+                            optional ? yaml.OptionalDependencies.Any(dependency => dependency.Name == modName)
+                                     : yaml.Dependencies.Any(dependency => dependency.Name == modName)));
```
coloursofnoise: This has become a hell of a one-liner and should probably be broken up, but is still mostly understandable

Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs
```diff
@@ -0,0 +1,81 @@
...
+            
+            return base.Leave(next);
+        }
+
+        private void toggleMods() {
+            if (newlyUnblacklistedMods == null)
+                return;
+            
+            // give it a second to transition
+            Thread.Sleep(1000);
```
coloursofnoise, line 35: Does this serve a functional purpose or is it just aesthetic? if it is required then could we instead wait to initialize the Task until the end of the `Enter` coroutine?

coloursofnoise (approved):
I'm happy to merge as-is, but there's a couple of nitpicks that could either be included with this PR or marked for a future cleanup

Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs
mirrorcult: its pretty much required since the whole UI locks up before the transition finishes--but yes that seems like a better solution

mirrorcult: okay i think doing this will actually turn out to be way uglier than just sleeping

coloursofnoise: What about something like this? OuiLoggedProgress won't call `OnFinished` until the Task is no longer null *and* is completed.
```diff
diff --git a/Celeste.Mod.mm/Mod/UI/OuiLoggedProgress.cs b/Celeste.Mod.mm/Mod/UI/OuiLoggedProgress.cs
index dc215a89..a2ee34f0 100644
--- a/Celeste.Mod.mm/Mod/UI/OuiLoggedProgress.cs
+++ b/Celeste.Mod.mm/Mod/UI/OuiLoggedProgress.cs
@@ -44,7 +44,7 @@ namespace Celeste.Mod.UI {
 
             OnFinish += () => Overworld.Goto<T>();
 
-            if (task.Status == TaskStatus.Created)
+            if (task?.Status == TaskStatus.Created)
                 task.Start();
 
             return this;
diff --git a/Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs b/Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs
index f4eadf76..571323bd 100644
--- a/Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs
+++ b/Celeste.Mod.mm/Mod/UI/OuiModTogglerProgress.cs
@@ -14,10 +14,17 @@ namespace Celeste.Mod.UI {
         public void Init(List<string> mods) {
             Everest.Loader.OnCrawlMod += logCrawlMod;
             newlyUnblacklistedMods = mods;
-            Init<OuiMainMenu>(Dialog.Clean("MODOPTIONS_MODTOGGLE_PROGRESS_TITLE"), new Task(toggleMods),
+            Init<OuiMainMenu>(Dialog.Clean("MODOPTIONS_MODTOGGLE_PROGRESS_TITLE"), default(Task),
                 newlyUnblacklistedMods.Count);
         }
 
+        public override IEnumerator Enter(Oui prev) {
+            yield return base.Enter(prev);
+            // Delay starting until the transition is finished
+            Task = new Task(toggleMods, 0);
+            Task.Start();
+        }
+
         public override IEnumerator Leave(Oui next) {
             Everest.Loader.OnCrawlMod -= logCrawlMod;
             newlyUnblacklistedMods = null;
@@ -30,9 +37,7 @@ namespace Celeste.Mod.UI {
         private void toggleMods() {
             if (newlyUnblacklistedMods == null)
                 return;
-            
-            // give it a second to transition
-            Thread.Sleep(1000);
+
             int oldDelayedMods = Everest.Loader.Delayed.Count;
             Everest.Loader.EnforceOptionalDependencies = true;
             foreach (string mod in newlyUnblacklistedMods) {
```

Kalobi: Did you ever end up looking into this?

mirrorcult: i did and could not get it to work

PR #547, opened 2022-11-05
Make GetTypesSafe filter out "unsafe" types

Popax21:
Currently, a lot of issues are emerging where code relying on `GetTypesSafe` fails in some way or another, because it directly or indirectly depends on a type contained in an unloaded optional dependency, as these will be contained in the output of `GetTypesSafe` and subsequently crash when specific reflection operations are performed. This PR would fix these kinds of issues at the source, as it would ensure all types returned by `GetTypesSafe` have been "tested" for dependencies on non-present assemblies.

If this were to be merged, commits like https://github.com/EverestAPI/Everest/commit/f479dbbbe9451635d1b1fd3f55ad0322baa39e11 can be reverted.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Extensions.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Extensions.cs
@@ -335,11 +335,30 @@ namespace Celeste.Mod {
         public static bool IsUp(this TouchLocationState state)
             => state == TouchLocationState.Released || state == TouchLocationState.Invalid;
 
+        public static bool IsSafe(this Type type) {
+            try {
+                // "Probe" the type
+                _ = type.Name;
+                _ = type.Assembly.FullName;
+                _ = type.Module.FullyQualifiedName;
+
+                // Check declaring and base type
+                if (!type.DeclaringType?.IsSafe() ?? false)
+                    return false;
+                if (!type.BaseType?.IsSafe() ?? false)
+                    return false;
+
+                return true;
+            } catch {
+                return false;
+            }
+        }
+
         public static Type[] GetTypesSafe(this Assembly asm) {
             try {
-                return asm.GetTypes();
+                return asm.GetTypes().Where(t => t.IsSafe()).ToArray();
             } catch (ReflectionTypeLoadException e) {
-                return e.Types.Where(t => t != null).ToArray();
+                return e.Types.Where(t => t != null && t.IsSafe()).ToArray();
             }
         }
 


Popax21:
The newline at the end is from GitHub's web editor - that should probably be cleaned up before this gets merged.

Celeste.Mod.mm/Mod/Everest/Extensions.cs
```diff
@@ -335,11 +335,30 @@ public static bool IsDown(this TouchLocationState state)
...
+            try {
+                // "Probe" the type
+                _ = type.Name;
+                _ = type.Assembly.FullName;
+                _ = type.Module.FullyQualifiedName;
+
+                // Check declaring and base type
+                if (!type.DeclaringType?.IsSafe() ?? false)
+                    return false;
+                if (!type.BaseType?.IsSafe() ?? false)
```
Kalobi, line 348: Won't this always cause a return of `false` because we recurse down to `Object`, where `BaseType` will be `null`?

Popax21: The inversion using `!` should have operator precedence over the null coalescing operator (see <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/#operator-precedence>), which means that the nullable is inverted first before having its default value be applied.

Kalobi: > The predefined unary and binary [operators](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/) or any overloaded operators that are supported by a value type T are also supported by the corresponding nullable value type T?. These operators, also known as lifted operators, produce null if one or both operands are null; otherwise, the operator uses the contained values of its operands to calculate the result.

I.e. `!null` is just `null`, so I don't see why that would change anything.
Did you *test* this PR btw?

Popax21: > > The predefined unary and binary [operators](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/) or any overloaded operators that are supported by a value type T are also supported by the corresponding nullable value type T?. These operators, also known as lifted operators, produce null if one or both operands are null; otherwise, the operator uses the contained values of its operands to calculate the result.
> 
> I.e. `!null` is just `null`, so I don't see why that would change anything. Did you _test_ this PR btw?

Exactly - base type is `null`, which means both `type.BaseType?.IsSafe()` and `!type.BaseType?.IsSafe()` are `null`, which coalesces to `false`, meaning the if condition isn't met and the `return false;` isn't run.

Kalobi: Oh I am clueless

coloursofnoise (approved):

PR #548, opened 2022-11-12
Add RestartedFromGolden field to Session

bigkahuna443:
Field will be `true` for a session created from a golden berry restart.

Dashless goldens and similar challenges usually rely on `Session.StartedFromBeginning`, which is `false` after a golden restart if the golden is not in the first room. This new field lets you account for that instead of requiring the player to restart the chapter to respawn the berry/restart the challenge.

The dashless golden handler was also patched to use this new field.

The IL patches that were edited have also been reworked to use `ILCursor`. One side effect is that many invalid strings that were saved in _LoadStrings have been removed (previously the patch was saving every `ldstr` used in `Level.orig_LoadLevel()` as an entity name). Apart from that, the decompiled output is the same.

### DIFF
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -556,103 +556,99 @@ namespace MonoMod {
 
     static partial class MonoModRules {
 
-        public static void PatchLevelLoader(MethodDefinition method, CustomAttribute attrib) {
-            // We also need to do special work in the cctor.
-            MethodDefinition m_cctor = method.DeclaringType.FindMethod(".cctor");
+        public static void PatchLevelLoader(ILContext context, CustomAttribute attrib) {
+            FieldReference f_Session = context.Method.DeclaringType.FindField("Session");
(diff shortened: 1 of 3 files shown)


bigkahuna443:
Removed strings:
```"0", "ignore_darkness_", "-1", "platforms", "clockwise", "d-", "perLevel", "disable_lightning", "condition", "conditionID", "", "Condition type not supported!", "broken", "number", "npc", "granny_00_house", "theo_01_campfire", "theo_02_campfire", "theo_03_escaping", "resort_theo", "theo_03_vents", "oshiro_03_lobby", "oshiro_03_hallway", "oshiro_03_hallway2", "oshiro_03_bigroom", "oshiro_03_breakdown", "oshiro_03_suite", "oshiro_03_rooftop", "granny_04_cliffside", "theo_04_cliffside", "theo_05_entrance", "theo_05_inmirror", "evil_05", "theo_06_plateau", "granny_06_intro", "badeline_06_crying", "granny_06_ending", "theo_06_ending", "granny_07x", "theo_08_inside", "granny_08_inside", "granny_09_outside", "granny_09_inside", "gravestone_10", "granny_10_never", "deleteFlag", "tilesets/scenery"```

PR #552, opened 2022-11-22
add ambience and ambience volume triggers

### LINKED ISSUE #551: music / ambience trigger parity

Yes

### Describe your request

currently, vanilla+everest contains these triggers:
- music
- music fade
- music layer
- alt music
- ambience param

i suggest the following triggers to be added to everest, to achieve parity between the two:
- [x] music param (can be done with vanilla's music fade trigger)
- [x] ambience
- [x] ambience fade (can be done with vanilla's ambience param trigger)
- [x] alt ambience (can be done with ambience trigger)

### Additional context

- see [here](https://discord.com/channels/403698615446536203/429775320720211968/1044410751898632274) for initial suggestion
- see [here](https://discord.com/channels/403698615446536203/429775439423209472/994864805015998514) for sardine7's ambience trigger implementation

microlith57:
closes https://github.com/EverestAPI/Everest/issues/551
code from [here](https://discord.com/channels/403698615446536203/429775439423209472/994864805015998514) thanks to fishtank_overflow

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/AmbienceTrigger.cs
+++ b/Celeste.Mod.mm/Mod/Entities/AmbienceTrigger.cs
@@ -1,0 +1,42 @@
+using Microsoft.Xna.Framework;
+
+namespace Celeste.Mod.Entities {
+    /// <summary>
+    /// A trigger to change the current ambience track.
+    ///
+    /// Attributes:
+    /// - `track`: the audio event to set the ambience to
+    /// - `resetOnLeave`: if true, remembers what the ambience was before entering the trigger, and restores it after leaving
+    /// </summary>
+    [CustomEntity("everest/ambienceTrigger", "Sardine7/AmbienceTrigger")]
+    public class AmbienceTrigger : Trigger {
+
+        public string Track;
+        public bool ResetOnLeave;
+
+        private string oldTrack;
+
+        public AmbienceTrigger(EntityData data, Vector2 offset) : base(data, offset) {
+            Track = data.Attr("track");
+            ResetOnLeave = data.Bool("resetOnLeave", defaultValue: false);
+        }
+
+        public override void OnEnter(Player player) {
+            if (ResetOnLeave) {
+                oldTrack = Audio.GetEventName(Audio.CurrentAmbienceEventInstance);
+            }
+
+            Session session = SceneAs<Level>().Session;
+            session.Audio.Ambience.Event = SFX.EventnameByHandle(Track);
+            session.Audio.Apply();
+        }
+
+        public override void OnLeave(Player player) {
+            if (ResetOnLeave) {
+                Session session = SceneAs<Level>().Session;
+                session.Audio.Ambience.Event = oldTrack;
+                session.Audio.Apply();
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


microlith57:
concerns currently:
- ~~i haven't tested this yet; i will do so once i make a lönn plugin~~
- ~~it isn't possible to change the volume of ambience tracks as this isn't done with a `fade` param~~
  - ~~@max4805 would it be reasonable to move your ambience volume trigger into everest?~~
- it might be a good idea to have multiple lönn placements per entity as it's counterintuitive having the 'music fade trigger' placement also be used as the way to do things like changing other params to constant values
- some of the existing vanilla/everest triggers use `bool leftToRight` rather than the more flexible `PositionModes positionMode`; it might be worth adding support for the latter

microlith57:
lönn plugin for ambience trigger:
```lua
local enums = require("consts.celeste_enums")
local ambient_sounds = table.keys(enums.ambient_sounds)

table.sort(ambient_sounds)

local ambience = {}

ambience.name = "everest/ambienceTrigger"
ambience.fieldInformation = {
    track = {
        options = ambient_sounds
    }
}
ambience.placements = {
    name = "ambience",
    data = {
        track = "",
        resetOnLeave = true
    }
}

return ambience
```
seems to work fine!

microlith57:
now has an ambience volume trigger also; i used this plugin for testing:
```lua
local enums = require("consts.celeste_enums")

local trigger = {}

trigger.name = "everest/ambienceVolumeTrigger"
trigger.fieldInformation = {
    direction = {
        options = enums.trigger_position_modes,
        editable = false
    }
}
trigger.placements = {
    name = "ambience_volume",
    data = {
        from = 0.0,
        to = 0.0,
        direction = "NoEffect"
    }
}

return trigger
```

Celeste.Mod.mm/Patches/AudioState.cs
```diff
@@ -1,9 +1,50 @@
...
+            }
+            AmbienceVolume = 1f;
+        }
+
+        [MonoModConstructor]
+        public patch_AudioState(string music, string ambience) {
+            Music.Event = music;
+            Ambience.Event = ambience;
+            AmbienceVolume = 1f;
+        }
```
microlith57: initially i had `[MonoModReplace]` on these as i'm not calling the orig constructors, but that caused MiniInstaller to crash with a cryptic error.
i'd prefer to use `[MonoModReplace]` here, so unless that's impossible i'd prefer to hold off on merging this

Celeste.Mod.mm/Mod/Entities/AmbienceTrigger.cs
```diff
@@ -0,0 +1,42 @@
...
+    ///
+    /// Attributes:
+    /// - `track`: the audio event to set the ambience to
+    /// - `resetOnLeave`: if true, remembers what the ambience was before entering the trigger, and restores it after leaving
+    /// </summary>
+    [CustomEntity("everest/ambienceTrigger", "Sardine7/AmbienceTrigger")]
+    public class AmbienceTrigger : Trigger {
+
+        public string Track;
+        public bool ResetOnLeave;
```
microlith57: should these be `public` or `private`?

PR #554, opened 2022-11-24
Fix Delete handler in Map Select

### LINKED ISSUE #553: Pressing Delete in the Map Search does not clear the entire field on XNA

Yes

Yes

### Describe the bug

In XNA, when entering the **Map List** and searching for anything, pressing <kbd>Delete</kbd> does not clear the input field.
FNA does not exhibit the same behavior.

### Steps to reproduce

0. Ensure your build is XNA 
1. Pick a save file
1. Go to the **Map List** (usually <kbd>ESC</kbd>)
1. Open **Search**
1. Enter any text
1. Press the <kbd>Delete</kbd> key
1. Observe absolutely nothing happen *(tada)*

### Expected behavior

The text input is supposed to be cleared entirely.

### Operating System

Windows 10 19044.2251

### Everest Version

3668

_No response_

### Additional context

The **Map Search** uses the character `0x7F` to detect when <kbd>Delete</kbd> is pressed.

https://github.com/EverestAPI/Everest/blob/8cb5b5ee6222505d88104b214f29c49a235bd8c9/Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs#L180-L190

It seems that for whatever reason `TextInput.OnInput` does not fire a `0x7F` char on XNA when <kbd>Delete</kbd> is pressed, but does so when <kbd>Ctrl</kbd>+<kbd>Backspace</kbd> are hit. *(which shouldn't be happening)*
Fortunately, `MInput.Keyboard.Check` does report that the key has been pressed.

Changing frameworks to FNA works as intended however.

A similar bug has already been experienced in **CelesteNet**: https://github.com/0x0ade/CelesteNet/pull/57

SnipUndercover:
Fixes the <kbd>Delete</kbd> handler by checking `MInput.Keyboard.Pressed(Keys.Delete)` in `Update()` instead of using `OnTextInput(char c)`. 

Closes #553.

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs
+++ b/Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs
@@ -120,7 +120,6 @@ namespace Celeste.Mod.UI {
 
         private void clearSearch() {
             search = "";
-            searchPrev = "";
         }
 
         private void cleanExit() {
--- a/Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs
+++ b/Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs
@@ -508,7 +508,7 @@ namespace Celeste.Mod.UI {
                         clearSearch();
                         Audio.Play(SFX.ui_main_rename_entry_backspace);
                     } else {
-                        Audio.Play(SFX.ui_main_button_invalid);
+                        Audio.Play(SFX.ui_main_button_back);
                         cleanExit();
                     }
                     searchConsumedButton = true;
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs
```diff
@@ -179,15 +179,7 @@ private class SearchMenu : Entity {
 
             } else if (c == (char) 127) {
```
Kalobi: This branch of the if should just be removed outright IMO, there's no reason to keep it.

SnipUndercover: The reason I left it is because I saw several other files leave the branch alone:
https://github.com/EverestAPI/Everest/blob/611172ef3cc4b2d486db69d57e77fc9640e5c08a/Celeste.Mod.mm/Patches/OuiFileNaming.cs#L45-L48
Going to remove it now.

PR #557, opened 2022-11-30
Ensure dream blocks have the same number of particles on all platforms

### LINKED ISSUE #556: Desync on macOS due to floating point rounding in DreamBlock.Setup

Yes

Yes

### Describe the bug

The code in `DreamBlock.Setup()` that calculates the desired number of dream block particles has inconsistent rounding errors between macOS and Windows.  Untested on Linux but I assume this is a Mono issue rather than specifically macOS, so the issue likely exists there too.

The calculation is `(int)(width / 8f * (height / 8f * 0.7f))`.  With a width and height of 120x32, this results in 42 particles on macOS and 41 on Windows.  Performing the calculation with doubles gives 42 on both platforms.

Due to this inconsistency, a different number of random samples are taken, causing a TAS desync in certain places.  One example is Farewell `j-08`.  The change in the random sequence causes the `CrumbleWallOnRumble` entities to break in a different order such that the TAS no longer gets a corner boost into the next screen.

After some discussion in Celestecord, one potential solution is to subtract a small value (say 0.0001f) before truncating to int, which in this case will result in 41 particles on both platforms.  I'm willing to implement and test this on both Windows and macOS.

Note that this is a bug in vanilla, not Everest.

### Steps to reproduce

Run 9.tas and skip to j-08 (Reconciliation).
Note that Madeline does not a proper corner boost at the end of the room and thus dies to lightning in the next room.

Test code that simulates the 120x32 dream block (running in module Load):
```cs
float width = 120f;
float height = 32f;
int floatValue = (int) (width / 8f * (height / 8f * 0.7f));
int doubleValue = (int) (width / 8.0 * (height / 8.0 * 0.7));
int adjustedFloatValue = (int) (width / 8f * (height / 8f * 0.7f) - 0.0001f);
Logger.Log(LogLevel.Error, nameof(SortFixModule), $"(float) 120f/8f * (32f/8f * 0.7f) = {floatValue}");
Logger.Log(LogLevel.Error, nameof(SortFixModule), $"(double) 120f/8.0 * (32f/8.0 * 0.7) = {doubleValue}");
Logger.Log(LogLevel.Error, nameof(SortFixModule), $"(adjusted float) 120f/8f * (32f/8f * 0.7f) - 0.0001f = {adjustedFloatValue}");
```

macOS:
```
(11/30/2022 20:43:02) [Everest] [Error] [SortFixModule] (float) 120f/8f * (32f/8f * 0.7f) = 42
(11/30/2022 20:43:02) [Everest] [Error] [SortFixModule] (double) 120f/8.0 * (32f/8.0 * 0.7) = 42
(11/30/2022 20:43:02) [Everest] [Error] [SortFixModule] (adjusted float) 120f/8.0 * (32f/8.0 * 0.7) - 0.0001f = 41
```

Windows:
```
(11/30/2022 20:44:33) [Everest] [Error] [SortFixModule] (float) 120f/8f * (32f/8f * 0.7f) = 41
(11/30/2022 20:44:33) [Everest] [Error] [SortFixModule] (double) 120f/8.0 * (32f/8.0 * 0.7) = 42
(11/30/2022 20:43:02) [Everest] [Error] [SortFixModule] (adjusted float) 120f/8.0 * (32f/8.0 * 0.7) - 0.0001f = 41
```

### Expected behavior

Blocks break in the correct order and Madeline does not die in the next room.

### Operating System

macOS and Windows 10

### Everest Version

3650

CelesteTAS and 9.tas if you want to test desync, but the issue exists with no mods and can be seen just based on the different autotiling of the blocks at the end of the room.

### Additional context

_No response_

swoolcock:
Windows can sometimes give a different value to macOS for float calculations close to a whole number, whereas they both agree if calculating with doubles.

Due to this, `DreamBlock.Setup` calculates a different number of particles for certain dream blocks.  This means more random samples taken, and a desync later in the room.

Solution is to subtract a small value before truncating to int, which gives the desired value on both platforms.

Bug and resolution tested on both Windows 10 and macOS 12.4 (Monterey).

Resolves #556

### DIFF
--- a/Celeste.Mod.mm/Patches/DreamBlock.cs
+++ b/Celeste.Mod.mm/Patches/DreamBlock.cs
@@ -199,6 +199,23 @@ namespace Celeste {
 namespace MonoMod {
     /// <summary>
     /// Patches <see cref="DreamBlock.Setup()" /> to fix issue #556.
+    /// 
+    /// When Celeste calculates the required number of particles for the dream block,
+    /// there is a rounding error for values near whole numbers.
+    /// 
+    /// The calculation is (int)((width / 8f) * (height / 8f) * 0.7f)
+    /// 
+    /// For example, a 120x32 dream block should theoretically give 42 particles.
+    /// Due to single precision floating point calculations, this actually comes out
+    /// to 41.999 and is truncated to 41.
+    /// 
+    /// On macOS however, this is still calculated as 42.  This means that it consumes an
+    /// extra 5 random numbers in the sequence and desyncs other things in the room.
+    ///
+    /// Regardless of which value is "correct", the following fix makes macOS consistent
+    /// with Windows/Linux by subtracting a small value (0.001f) before truncating to int.
+    /// This should only affect values close to whole numbers, which happens only when
+    /// the total number of tiles in a dream block is a multiple of 10.
     /// </summary>
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchDreamBlockSetup))]
     class PatchDreamBlockSetupAttribute : Attribute { }
--- a/Celeste.Mod.mm/Patches/DreamBlock.cs
+++ b/Celeste.Mod.mm/Patches/DreamBlock.cs
@@ -1,6 +1,10 @@
-﻿using Microsoft.Xna.Framework;
+﻿using Celeste;
+using Microsoft.Xna.Framework;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
 using Monocle;
 using MonoMod;
+using MonoMod.Cil;
 using System;
 using System.Collections;
 
@@ -27,6 +31,10 @@ namespace Celeste {
         public void ctor(Vector2 position, float width, float height, Vector2? node, bool fastMoving, bool oneUse) {
             ctor(position, width, height, node, fastMoving, oneUse, false);
         }
+
+        [MonoModIgnore]
+        [PatchDreamBlockSetup]
+        public new extern void Setup();
 
         public void DeactivateNoRoutine() {
             if (playerHasDreamDash) {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/DreamBlock.cs
```diff
@@ -187,3 +195,29 @@ public patch_DreamBlock(EntityData data, Vector2 offset)
         }
     }
 }
+
+namespace MonoMod {
+    /// <summary>
+    /// Patches <see cref="DreamBlock.Setup()" /> to fix issue #556.
```
Kalobi, line 201: I think the description should actually tell you what the patch does without having to check a Github issue.

swoolcock: Fair point, I was just copying the format from another patch class  that directly references GitHub.

microlith57 (approved):

Kalobi:
Do we know that this cannot cause a change in the particle count on Windows for some dreamblocks? It seems like it could.

swoolcock:
> Do we know that this cannot cause a change in the particle count on Windows for some dreamblocks? It seems like it could.

This is a real edge case that only happens with dream blocks with a multiple of 10 tiles squared.  If you're concerned then we could put an "if macOS" around it.

microlith57:
i'd favour a macOS check personally, even if in practice there isn't a difference

swoolcock:
> i'd favour a macOS check personally, even if in practice there isn't a difference

Sounds good, can you provide an example that will work with MonoMod?  I tried something earlier (I can't remember what exactly, it was about a month ago), but it seemed to ignore it.

Celeste.Mod.mm/Patches/DreamBlock.cs
```diff
@@ -187,3 +195,46 @@ public patch_DreamBlock(EntityData data, Vector2 offset)
...
+    /// This should only affect values close to whole numbers, which happens only when
+    /// the total number of tiles in a dream block is a multiple of 10.
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchDreamBlockSetup))]
+    class PatchDreamBlockSetupAttribute : Attribute { }
+
+    static partial class MonoModRules {
+
+        public static void PatchDreamBlockSetup(ILContext context, CustomAttribute attrib) {
+            ILCursor cursor = new ILCursor(context);
```
microlith57, line 226:
```suggestion
            float width = 120;
            float height = 32;
            int particles = (int)((width / 8f) * (height / 8f) * 0.7f);
            
            if (particles == 41)
                return;

            ILCursor cursor = new ILCursor(context);
```

microlith57: if we can guarantee the float-truncation behaviour is the same between celeste.exe and miniinstaller.exe, i think this suffices

swoolcock: Nice, I'll give this a shot.

Kalobi:
This issue is fixed by the move to .NET Core, which addresses float platform differences much more generically. The TAS has been confirmed to sync on that version on all platforms.

PR #562, opened 2022-12-15
Backdrop.BeforeRender no longer gets called if not Visible + small optimisations

JaThePlayer:
* `Backdrop.BeforeRender` currently gets called all the time, even if the backdrop is not visible. While *some* modded backdrops manually add a `if (!Visible) return;` check at the start of the method, most of them (and all of vanilla) don't. Since this method doesn't get used for updating state, it can be relatively safely changed to only get called when the backdrop is visible. This gives major performance boosts to large maps like Heartsides.
Might be worth checking mods for improper usage of this method first though.

This PR also has some smaller optimisations:
* Added culling for Spikes and Seeker Barriers
* Made Everest.Events.LoadEntity no longer use the `InvokeWhileFalse` extension method, as it uses dynamic dispatch behind the scenes, which actually showed up quite a bit in performance checks. The method got essentially inlined into `LoadEntity` now, but with types known ahead of time.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
@@ -119,8 +119,20 @@ namespace Celeste.Mod {
                 /// Called during <see cref="patch_Level.LoadCustomEntity(EntityData, _Level)"/>.
                 /// </summary>
                 public static event LoadEntityHandler OnLoadEntity;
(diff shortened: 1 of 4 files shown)


microlith57:
this seems like it could break styleground windows etc. that rely on being able to render otherwise invisible effects

coloursofnoise:
This patch has been determined to have a a low likelihood of conflicting with mods that try to render stylegrounds manually, as they will already be needing to set `Visible=true` to work proprerly with modded stylegrounds.

PR #563, opened 2022-12-20
allow recolouring decals (with alpha channel)

microlith57:
along the same lines as #467 (and using the same hooks), this allows decals to be tinted with arbitrary colours, including opacity.


microlith57:
map editor implementation note: this works by adding a `color` attr for the hex colour, which ideally should be in the most compact form possible.
- if the decal isn't recoloured (RGBA `#ffffffff`), leave out the attr
- if only the opacity is changed, put only those two characters in (i.e. RGBA `#ffffffaa` → `color="aa"`)
- if only the colour is changed, put those 6 characters (RGBA `#coffeeff` → `color="coffee"`)
- if both are changed, put all 8 characters (RGBA `#coffeeaa` → `color="coffeeaa"`)

microlith57:
after feedback from lönn devs, it's better to have just the two cases of leaving out the attr or including all 8 characters.

DemoJameson:
the documentation should be updated when the stable version is released

PR #576, opened 2023-01-30
Fix AmbienceVolumeTrigger breaking when MaxHelpingHand is installed

microlith57:
`MaxHelpingHand/AmbienceVolumeTrigger` placements broke when helping hand was installed; this should fix that by only setting the event volume when the everest trigger is being used.

currently the value is set to `null` in the ctor's even though this shouldn't be strictly necessary; i'd be willng to change this if that's not Everest code style.


Celeste.Mod.mm/Patches/AudioState.cs
```diff
@@ -1,4 +1,4 @@
-﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
```
microlith57, line 1: ??? not sure what this diff is

Celeste.Mod.mm/Patches/AudioState.cs
```diff
@@ -1,36 +1,37 @@
...
     class patch_AudioState : AudioState {
 
-        public float AmbienceVolume;
+        public float? AmbienceVolume;
 
         [MonoModConstructor]
         [MonoModReplace]
         public void ctor() {
-            AmbienceVolume = 1f;
+            AmbienceVolume = null;
```
bigkahuna443: like you mentioned in the desc, these initializations aren't needed anymore since nullables default to null. not much point in having three unnecessary patches

bigkahuna443 (approved):

Issue #577, opened 2023-02-01
Proposal: Deprecate __Ext classes


coloursofnoise:
"__Ext" classes (such as [`AreaDataExt`](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Patches/AreaData.cs#L472-L595)) were originally created because mods were built using Everest as a project reference, and so were unable to access fields or methods added by Everest in `patch_` classes.

For the most part they are no longer needed, since almost all mods are now built against a patched `Celeste.exe`.

I propose that we explicitly deprecate the use of these classes, instead favoring adding members directly to the `patch_` classes. Existing methods that add an API not already available in the `patch_` class can be re-implemented within it, and the "__Ext" classes marked as obsolete.

In the future, most "__Ext" classes could be stubbed out or removed entirely and existing references re-linked to the associated `patch_` class members.

"__Ext" classes that explicitly extend the vanilla functionality, such as [`TextMenuExt`](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs) would be kept.

Kalobi:
Yeah that sounds sensible. Having these around without any indication that they're an old deprecated feature can cause unnecessary confusion.

PR #581, opened 2023-02-04
Deprecate __Ext classe members

coloursofnoise:
- First part of #577 

Adds `Obsolete` attributes to redundant members and fixes references to them in Everest.

This adds more casts to `patch_` classes that become redundant, and it would be nice to be able to remove them at patch time, however it could be tricky to determine which ones are actually unnecessary.


Celeste.Mod.mm/Patches/PlayerHair.cs
```diff
@@ -94,30 +95,32 @@ public patch_PlayerHair(PlayerSprite sprite)
...
-        // Mods can't access patch_ classes directly.
-        // We thus expose any new members through extensions.
-
         /// <inheritdoc cref="patch_PlayerHair.GetHairTexture(int)"/>
+        [Obsolete("Use PlayerHair.GetHairTexture instead.")]
         public static MTexture GetHairTexture(this PlayerHair self, int index)
             => ((patch_PlayerHair) self).GetHairTexture(index);
 
         /// <inheritdoc cref="patch_PlayerHair.GetHairScale(int)"/>
+        [Obsolete("Use PlayerHair.GetHairTexture instead.")]
```
maddie480:
```suggestion
        [Obsolete("Use PlayerHair.PublicGetHairScale instead.")]
```

PR #582, opened 2023-02-12
Remove the aggressive GC call in Level.Reload

JaThePlayer:
The `GC.Collect()` call in `Level.Reload` is often responsible for over 50% of all load times after dying.
image
This is because it does a *full* GC run through all generations, which is quite useless, since most of the RAM used by Celeste is used up by things that can never get GC'd anyway.

This patch is already available via Frost Helper's `No Reload GC` experimental setting, and didn't seem to cause any gameplay lag spikes.

### DIFF
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -69,6 +69,10 @@ namespace Celeste {
         [PatchLevelUpdate] // ... except for manually manipulating the method via MonoModRules
         public extern new void Update();
 
+        [MonoModIgnore] // We don't want to change anything about the method...
+        [PatchLevelReload] // ... except for manually manipulating the method via MonoModRules
+        public extern new void Reload();
+
         [MonoModReplace]
         public new void RegisterAreaComplete() {
             bool completed = Completed;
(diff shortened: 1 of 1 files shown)


caeyo:
Reported this to Ja directly but don’t know if he saw it so will put it here too: 
I used this for quite a while in Frost Helper and saw nothing but improvements to level reload times, but when my system was under significantly more stress (was streaming to Discord while playing SJ) it finally cracked and I started to see increased lag spikes to the point of making it unplayable. As soon as I turned the option off in Frost Helper it had one big old reload on the next death and was then fine for lag spikes (although death reload times were definitely worse), and any subsequent attempts to turn it back on in that session went through the same process of an increasing amount of lag spikes over time. I didn’t have a look at specific memory usage during that session (can go back and try to replicate if needed) and it’s obviously only a small sample size, but figured it was worth reporting since I don’t know how many people have actually tested this

maddie480 (changes requested):
Since the feedback seems to be that it can make the situation better or worse depending on the situation, this should at least be turned into a toggle... and we should check if this is still relevant on the core branch. 🤔

maddie480:
Some experimenting has shown that while it removes room reloading times, it can add some lag during gameplay, so we agreed that it probably shouldn't make it into Everest

PR #584, opened 2023-02-14
fix AmbienceVolume being serialised when null

microlith57:
previously, the AmbienceVolume field was written to the save file unconditionally, which produced unreadable XML when it was `null` (i.e. when the player hadn't interacted with an `everest/ambienceVolumeTrigger`).

this fixes that issue by saving the field only if it contains a value, using a magic property.

### DIFF
--- a/Celeste.Mod.mm/Patches/AudioState.cs
+++ b/Celeste.Mod.mm/Patches/AudioState.cs
@@ -1,9 +1,14 @@
 #pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
 
+using System.Xml.Serialization;
+
 namespace Celeste {
     class patch_AudioState : AudioState {
 
         public float? AmbienceVolume;
+
+        [XmlIgnore]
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/AudioState.cs
```diff
@@ -1,10 +1,15 @@
...
 
+using System.Xml.Serialization;
+
 namespace Celeste {
     class patch_AudioState : AudioState {
 
         public float? AmbienceVolume;
 
+        [XmlIgnore]
+        public bool AmbienceVolumeSpecified { get { return AmbienceVolume.HasValue; } }
```
Kalobi:
```suggestion
        public bool AmbienceVolumeSpecified => AmbienceVolume.HasValue;
```

DemoJameson:
I think this only happens when downgrading to a Everest version with non-nullabel `AmbienceVolume`
so it's usually not a problem that needs to be fixed anymore
can we close this pr?

microlith57:
yea i think so; nobody seems to be having any issues with it

PR #592, opened 2023-03-11
Fix explosion boost leniency inconsistency

Kalobi:
In vanilla, the one frame of leniency added to ExplodeLaunch boosts in 1.4 does not work consistently for Puffers if the player's Depth was set after the Puffers (e.g. when the player respawns after death vs. entering the room) because Puffers have Depth 0 and often call ExplodeLaunch from their Update method instead of a PlayerCollider callback, causing the "leniency" check in Player.Update to happen on the *same frame* as the check in ExplodeLaunch. This patch fixes that issue.
The simplest approach to fix this would be to give Puffers a Depth of -1, forcing them to update after the player, However, this would potentially change the update order of Puffers and other entities with Depth -1, such as FlingBirds, which are likely to appear in the same levels. Therefore, I'm choosing a different approach here, saving in the Level when ExplodeLaunch is called in a frame.
Because this is technically a change to vanilla "mechanics", a check is included to make this fix only apply in custom maps.

### DIFF
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -41,6 +41,8 @@ namespace Celeste {
 
         private float unpauseTimer;
 
+        internal bool playerWasExplodeLaunchedThisFrame;
+
         /// <summary>
         /// If in vanilla levels, gets the spawnpoint closest to the bottom left of the level.<br/>
         /// Otherwise, get the default spawnpoint from the level data if present, falling back to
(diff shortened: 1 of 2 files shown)


DemoJameson:
It might be better to patch ExplodeLaunch with il, otherwise it will cause the existing mod's ilhook to fail on it

DemoJameson:
It might be better to patch ExplodeLaunch with il, otherwise it will cause the existing mod's ilhook to fail on it

PR #598, opened 2023-03-14
Fix TextInput.OnInput not firing on Wayland

Lycan0x:
### Bug Description

On Linux using Wayland the TextInput.OnInput event was not firing resulting in the debug console and CelesteNet chat not accepting any input after being opened.

### System info

OS: Arch Linux x86_64
Kernel: Linux 6.2.1-arch1-1
WM: sway

### Reproduction

1) Run Celeste with Everest loaded
2) Open debug console using `~` or `.`
3) Try typing in the console

### Fix description

This fix just adds a call to `TextInputEXT.StartTextInput()` on initialization of TextInput.

As per [TextInputEXT documentation](https://github.com/FNA-XNA/FNA/wiki/5:-FNA-Extensions#textinputext) one should call `TextInputEXT.StartTextInput()` before trying to receive events through the `TextInputEXT.TextInput` event.

A similar [bug](https://github.com/libsdl-org/SDL/issues/4857#issuecomment-962843448) was reported on the SDL2 repository. These seem correlated as FNA uses SDL2 internally AFAIK.

### Considerations

The method `SDL_StartTextInput` seems to intend on opening a virtual keyboard when the platform requires it. This should possibly be checked on Platforms like the SteamDeck or similar. I was not able to produce this behaviour using my system and do not own a SteamDeck.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/TextInput.cs
+++ b/Celeste.Mod.mm/Mod/Everest/TextInput.cs
@@ -31,6 +31,10 @@ namespace Celeste.Mod {
                 // public static event Action<char> TextInput;
                 e_TextInput.AddEventHandler(null, new Action<char>(ReceiveTextInput).CastDelegate(e_TextInput.EventHandlerType));
 
+                // Some platforms like Linux/Wayland may require calling TextInputEXT.StartTextInput to receive events for TextInputEXT.TextInput
+                MethodInfo m_StartTextInput = t_TextInputExt?.GetMethod("StartTextInput", new Type[] { } );
+                m_StartTextInput?.Invoke(t_TextInputExt, null);
+
                 // SDL2 offers SDL_GetClipboardText and SDL_SetClipboardText
                 Type t_SDL2 = typeof(Keyboard).Assembly.GetType("SDL2.SDL");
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Mod/Everest/TextInput.cs
```diff
@@ -31,6 +31,10 @@ public static class TextInput {
                 // public static event Action<char> TextInput;
                 e_TextInput.AddEventHandler(null, new Action<char>(ReceiveTextInput).CastDelegate(e_TextInput.EventHandlerType));
 
+                // Some platforms like Linux/Wayland may require calling TextInputEXT.StartTextInput to receive events for TextInputEXT.TextInput
+                MethodInfo m_StartTextInput = t_TextInputExt?.GetMethod("StartTextInput", new Type[] { } );
```
Popax21: Small nitpick: you can use `Type.EmptyTypes` here

Popax21 (changes requested):
Looks good, however according to the docs there is also a `StopTextInput` method. I suggest you call that during the game's shutdown procedure as well.

Popax21 (approved):

Issue #597, opened 2023-03-14
Add custom keybinds for the slight camera panning on the map while a level is selected.


AdoHTQ:
### I am planning on implementing this myself

Yes

### Describe your request

![image](https://user-images.githubusercontent.com/19363927/224935185-af70fc35-f6fe-4080-bf44-c35b28587eb3.png)

The slight camera rotation on the mountain view is hard-coded to use the WASD key-binds as shown in the image above. This can get annoying if your menu key-binds contain WASD.

This could be fixed by setting these keys to be what you set your debug camera movement keys to in the mod settings key/controller config. An option could also be added to completely disable this movement as it feels unnecessary but this may be better suited for a separate mod.

### Additional context

I'm going to develop a code-mod that adds the same functionality, but it should probably be added into base Everest as well.

FenyLabs:
Also applies to debug 3d camera movement and clear screen panning I believe

AdoHTQ:
Yes, I think it should apply to clear screen panning. Debug 3d camera movement already had customizable keybinds in everest and they are what I used to determine the other ones.

swoolcock:
@AdoHTQ any update on this?  I'd like to use `MountainAim` for looking around the map in CU2 lobby maps if possible.  Making it customisable should absolutely be in Everest.

AdoHTQ:
I created a PR with the fixed changes. No idea if it's going to be merged. You can compile my branch if you really need the modifications.

I also don't know if the CU2 lobby map uses the MountainAim controls but if it does then it should work with my changes.

PR #600, opened 2023-03-15
Make the debug camera controls also control the slight camera movemen…

AdoHTQ:
…t on the map.

The slight camera rotation on the map is hardcoded to use WASD. This can get annoying for people who have menu buttons bound to WASD. This changes those buttons to be set by the debug camera controls in the Everest mod settings. It also adds default controller controls for the camera movement so that it will still behave like normal if you don't change settings from default.

https://github.com/EverestAPI/Everest/issues/597

### DIFF
--- a/Celeste.Mod.mm/Patches/Input.cs
+++ b/Celeste.Mod.mm/Patches/Input.cs
@@ -2,6 +2,7 @@
 #pragma warning disable CS0649 // Field is never assigned to, and will always have its default value
 
 using Celeste.Mod;
+using Celeste.Mod.Core;
 using Microsoft.Xna.Framework.Input;
 using Monocle;
 using MonoMod;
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/Input.cs
```diff
@@ -22,6 +23,15 @@ static class patch_Input {
                 mod.OnInputInitialize();
 
             Everest.Events.Input.Initialize();
+
+            //Sets the slight camera movement on the map to the set debug camera movement keys in Everest mod settings
+            Input.MountainAim = new VirtualJoystick(
```
Kalobi, line 23: This should probably be done before the event is invoked so that mods can rely on the fact that the event is invoked after Everest's input initialization is complete.

AdoHTQ: I'm not sure if I completely understand what you are saying, but I think I fixed it.

PR #605, opened 2023-03-17
Create scheduled release pipeline for automated rolling releases

coloursofnoise:
# CI Builds
Automated CI builds will be versioned as `0.0.0-<branch>.<build>+azure.<hash>`
- this means that mods cannot depend on dev builds of Everest
- including the build number as a prerelease version allows them to still be ordered properly, and also conveniently gives them a lower priority than the base `0.0.0-dev` version under semver. 

# Release Builds
Release versions will now be a "proper" rolling release, essentially just acting as a buffer for changes to the `dev` branch. They will also now be versioned based on date, since we don't really make a distinction between major and minor changes. 

## Beta
`beta` builds will be released twice a month, and will include the changes from the `dev` branch up until a month before.
They will be versioned as `1.(<yymm>+<offset>).<d>.<incr>+azure.<hash>` where `offset` is an arbitrary number that keeps the resulting minor higher than it is now (for backwards compatibility), and `incr` is just a counter in case of multiple releases on one day. 
- the longest a change will wait before being added to the `beta` branch is 1.5 months.

## Stable
`stable` builds will be released every two months, and will be include the changes from the `dev` branch up until 2 months before. 
They will be versioned the same way as beta builds, but will use the date of the changes they are based off of.
- the longest a change will wait before being added to the `stable` branch is 4 months.

Release builds will be tagged in git as `<branch>-<version>`, omitting the build metadata.


maddie480:
Over time, I ended up *becoming* the automated rolling release bot, so...

PR #606, opened 2023-03-26
Patch FNA float behaviour to behave like XNA

Popax21:
As decided by [this poll](https://discord.com/channels/403698615446536203/1068379326589976596/1089278990105182238), this patches FNA versions of the game to behave like XNA installs in terms of floating point behavior. It does so by patching the `MoveToX/Y` methods of `Actor` and `Platform`, as these are affected by a slight JIT optimization difference causing them to calculate values with more precision on XNA than on FNA. Additionally, it patches `Calc.Approach` for vector types as well as `FinalBossMovingBlock`. XNA's behavior is emulated using `double`s for relevant calculations on all platforms.


DemoJameson (approved):

DemoJameson:
tested with 100% tas, player data is completely consistent with xna

PR #608, opened 2023-04-10
Remove hardcoded debug console keys

Kalobi:
Some people actually use these keys as inputs, or they're close to keys they use and they often accidentally hit them. People who want those keys can add them as bindings manually.


DemoJameson:
maybe change the default key for DebugConsole to OemTilde, since it is the original key in vanilla
https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs#L306

Kalobi:
I think we're purposefully changing the default key to something that every keyboard layout has.

DemoJameson:
open and close may not be the same key
I think we should print a instruction on console that tell the player how to close the console

PR #615, opened 2023-04-19
write the savepath to a file during boot

### LINKED ISSUE #612: external program interop wishlist

Yes

### Describe your request

it'd be nice to have the the save directory written to an `everestsavepath.txt` file on launch, as there's no other way to access that path without knowing launch args, debugrc port, etc.

additionally, having separate files for `CurrentVersion` and `DebugRCPort` would be nice to avoid having to parse the `modsettings-Everest.celeste` yaml (eg. if making a commandline tool)

### Additional context

_No response_

microlith57:
closes #612 by writing the path to the Saves folder to `everest-savepath.txt` in the Celeste folder.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.cs
@@ -415,6 +415,9 @@ namespace Celeste.Mod {
                     Directory.Move(modSettingsOld, modSettingsRIP);
             }
 
+            string savePathFile = Path.Combine(PathEverest, "everest-savepath.txt");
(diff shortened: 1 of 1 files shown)


microlith57:
not sure what the commit history is doing
will force push to fix that tomorrow

microlith57:
ended up not saving the version / debugrc port, as they are already in a settings file and don't need full yaml parsing to retrieve, only a simple string comparison

DemoJameson:
the path is just `Saves` on xna, you might want to convert it to full path by `Path.GetFullPath` method

PR #619, opened 2023-04-26
Better support for transitive dependencies

maddie480:
Using the mod dependency graph from the updater, we can check if any missing mod has dependencies that are also missing, and add them to the missing dependencies list. This way, we can install all transitive dependencies at once.

Here is a test mod: https://maddie480.ovh/celeste/dl?id=TransDependencies1&twoclick=1. It indirectly depends on TransDependencies2 through 4, Collab Utils, Extended Variants and Helping Hand.

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/ModUpdaterHelper.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/ModUpdaterHelper.cs
@@ -29,7 +29,7 @@ namespace Celeste.Mod.Helpers {
 
                 Logger.Log(LogLevel.Verbose, "ModUpdaterHelper", $"Downloading last versions list from {modUpdaterDatabaseUrl}");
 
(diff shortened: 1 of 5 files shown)


Celeste.Mod.mm/Mod/Helpers/ModUpdaterHelper.cs
```diff
@@ -145,10 +188,10 @@ private class MostRecentUpdatedFirst : IComparer<ModUpdateInfo> {
         /// Retrieves the mod updater database location from everestapi.github.io.
         /// This should point to a running instance of https://github.com/max4805/EverestUpdateCheckerServer.
```
Kalobi: While you're there, maybe update the github link

maddie480: good point 😅

Issue #501, opened 2022-08-18
Mod blacklist changes


Mnstrman06:
### I am planning on implementing this myself

No

### Describe your request

More color formatting in the enable/disable mods list, making it easier to sort through your mods.

Changes:

- Pressing the Journal key will "favorite" a mod, changing it to a different color. This can be used to mark mods you want to keep on, such as a utility mod (CelesteTAS, SpeedrunTool, etc) or a skin mod. Utility mods tend to inflate the list which makes it a lot harder to sort through what you do and don't need to have on.

- Enabled mods are automatically changed to a different color (different from the favorited mod color as well).
- Disabled mods stay white.

Color choice for enabled mods or favorited mods are up for debate, but these changes would make it a lot easier to sort through mods.

mayromr:
I want to get your opinion on something I started playing around with.
![fav](https://user-images.githubusercontent.com/32033791/234896502-3470ba6f-fc8f-41c4-8891-b11180b05fa8.png)
Right now pressing the journal key will mark an item as a favorite and it ill be marked in pink, its dependencies will be marked in light pink.

Feel free to play around with it: https://github.com/nhruo123/Everest/tree/fav

PR #620, opened 2023-04-27
Favorite mods in enable/disable mods list

mayromr:
Started working on the changes suggested at:  #501.
TODO:

- [x] Add keyboard hints (or whatever the little key indicator are called).
- [x] Add localizations.
- [x] Test with nested mod dependencies.
- [X] Handle edge case where cyclic mod dependencies will loops forever ( not sure if its even possible to load with Everest ).

Feedback would be appreciated.

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
+++ b/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
@@ -1353,5 +1353,93 @@ namespace Celeste {
             }
         }
 
+        public class SubMenuWithInputs : TextMenu.Item {
+            public object[] Items { get; set; }
+            public bool ControllerMode { get; set; }
+            public Color TextColor { get; set; } = Color.Gray;
+            public Color ButtonColor { get; set; } = Color.White;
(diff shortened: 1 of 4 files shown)


mayromr:
Basic functionalities are done, but this is a huge mess because I am new to celeste modding.
I would love if someone could review this.

Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
```diff
@@ -1353,5 +1353,93 @@ public class BatchModeContext : IDisposable {
             }
         }
 
+        public class SubMenuWithInputs : TextMenu.Item {
```
mayromr: Might be worth to try and implement a `EaseInDecorator` in order to reduce code repetition.
something like this:

``` c#
public class EaseInDecorator<T> : TextMenu.Item, IItemExt where T : TextMenu.Item, IItemExt {
            private T inner;

            public EaseInDecorator(T inner) {
                this.inner = inner;
            }

}
```

Issue #516, opened 2022-09-04
janky rendering when a decal has odd dimensions


microlith57:
currently, decals are drawn with `DrawCentered`, which behaves weirdly if the texture's has odd width or height (leading to the belief that they must have dimensions divisible by 8, which i don't think is actually necessary, only that they are even)
this could be fixed (i think) by simply adding `0.5f` when necessary just before drawing the decal image.

i intend to implement this soon, after #515 is resolved (to avoid merge conflicts)

PR #624, opened 2023-05-02
Fix non-even decal rendering

horizon-wings:
Decal rendering is currently broken on decals whose dimensions are non-even (not a multiple of two). This PR patches rendering calls from decal image components so that they behave as expected.

Decals are drawn with MTexture.DrawCentered(), which centers the decal on its Position. If the center of the decal is non-integer, the pixels of the image are not aligned to integer values, which results in decals being rendered in a way that no longer resembles the source image. My approach is to forcibly offset the rendering position of the image so that the image is "realigned". Any non-rotated decal whose scaled dimensions are even integers should be entirely unaffected; however, decals rotated at non-90 degree angles or with exotic scaling will render differently than before.

I am a little worried about performance since there can be a lot of decals onscreen in heavily-decoed areas, and this patch is adding quite a bit to each decal's Render. Anecdotally, it didn't seem _very_ noticeable in a quick walk through an SJ lobby, but I can see it adding up.

I did not exhaustively test every possible combination of decal attributes, so help with that would be appreciated. I eyeballed a few basic combinations of scale and rotation, as well as checking the vanilla versions of mirror, banner, and core swap decals, and nothing is obviously broken.

This is my first time writing such a large amount of IL and also using MonoMod patching, so correction is appreciated if I did something silly. (In hindsight I probably should have just written a method and called it.)

### DIFF
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -504,6 +504,7 @@ namespace MonoMod {
             cursor.Emit(OpCodes.Ldfld, f_Decal_Rotation);
             // ...and replace the draw call to accept it
             cursor.Emit(OpCodes.Callvirt, m_Draw_new);
+            cursor.Remove();
 
             // go back to the start
             cursor.Index = 0;
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -84,7 +84,6 @@ namespace Celeste {
             }
 
             [MonoModIgnore]
-            [PatchDecalImageRender]
             public extern override void Render();
 
         }
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -110,8 +110,11 @@ namespace Celeste {
             }
 
             public override void Render() {
-                if (activeTextures.Count > 0)
-                    activeTextures[(int) frame % activeTextures.Count].DrawCentered(Decal.Position, Decal.Color, Decal.scale, Decal.Rotation);
+                if (activeTextures.Count > 0) {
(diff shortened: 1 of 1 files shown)


DemoJameson:
@microlith57 can you take a look, since you mentioned the issue in #516

microlith57:
will do

Kalobi:
This should probably also be applied to FlagSwapImage (added by Everest), which also draws decals using DrawCentered. CoreSwapImage (vanilla) also uses it, but is never used for custom decals (the `coreSwap` attribute just creates a FlagSwapImage), so there's no need to add extra code to rendering vanilla core decals.

horizon-wings:
@Kalobi it actually does apply the fix to rendering position for FlagSwapImage, just without the MonoMod attribute - since it's an Everest class, the change is directly in FlagSwapImage.Render(). Though if custom decals never use CoreSwapImage, does that mean the render patch attribute should be removed from there? 

Edit: Whoops, I should test things before pushing them.

Kalobi:
Oh, I somehow missed the change to FlagSwapImage. And yeah, it's probably not the worst idea to remove the patch from CoreSwapImage. The fewer useless patches, the better.

horizon-wings:
Made the CoreSwapImage change since I finally had time to "test" that. @microlith57 if you're able to take a look that would be lovely.

microlith57:
had a look; code looks reasonable, and worked on various non-even decals i tried it on. however it seems recolouring isn't in this PR's history (i think because of the force push), and I think merging it would clobber that.
i'd be ok with merging this after that's fixed :+1:

horizon-wings:
As far as I can tell, decal recoloring _is_ in the PR branch's history? I'm not 100% sure I'm looking in the right place and this is somewhat beyond my git skill level, though.

microlith57:
huh weird, apparently the way i checked out the branch messed things up somehow. sorry for the false alarm i guess
checking it out properly resulted in correct decal recolouring

microlith57 (approved):

PR #631, opened 2023-05-25
Add search to Enable/Disable mods menu

mayromr:
Added support for searching in the Enable/Disable mods menu, This idea was suggested at #320.
Feedback would be appreciated.

https://github.com/EverestAPI/Everest/assets/32033791/c060bbb1-52d1-44bc-b63d-a17ba8599004


mayromr:
After farther comments from Discord I still need to:
- [x] Scale the search icon down if it's overlapping with the text menu
- [x] Display the controls of the search bar inside of it while it has no text
- [x] Finalize the controls for the search bar ( mb add a focus / unfocus )

Issue #626, opened 2023-05-08
Game crash when loading save into a non-existent room


DemoJameson:
### My mods and Everest install are up to date

Yes

### I have recreated the bug with only Everest OR a minimum number of mods enabled

Yes

### Describe the bug

Game crash when loading save into a non-existent room, since session.LevelData is null
Note this crash is also happens withou any mod is enabled
https://discord.com/channels/403698615446536203/1104654739431510096/1104654739431510096

### Steps to reproduce

1. Enter any level then `Save & Quit`
2. Change the name of the room you were in when you saved and reload the map
3. Load into the save

### Expected behavior

Show postcard with error message

### Operating System

Windows 11

### Everest Version

3971

mayromr:
After debugging this for a bit, I found out that this line of code causes the issue:
https://github.com/EverestAPI/Everest/blob/fdc6ddbf3f897799cff2a5e3c0ac72514621d64e/Celeste.Mod.mm/Patches/Level.cs#L228

After failing to load the level, we try to log `Session.LevelData.Name` but `Session.LevelData` is null, so we get `System.NullReferenceException`.

I am not on my windows machine so I can't dig into why is `Session.LevelData` null, I'll look into it later on today.

For now an easy fix would be to remove the log call temporary, the logic to display the postcard is already present.

mayromr:
Might be duplicate of #583.

PR #633, opened 2023-06-11
Fix LoadLevel crashing on missing level

mayromr:
As requested at #626.
In: https://github.com/EverestAPI/Everest/blob/fdc6ddbf3f897799cff2a5e3c0ac72514621d64e/Celeste.Mod.mm/Patches/Level.cs#L228 
we took the name of the level from `Session.LevelData.Name`, but we fail because the level name was changed, now we take the level name directly from the `Session` object.

### DIFF
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -225,7 +225,7 @@ namespace Celeste {
                     }
                 }
 
-                Logger.Log(LogLevel.Warn, "LoadLevel", $"Failed loading room {Session.LevelData.Name} of {Session.Area.GetSID()}");
+                Logger.Log(LogLevel.Warn, "LoadLevel", $"Failed loading room {Session.Level} of {Session.Area.GetSID()}");
                 e.LogDetailed();
                 return;
             }
--- a/Celeste.Mod.mm/Patches/LevelLoader.cs
+++ b/Celeste.Mod.mm/Patches/LevelLoader.cs
@@ -195,7 +195,7 @@ namespace Celeste {
 
         [MonoModIgnore] // We don't want to change anything about the method...
         [PatchLoadingThreadAddEvent] // ... except for manually manipulating the method via MonoModRules
-        [PatchLoadingThreadAddSubHudRenderer] 
+        [PatchLoadingThreadAddSubHudRenderer]
         private extern void LoadingThread();
 
         private void LoadingThread_Safe() {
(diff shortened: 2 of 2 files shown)


DemoJameson:
ig it's better to catch the exception outside of `LoadLevel` (maybe `LevelLoader.StartLevel`), so it can cover the hooked mehtods form mods

the crash mentioned in the issue is thrown from a hooked method, it cannot be caught by the current approach

mayromr:
Right, that makes sense, it would be nice to catch `LoadLevel`  errors from mods as well. I have recreated the error from the discord post with the `SpriteSwapTrigger` it seems that they too read from `Session.LevelData` while it's null. I'll add an additional try catch clause in `StartLevel`.

DemoJameson:
don't know if any mod il hook the `StartLevel` method, but at least not with strawberry jam and its dependencie

DemoJameson:
can you just try catch the `StartLevel()` in `Update()`

mayromr:
Sure, I'll fix it later on today. Just a small question regrading the IL hooks, my current understanding is that function (normal) hooks, are like a stack, e.g a mod hooks into `StartLevel`  at run time it would hook on top of our  `StartLevel` ( the try and catch ) and if it would to throw an exception we wouldn't catch it because the mod is above us in the call stack.
But  with IL hooks, what happens if two different methods modify the same function? Does Monomod just let them both hook the same method and it's the responsibility of the hooks to cooperate?
And in our case, if we assume that a mod IL hooks into `StartLevel` would it hook into the original  `StartLevel` or to the modify version with the try and catch clause?

DemoJameson:
il hooks can modify the same method multiple times, that later can see the former's modifications

> And in our case, if we assume that a mod IL hooks into StartLevel would it hook into the original StartLevel or to the modify version with the try and catch clause?

it modifies the latter, so it breaks mod compatibility

mayromr:
> il hooks can modify the same method multiple times, that later can see the former's modifications
> 
> > And in our case, if we assume that a mod IL hooks into StartLevel would it hook into the original StartLevel or to the modify version with the try and catch clause?
> 
> it modifies the latter, so it breaks mod compatibility

Hey I was thinking about this, and I am not quite sure how can we add/mod features to the base game if even the smallest change breaks compatibility ( even if we change the content of a function we might effect IL Hooks ).
I have a [PR](https://github.com/EverestAPI/Everest/pull/634) that changes a bit of the postcard logic, but that change can cause issues with other mods that might hook into that function.
So I wonder what are the rules for stuff like that, does Everest just breaks some mods from time to time?
 What are the expectations from the mods? it's not like they can pin it themselves to a specific Everest version.
And if Everest does break compatibility by accident who's responsible is it to fix the issue?

( I am not sure if this is the platform for this, sorry in advance if its not )

DemoJameson:
everest is always aware of compatibility
if there are really breaking changes and the mod can't get an update, consider fixes for specific mods in everest
yeah, we are better talk at discord dev channle

PR #654, opened 2023-08-03
Make EntityList Add(entity) and Remove(entity) methods to crash when entity is null

DemoJameson:
Currently if you invoke `level.Add(null as Entity)`, you will get the following useless crash logs
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Monocle.TagLists.EntityAdded(Entity entity)
   at DMD<Monocle.EntityList::UpdateLists>(EntityList this)
   at DMD<Monocle.Scene::BeforeUpdate>(Scene this)
```

Make EntityList `Add(entity)` and `Remove(entity)` methods to crash when entity is null, so modders can catch bugs in time

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/EntityList.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/EntityList.cs
@@ -140,14 +140,12 @@ namespace MonoMod {
 
             ILCursor cursor = new ILCursor(context);
             ILLabel label = cursor.DefineLabel();
-            cursor.Emit(OpCodes.Ldarg_1);
-            cursor.Emit(OpCodes.Ldnull);
-            cursor.Emit(OpCodes.Ceq);
-            cursor.Emit(OpCodes.Brfalse_S, label);
-            cursor.Emit(OpCodes.Ldstr, "entity");
-            cursor.Emit(OpCodes.Newobj, ctor_ArgumentNullException);
-            cursor.Emit(OpCodes.Throw);
-            cursor.MarkLabel(label);
+            cursor.Emit(OpCodes.Ldarg_1)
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/Monocle/EntityList.cs
```diff
@@ -115,5 +131,24 @@ static partial class MonoModRules {
...
+
+            TypeDefinition t_ArgumentNullException = MonoModRule.Modder.FindType("System.ArgumentNullException").Resolve();
+            MethodReference ctor_ArgumentNullException = MonoModRule.Modder.Module.ImportReference(t_ArgumentNullException.FindMethod("System.Void .ctor(System.String)"));
+
+            ILCursor cursor = new ILCursor(context);
+            ILLabel label = cursor.DefineLabel();
+            cursor.Emit(OpCodes.Ldarg_1);
+            cursor.Emit(OpCodes.Ldnull);
+            cursor.Emit(OpCodes.Ceq);
+            cursor.Emit(OpCodes.Brfalse_S, label);
```
mayromr: Not 100% sure if it's semantically the same but can't we replace this with:
```Csharp
cursor.Emit(OpCodes.Brtrue_S, label);
```

mayromr: I **Think** that if someone somehow passes `true` or non-zero c# boxes it so we won't throw false positives on the null check.

DemoJameson: i think they are the same, but use brture is more cleanner

mayromr (approved):
Not sure if it's worth anything but this looks good to me

microlith57 (approved):

PR #657, opened 2023-08-05
Fixed Badeline last node crash

mayromr:
This PR fixes an issue where if the player this the last node of `FinalBoss` (Badeline) the game crashes.

Additinal details:
https://discord.com/channels/403698615446536203/1137280308887175178

### DIFF
--- a/Celeste.Mod.mm/Content/Dialog/English.txt
+++ b/Celeste.Mod.mm/Content/Dialog/English.txt
@@ -27,7 +27,7 @@
 	POSTCARD_MISSINGTUTORIAL=	{big}Oops!{/big}{n}The playback tutorial {#ff1144}((tutorial)){#} could not be found.{n}Check your {#44adf7}log.txt{#} for more info.
 	POSTCARD_TILEXMLERROR=		{big}Oops!{/big}{n}There was an error parsing a tileset in {#ff1144}((path)){#}{n}Check your {#44adf7}log.txt{#} for more info.
 	POSTCARD_XMLERROR=			{big}Oops!{/big}{n}{#ff1144}((path)){#} has a syntax error.{n}Check your {#44adf7}log.txt{#} for more info.
-	POSTCARD_BOSSLASTNODEHIT=	{big}Oops!{/big}{n}FinalBoss entity was hit on its last node, please add an additional node outside of the current room to ensure the player never hits it.
+	POSTCARD_BOSSLASTNODEHIT=	{big}Oops!{/big}{n}{#ff1144}Badeline Boss{#} entity was hit on its last node, please add an additional node outside of the current room to ensure the player never hits it.
 
 # Main Menu
 	MENU_TITLETOUCH= 		TOUCH
--- a/Celeste.Mod.mm/Content/Dialog/English.txt
+++ b/Celeste.Mod.mm/Content/Dialog/English.txt
@@ -27,6 +27,7 @@
 	POSTCARD_MISSINGTUTORIAL=	{big}Oops!{/big}{n}The playback tutorial {#ff1144}((tutorial)){#} could not be found.{n}Check your {#44adf7}log.txt{#} for more info.
 	POSTCARD_TILEXMLERROR=		{big}Oops!{/big}{n}There was an error parsing a tileset in {#ff1144}((path)){#}{n}Check your {#44adf7}log.txt{#} for more info.
 	POSTCARD_XMLERROR=			{big}Oops!{/big}{n}{#ff1144}((path)){#} has a syntax error.{n}Check your {#44adf7}log.txt{#} for more info.
+	POSTCARD_BOSSLASTNODEHIT=	{big}Oops!{/big}{n}FinalBoss entity was hit on its last node, please add an additional node outside of the current room to ensure the player never hits it.
 
 # Main Menu
 	MENU_TITLETOUCH= 		TOUCH
--- a/Celeste.Mod.mm/Patches/FinalBoss.cs
+++ b/Celeste.Mod.mm/Patches/FinalBoss.cs
@@ -1,4 +1,5 @@
 ﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
+﻿#pragma warning disable CS0649 // Field is never assigned to, and will always have its default value
 
 using System;
 using Celeste.Mod;
--- a/Celeste.Mod.mm/Patches/FinalBoss.cs
+++ b/Celeste.Mod.mm/Patches/FinalBoss.cs
@@ -1,6 +1,7 @@
 ﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
 
 using System;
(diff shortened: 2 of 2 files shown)


DemoJameson:
don't modify the `OnPlayer()`, since SJ il hook this method
maybe move your code into `PushPlayer()`

mayromr:
That could work, we could also patch it at the IL level and call some kind of static helper function to disply the card, not sure whats better.

Celeste.Mod.mm/Patches/FinalBoss.cs
```diff
@@ -30,14 +35,29 @@ public patch_FinalBoss(EntityData e, Vector2 offset)
         [PatchBadelineBossOnPlayer] // ... except for manually manipulating the method via MonoModRules
         public new extern void OnPlayer(Player player);
 
+        public extern void orig_PushPlayer(Player player);
```
DemoJameson, line 38:
```suggestion
        private extern void orig_PushPlayer(Player player);
```

Celeste.Mod.mm/Patches/FinalBoss.cs
```diff
@@ -30,14 +35,29 @@ public patch_FinalBoss(EntityData e, Vector2 offset)
...
+
+        public void PushPlayer(Player player) {
```
DemoJameson, line 40:
```suggestion
        private void PushPlayer(Player player) {
```

DemoJameson (changes requested):

Celeste.Mod.mm/Patches/FinalBoss.cs
DemoJameson:
```suggestion
        private extern void orig_PushPlayer(Player player);

        private void PushPlayer(Player player) {
```

DemoJameson (changes requested):

Celeste.Mod.mm/Patches/FinalBoss.cs
mayromr: Weird I just accepted that but it rolled back I'll accept that again

DemoJameson: :) it's me deleted them and merged two suggestions into one

mayromr: Btw there is a new Github feature that allowed me to accept both of the changes at once into the same commit

DemoJameson (approved):

Celeste.Mod.mm/Content/Dialog/English.txt
```diff
@@ -27,6 +27,7 @@
 	POSTCARD_MISSINGTUTORIAL=	{big}Oops!{/big}{n}The playback tutorial {#ff1144}((tutorial)){#} could not be found.{n}Check your {#44adf7}log.txt{#} for more info.
 	POSTCARD_TILEXMLERROR=		{big}Oops!{/big}{n}There was an error parsing a tileset in {#ff1144}((path)){#}{n}Check your {#44adf7}log.txt{#} for more info.
 	POSTCARD_XMLERROR=			{big}Oops!{/big}{n}{#ff1144}((path)){#} has a syntax error.{n}Check your {#44adf7}log.txt{#} for more info.
+	POSTCARD_BOSSLASTNODEHIT=	{big}Oops!{/big}{n}FinalBoss entity was hit on its last node, please add an additional node outside of the current room to ensure the player never hits it.
```
DemoJameson:
```suggestion
	POSTCARD_BOSSLASTNODEHIT=	{big}Oops!{/big}{n}{#ff1144}Badeline Boss{#} entity was hit on its last node, please add an additional node outside of the current room to ensure the player never hits it.
```

DemoJameson: use the name from Lönn instead of class name

DemoJameson (changes requested):

Celeste.Mod.mm/Content/Dialog/English.txt
DemoJameson: use the name from Lönn instead of class name

Celeste.Mod.mm/Patches/FinalBoss.cs
```diff
@@ -1,6 +1,7 @@
 ﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
```
DemoJameson, line 1:
```suggestion
﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
﻿#pragma warning disable CS0649 // Field is never assigned to, and will always have its default value
```

DemoJameson:
> // we want to call LevelEnter.Go explicitly instant of letting CheckForErrors call it to prevent double triggers of postcard
                LevelEnter.Go((Scene as Level).Session, false);

why the postcard will be triggered twice, since CheckForErrors only called once every frame

mayromr:
> > // we want to call LevelEnter.Go explicitly instant of letting CheckForErrors call it to prevent double triggers of postcard
> > LevelEnter.Go((Scene as Level).Session, false);
> 
> why the postcard will be triggered twice, since CheckForErrors only called once every frame

I used to set `patch_LevelEnter.ErrorMessage`  at `OnPlayer` it was getting called after `CheckForErrors` consumed the error message somehow (I'll add a video that shows the issue), I didn't take the time to debug it and I just worked around it.

But now when its on `PushPlayer` that is called from Coroutine `MoveSequence` it does work, so it can be removed.

Is this behaviour with  `OnPlayer` and `patch_LevelEnter.ErrorMessage` expected or should I debug it tomorrow?

```
(08/05/2023 21:51:11) [Everest] [Info] [LevelLoader] Loading bossfight
(08/05/2023 21:51:19) [Everest] [Warn] [FinalBoss] FinalBoss entity was hit on its last node, please add an additional node outside of the current room to ensure the player never hits it.
(08/05/2023 21:51:19) [Everest] [Warn] [FinalBoss] FinalBoss entity was hit on its last node, please add an additional node outside of the current room to ensure the player never hits it.
(08/05/2023 21:51:33) [Everest] [Info] [LevelLoader] Loading bossfight
```

https://github.com/EverestAPI/Everest/assets/32033791/2f4ce2a6-0bf7-41a6-af1e-67c614e0c432

mayromr:
Okay I figured out what's going on. it seems that even after `CheckForErrors` finds an error it keeps updating updating the level, so the state of `patch_LevelEnter.ErrorMessage` is as follows:

1. first update: `patch_LevelEnter.ErrorMessage` get set by `OnPlayer`
2. second update: `patch_LevelEnter.ErrorMessage` gets cleared but `CheckForErrors`
3. second update: `patch_LevelEnter.ErrorMessage` gets set again by a call to `OnPlayer` 

then we leave the level with `patch_LevelEnter.ErrorMessage` already set.
I am not 100% sure about it but I feel like if we just stop updating the scene after we find an error at `CheckForErrors` it will work as expected, If it will I'll open a new PR for that.

PR #659, opened 2023-08-10
Fixed priority getting overwritten by before/after patch

mayromr:
Currently if a `DetourContext` has a `DetourConfig` with a priority value, it would get overwritten even if `Before=*` or `After=*` isn't present on the `DetourConfig` , this PR fixes it by adding an addental check.


mayromr:
This was fixed in the latest core version

PR #661, opened 2023-08-20
Fix strawberry seed crash when player collects the last seed and dies at the same frame

mayromr:
This issue was posted on [Discord](https://discord.com/channels/403698615446536203/1142608955831492799/1142608955831492799),  after an investigation it seems that `StrawberrySeed.Leader` could be null when `OnAllCollected` is called.
This can happen when `Player.Die` gets called and then `this.Leader.LoseFollowers()` gets called which nulls the `Leader` value of all followers, and on the same tick the game detects that the player has all the strawberry seeds and start a `CSGEN_StrawberrySeeds` cutscene, then in the next game tick the cutscene progresses and calls `OnAllCollected` and then  we get a NRE due to the nulled `Leader` value.

This PR fixes the issue by adding a null check before using `Leader` in `OnAllCollected`.

### DIFF
--- a/Celeste.Mod.mm/Patches/StrawberrySeed.cs
+++ b/Celeste.Mod.mm/Patches/StrawberrySeed.cs
@@ -1,5 +1,10 @@
 ﻿using Microsoft.Xna.Framework;
 using System.Linq;
+using MonoMod;
+using System;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using MonoMod.Cil;
 
 namespace Celeste {
     public class patch_StrawberrySeed : StrawberrySeed {
@@ -23,6 +28,10 @@ namespace Celeste {
         private extern void orig_OnPlayer(Player player);
 #pragma warning restore CS0626
 
+        [MonoModIgnore]
+        [PatchPatchStrawberrySeedOnAllCollected]
+        private extern new void OnAllCollected();
+
         private void OnPlayer(Player player) {
             orig_OnPlayer(player);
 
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/StrawberrySeed.cs
```diff
@@ -23,6 +28,10 @@ public patch_StrawberrySeed(Strawberry strawberry, Vector2 position, int index,
         private extern void orig_OnPlayer(Player player);
 #pragma warning restore CS0626
 
+        [MonoModIgnore]
+        [PatchPatchStrawberrySeedOnAllCollected]
+        private extern new void OnAllCollected();
```
DemoJameson:
```suggestion
        public new extern void OnAllCollected();
```

DemoJameson (changes requested):

DemoJameson (approved):

Celeste.Mod.mm/Patches/StrawberrySeed.cs
mayromr: Should we always keep the original acsses access modifier of the pached method?

Celeste.Mod.mm/Patches/StrawberrySeed.cs
```diff
@@ -38,3 +47,42 @@ public patch_StrawberrySeed(Strawberry strawberry, Vector2 position, int index,
...
+            cursor.Emit(OpCodes.Brtrue, beforeLoseFollower);
+            // on null discard duplicated value and jump over the calling path
+            cursor.Emit(OpCodes.Pop);
+            cursor.Emit(OpCodes.Br, afterLoseFollower);
+            
+            cursor.GotoNext(MoveType.Before, inst => inst.MatchLdarg(0));
+            cursor.MarkLabel(beforeLoseFollower);
+            
+            cursor.GotoNext(MoveType.After, inst => inst.MatchCallvirt("Celeste.Leader", "LoseFollower"));
+            cursor.GotoNext();
```
DemoJameson: this `cursor.GotoNext();` seems unnecessary

mayromr: You are right, it is

PR #663, opened 2023-08-27
Fixed Theo up transition crash

### LINKED ISSUE #644: theo crystals don't prevent up-transitions while not held
### describe the bug

currently, theo crystals block right-transitions when not held, but not up-transitions.
when transitioning upwards while there is an unheld theo crystal in the room, it dies because it is below the current room's bounds; this kills the player, and crashes the game with a nullref (because the transition expects the player to be non-null).
with assist mode invincibility, instead of crashing this teleports the theo crystal upwards into the new room.

### steps to reproduce

1. place two rooms vertically above each other, with a spawnpoint in each, and a theo crystal in the lower one
2. transition from the lower room to the upper one while the crystal is not being held

### expected behavior

instead of crashing or teleporting the crystal, the player should simply bonk off the transition, similarly to a right-transition under the same circumstances.

mayromr:
Currently if the player transition to a level above the current one while a Theo crystal is placed in the level, the game crash.
This PR fixes the issue by blocking the player from transitioning to the level above if a Theo crystal is present in the level and isn't held by the player.
This patch stays faithful to the original game by letting the player transition with any holdable item ~~totally not because then it's a headache to search which Theo crystal the player is holding or something~~.

This fixes #644.


Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -780,5 +802,22 @@ ldc.i4.s 9
...
+            new ILContext(method).Invoke(il => {
+                ILCursor curser = new(il);
+
+                curser.GotoNext(MoveType.After,
+                    instr => instr.MatchCallvirt("Monocle.Tracker", "GetEntity"),
+                    instr => instr.MatchStloc(2));
+                curser.Emit(OpCodes.Ldloc_2);
+                curser.Emit(OpCodes.Ldarg_1);
+                curser.Emit(OpCodes.Ldloc_0);
+                curser.Emit(OpCodes.Call, m_BlockUpTransitionsWithoutHoldable);
```
maddie480:
```suggestion
                ILCursor cursor = new(il);

                cursor.GotoNext(MoveType.After,
                    instr => instr.MatchCallvirt("Monocle.Tracker", "GetEntity"),
                    instr => instr.MatchStloc(2));
                cursor.Emit(OpCodes.Ldloc_2);
                cursor.Emit(OpCodes.Ldarg_1);
                cursor.Emit(OpCodes.Ldloc_0);
                cursor.Emit(OpCodes.Call, m_BlockUpTransitionsWithoutHoldable);
```

PR #664, opened 2023-09-03
Add colored terminal logging support

psyGamer:
Adds colored terminal logging, if the terminal supports it.
The log file doesn't contain any color escape codes.

This especially helps while developing mods, because it makes non-info logs stand out.
The following colors are used, can be changed to any other ANSI color if desired:
- Verbose: Purple
- Debug: Blue
- Info: White
- Warn: Yellow
- Error: Red

There's a new CLI flag: `--color`, which controls whether to use ANSI colors in the termianl output.
- `on`: Force enable ANSI colors
- `off`: Force disable ANSI colors
- `auto`(default): Tries to enable color support on Windows and assums that termials on MacOS/Linux support colors.
While the assumption is a bit naive, there isn't really an easy way to figure it out and most terminals atleast support 4bit ANSI colors, which this is using.

Additionally there are new shorthands for logging at a specific level, since they are a nice QoL feature and it felt fitting to include them with this PR. This deprecates `Logger.Log(tag, msg)`, which seems to be quite extensivly used, but it doesn't make sense to keep using it when `Logger.Verbose(tag, msg)` exists and does the exact same thing.

If these shorthands are accepted I'd push a commit, replacing every `Logger.Log(LogLevel...` inside Everest, but havn't done so yet, to keep the diff reasonable sized.


Celeste.Mod.mm/Mod/Everest/Logger.cs
```diff
@@ -82,13 +92,51 @@ public static class Logger {
...
+        /// <param name="str">The string / message to log.</param>
+        public static void Error(string tag, string str)
+            => Log(LogLevel.Error, tag, str);
+
         /// <summary>
         /// Log a string to the console and to log.txt
         /// </summary>
         /// <param name="tag">The tag, preferably short enough to identify your mod, but not too long to clutter the log.</param>
         /// <param name="str">The string / message to log.</param>
+        [Obsolete("Use Logger.Verbose instead")]
```
JaThePlayer: Imo it really doesn't make sense to deprecate this, as it doesn't hurt anyone to keep it around.

JaThePlayer (commented):
I'm quite concerned about this making logging even slower due to the double write and additional string allocations, and as such I'd like for this to be **opt-in** via an in-game setting. (since it's entirely useless for 99% of players)

psyGamer:
I can add an in-game options. Should the CLI argument be removed in favour of this?
EDIT: Also, does it make sense to keep auto-detection around? I'd say no?
EDIT 2: The double write was there the entire time, just inside LogWriter and not Logger.

psyGamer:
I understand that most players won't use this, however I fear that those who would use it wouldn't know about it existing when it's disabled by default.
Maybe having the auto-state also check if the user has a console open would be a good idea, but IDK how this could even be implemented besides checking for `--console` on Windows. The problem with that is that it's not the only option to have a termianl open.

JaThePlayer:
Alright, checked with a profiler and it's not that much worse than before, I forgot that the double write always happened ⛄. I'm fine with it being on by default in this case, but I'm still glad it's an in-game setting though.

psyGamer:
Alright, I assume i can remove the `--color` CLI option again in that case?
Also can I push the commit for replacing `Logger.Log(LogLevel.Info, ...)` with `Logger.Info(...)` (and all other levels), since it'll kinda bloat the diff?

JaThePlayer:
Yea the CLI option can go, I can't see a use for it tbh. As for the mass-replace, I don't really mind personally, but I'm kind of worried about potential merge conflicts with the core branch with so many changes 🤔

psyGamer:
There seem to be about 240 instances, but I don't think core changes many of logs so I don't think it should be an issue.
However what should be accounted for when merging into core is replacing new log statements as well and maybe change some exception logs as well (see 816b77a)
EDIT: Also when merging into core, PlatformHelper can be replaced with RuntimeInformation.

psyGamer:
I wonder if there should be an option to customize the colors. I don't think supporting truecolor would be worth it, but maybe limiting it to 4bit ANSI colors?
However that should probably only be an option inside the settings file and not in-game, since it would just clutter the core module settings.

psyGamer:
I wonder if the `(11/18/2023 19:34:13) [Everest]` prefix should maybe be gray?

maddie480:
Merge conflicts to fix

Celeste.Mod.mm/Mod/Everest/Everest.cs
```diff
@@ -437,6 +437,10 @@ public static partial class Everest {
             Loader.ProcessAssembly(core.Metadata, asm, types);
             core.Metadata.RegisterMod();
 
+            if (CoreModule.Settings.ColorizedLogging && RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && !Logger.TryEnableWindowsVTSupport()) {
+                Logger.Error("core", "Failed to enalbe Windows VT support!");
```
maddie480: enable* :suspicioushiro:

Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
```diff
@@ -266,6 +267,19 @@ public class CoreModuleSettings : EverestModuleSettings {
             }
         }
 
+        private bool _ColorizedLogging = true;
```
maddie480, line 270: Are we OK with this being opt-_out_? 🤔

Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
```diff
@@ -266,6 +267,19 @@ public class CoreModuleSettings : EverestModuleSettings {
...
+        [SettingSubText("MODOPTIONS_COREMODULE_COLORIZEDLOGGING_DESC")]
+        [SettingInGame(false)]
+        public bool ColorizedLogging { 
+            get => _ColorizedLogging;
+            set {
+                if (value && RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && !Logger.TryEnableWindowsVTSupport()) {
+                    Logger.Error("core", "Failed to enalbe Windows VT support!");
```
maddie480: enable* again

maddie480 (changes requested):
Also, conflicts again :ohnoshiro:

Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
maddie480: I got referred to a previous conversation: yes we are OK because there's no performance impact

PR #669, opened 2023-09-23
Patch out hardcoded music in fall routine

XMinty77:
Patch out the reflection music that's hardcoded in vanilla to start playing after the fall cutscene

### DIFF
--- a/Celeste.Mod.mm/Patches/Player.cs
+++ b/Celeste.Mod.mm/Patches/Player.cs
@@ -7,6 +7,7 @@ using Microsoft.Xna.Framework;
 using Monocle;
 using MonoMod;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using Mono.Cecil;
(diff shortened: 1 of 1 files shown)


XMinty77:
Potential issue: DJMapHelper patches this method ( [here](https://github.com/DemoJameson/CelesteDJMapHelper/blob/693fbea3405090517ad66217b809673ff2725206/Entities/BadelineBoostDown.cs#L11) ) and on top of patching out the hardcoded music it also removes the two second delay present at the start of the vanilla routine. If this patch is removed from DJMapHelper or DJMapHelper is removed from being a dependency from maps that need its music patch and is not loaded on the game this can create TAS desyncs and/or speedrun time discrepancies.

Kalobi:
That's kinda on DJ Map Helper tbh. The bigger concern here is maps that weren't using DJ Map Helper and might be relying on the music change.

XMinty77:
That's true, what can we do about that?

XMinty77:
How about conditionally patching based on required Everest version? That sounds terribly confusing, but it's the only thing that comes to mind.

microlith57:
it might be worth making it a metadata option, though that does create the concern of too much clutter in the metadata window

Kalobi:
Also I kinda don't want to add new metadata options until #503 finally gets merged :catplant:

XMinty77:
I can work on that at notice. Please update this thread as fit, thanks.

PR #671, opened 2023-10-01
add toggle option to core mode trigger

microlith57:
this adds an `CoreModeTrigger.Modes` enum, which is the same as `Session.CoreModes` but with an extra `Toggle` option at the end.
this is used for the `mode` attribute, and when it is set to `Toggle`, the trigger toggles the core mode.
this means that when it is cold, it becomes hot, and vice versa; nothing happens if the core mode is `None`.


microlith57:
@Kalobi pointed out that this could in theory break existing core mode triggers; as such **do not merge** until this is tested

microlith57:
now tested with existing triggers!

PR #672, opened 2023-10-09
Fix module loading

Kalobi:
Everest module loading made some very strange assumptions, leading to every non-code mod crawling the Celeste assembly and multi-module assemblies being crawled multiple times. This PR attempts to fix that. Should be tested in depth. Monika's D Sides, which has multiple modules in one assembly, seems to load fine, but my attempts to test by running the full TAS were hampered by the TAS not fully syncing even on stable.


Kalobi:
SJ TAS syncs, so nothing is horribly broken about module loading. Several of the D side files synced on various versions, but a full sync check isn't possible because the files are all for different D side versions.

PR #676, opened 2023-10-15
make extensions case insensitive for type guessing and texture reloading

microlith57:
**do not merge until asset reloading is properly tested**

currently, files such as `image.PNG` are not recognised as PNG images, whereas `image.png` are.
this PR makes the type guesser and texture reloader case insensitive for extensions, so those files would both be recognised as PNGs.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
@@ -17,7 +17,7 @@ using System.Threading;
 namespace Celeste.Mod {
 
     /// <summary>
-    /// Special meta type for assets. 
+    /// Special meta type for assets.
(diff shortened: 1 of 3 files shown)


microlith57:
aparently `OrdinalIgnoreCase` is better than `InvariantCultureIgnoreCase`; i'll change that

Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
```diff
@@ -709,25 +709,33 @@ public static class Content {
             public static void Add(string path, ModAsset metadata)
                 => TryAdd(path, metadata);
 
+            private static readonly string[] METADATA_YAML_NAMES = {
```
Kalobi, line 712: metadata.yaml and multimetadata.yaml are no longer supported, I just forgot to remove them from the type guesser apparently

SnipUndercover:

PR #677, opened 2023-10-16
implement type fullname comparer for tracker dictionaries

microlith57:
currently, the tracker uses `Dictionary<Type, List<T>>` to track entities and components, which breaks debuggers (and the pooler works similarly, so probably has the same problem).
i'm not completely sure _why_ it breaks debuggers, but i believe it is to do with the debugger version of the type is a different instance to the one in the celeste process.
this means that when running eg. `level.Tracker.GetEntity<Player>()` in the debugger console, it fails to find the key in the dictionary.

this PR fixes that, by adding a new `TypeNameEqualityComparer` class that also takes into account the full name of the type.
this is then used to create the tracker/pooler dictionaries via a common IL patch.

~~it may be worth conditionally applying this change, based on whether `--debugger` is passed, but i don't think this is a performance concern.~~

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/TypeNameEqualityComparer.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/TypeNameEqualityComparer.cs
@@ -1,0 +1,19 @@
+using Monocle;
+using System;
+using System.Collections.Generic;
+
+namespace Celeste.Mod
+{
+    /// <summary>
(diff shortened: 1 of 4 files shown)


microlith57:
might be worth applying similar changes to other `Type` collections:
- `StrawberryRegistry` berry type cache?
- the `Tracker` `HashSet`s?

microlith57:
turns out the motivating issue seems not to exist anymore? on latest dev i can access `Tracker` contents from the debugger without issue

PR #685, opened 2023-10-18
Add 'Toggle Debug Console' hotkey, '~' as default key

DemoJameson:


Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
```diff
@@ -320,9 +320,13 @@ public class CoreModuleSettings : EverestModuleSettings {
...
 
         [SettingSubHeader("MODOPTIONS_COREMODULE_DEBUGMODE_SUBHEADER")]
+        [SettingInGame(false)]
+        [DefaultButtonBinding(0, Keys.OemTilde)]
+        public ButtonBinding ToggleDebugConsole { get; set; }
+
         [SettingInGame(false)]
         [DefaultButtonBinding(0, Keys.OemPeriod)]
-        public ButtonBinding DebugConsole { get; set; }
+        public ButtonBinding OpenDebugConsole { get; set; }
```
Kalobi: Renaming the option will get rid of people's existing bindings, we should just change the display name via dialog files instead.

PR #686, opened 2023-11-01
Catch exceptions when reading a module from the cache, and retry

Brokemia:
We get occasional issues that are due to cache corruption. Looking at the stack traces, most of them seem to originate from inside this particular call to ModuleDefintion.ReadModule. If we catch these exceptions, we can hopefully continue loading using the non-cached dll, rather than crashing as usual.


Brokemia:
No one seems to know if this is an issue still or if core fixes it. I'm closing the PR for now and we'll see if I need to redo it later

PR #687, opened 2023-11-02
add ability to override tileset ignores

microlith57:
adds an `ignoreOverrides` attr that overrides tileset `ignores`.
allows you to do eg.
```xml
<Tileset id="T" copy="z" path="example_tileset" ignores="*" ignoreOverrides="h,l"/>
```


JaThePlayer (approved):
Feel like TerrainType.Ignore could be heavily optimized, but ig that can be handled later

PR #689, opened 2023-11-07
Updates EntityData and EntityIDs for easier code access and Entity-EntityData relationships

vivianlonging:
UPDATE - This rewrite edits a lot of the previous changes to focus more on the relationships between Entity, EntityData, and EntityID. This is going to just be a writeup of all the changes made from what was previously added, inclusive of all the previous changes, to help contextualize the code (since there's a lot to read)

Changes:
Entity now gets a reference to the EntityData that constructed it `Entity.EntityData`. This covers all of the confirmable cases where an EntityData is used to construct an Entity.
EntityIDs are now implemented as a field of EntityData and modified within LevelData.CreateEntityData. The reason this is useful is explained more in-depth [here](https://docs.google.com/document/d/1kxbmGAh3ICYhCaYJ6mssORw-YgUqaU_Jsj9ahYlI3cY)
All EntityIDs used for loading entities use `EntityData.EntityID` within Everest's source code. It's not possible to edit external code sources doing things manually, so this is as good as we can get.
Trigger ID Offset is now managed in LevelData.CreateEntityData, since EntityID is a field of EntityData
LevelExt has a method to load any entity, vanilla or modded, from EntityData `level.LoadEntity(EntityData, ?LevelData, ?Vector2?)`
Added TypeHelper, which contains:
- Type Cache so that mods don't need to contain their own individual dictionaries and can look up values from the same cache
- Map between EntityData Name and Type (covering all values that can be confirmed)
- Methods for checking an Entity via content from its EntityData


Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -598,33 +598,62 @@ public static class Loader {
...
+
                             gen = type.GetMethod(genName, new Type[] { typeof(Level), typeof(LevelData), typeof(Vector2), typeof(EntityData) });
                             if (gen != null && gen.IsStatic && gen.ReturnType.IsCompatible(typeof(Entity))) {
-                                loader = (level, levelData, offset, entityData) => (Entity) gen.Invoke(null, new object[] { level, levelData, offset, entityData });
+                                loader = (level, levelData, offset, entityData) => {
+                                    Entity e = (Entity) gen.Invoke(null, new object[] { level, levelData, offset, entityData });
+                                    (e as patch_Entity).EntityData = entityData;
+                                    return e;
+                                };
                                 goto RegisterEntityLoader;
```
JaThePlayer: This branch is missing this check 
```cs
if (!TypeHelper.entityDataNameToType.ContainsKey(id))
    TypeHelper.entityDataNameToType[id] = type;
```
It also seems like that check can safely be extracted to after the `RegisterEntityLoader` label, I'd highly prefer that to avoid issues like these

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
```diff
@@ -0,0 +1,316 @@
...
+                ["moonGlitchBackgroundTrigger"] = typeof(MoonGlitchBackgroundTrigger),
+                ["blackholeStrength"] = typeof(BlackholeStrengthTrigger),
+                ["rumbleTrigger"] = typeof(RumbleTrigger),
+                ["birdPathTrigger"] = typeof(BirdPathTrigger),
+                ["spawnFacingTrigger"] = typeof(SpawnFacingTrigger),
+                ["detachFollowersTrigger"] = typeof(DetachStrawberryTrigger)
+            };
+        }
+
+        public static void Add_DataNameTypeMap(string name, Type type, bool overwrite = false) {
```
JaThePlayer: This naming scheme is quite weird, especially for a public method. I'd also like proper documentation for all new public methods.

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
```diff
@@ -0,0 +1,316 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Celeste.Mod.Helpers {
+    public static class TypeHelper {
+
+        public static Dictionary<string, Type> typeFullNameToTypeCache;
+
+        public static Dictionary<string, Type> entityDataNameToType;
```
JaThePlayer: Are these supposed to be public? The naming scheme suggests otherwise. If they are meant to be public, please make them readonly properties

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -21,5 +38,26 @@ class patch_Entity : Entity {
         internal void _PreUpdate() => PreUpdate?.Invoke(this);
 
         internal void _PostUpdate() => PostUpdate?.Invoke(this);
+
+        public EntityID __EntityID => new EntityID(EntityData.Level.Name, EntityData.ID);
```
JaThePlayer: Very weird name for a public getter

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -1,16 +1,33 @@
...
+        public extern void ctor(Vector2 position);
+
         public new Scene Scene {
             [MonoModIgnore]
             get;
             [MonoModIgnore]
             private set;
         }
 
+        public EntityData EntityData;
```
JaThePlayer, line 31: Maybe make this a `{ get; internal set; }`?

Celeste.Mod.mm/Patches/Monocle/Entity.cs
```diff
@@ -21,5 +38,26 @@ class patch_Entity : Entity {
...
+        public EntityID __EntityID => new EntityID(EntityData.Level.Name, EntityData.ID);
+    }
+}
+
+namespace MonoMod {
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchEntityCtor))]
+    class PatchEntityCtorAttribute : Attribute { }
+
+    static partial class MonoModRules {
+        public static void PatchEntityCtor(ILContext context, CustomAttribute attrib) {
```
JaThePlayer, line 55: Some comment explaining what this patch does would be nice

Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -3,10 +3,16 @@
...
+        [MonoModIgnore] // We don't want to change anything about the method...
+        [MonoModConstructor]
+        public patch_Session(AreaKey area, string checkpoint = null, AreaStats oldStats = null)
+            : base(area, checkpoint, oldStats) {
+            // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
+        }
+
+        public bool RestartedFromGolden;
 
         public extern void orig_ctor(AreaKey area, string checkpoint = null, AreaStats oldStats = null);
```
JaThePlayer: What's the point of this change?

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -360,10 +367,12 @@ public class LoadOverride {
         /// </summary>
         /// <param name="entityData"></param>
         /// <param name="level">The level to add the entity to.</param>
+        /// <param name="levelData">If you want to set a specific LevelData for the Entity spawn to, do so here.</param>
+        /// <param name="roomOffset">If you want to set a specific World Offset for the Entity to spawn at, do so here.</param>
         /// <returns></returns>
-        public static bool LoadCustomEntity(EntityData entityData, Level level) {
-            LevelData levelData = level.Session.LevelData;
-            Vector2 offset = new Vector2(levelData.Bounds.Left, levelData.Bounds.Top);
+        public static bool LoadCustomEntity(EntityData entityData, Level level, LevelData levelData = null, Vector2? roomOffset = null) {
```
JaThePlayer, line 382: Wouldn't this change break binary compatibility? Might be worth adding a `LoadCustomEntity(EntityData entityData, Level level)` overload that calls this method.
Also, is there any particular reason for these additional arguments?

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -602,6 +611,792 @@ public class LoadOverride {
 
             return errorPresent;
         }
+
+        public static void LinkEntityToData(Entity entity, EntityData data) {
```
JaThePlayer: Should this be public?

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -602,6 +611,792 @@ public class LoadOverride {
...
+            EntityID entityID = new EntityID(levelData.Name, iD);
+            if (Session.DoNotLoad.Contains(entityID)) {
+                temporaryEntityData = null;
+                return false;
+            } else if(LoadCustomEntity(entity3, this, levelData, roomOffset)) {
+                temporaryEntityData = null;
+                return true;
+            } else {
+
+                Vector2 vector = roomOffset ?? new Vector2(levelData.Bounds.Left, levelData.Bounds.Top);
```
JaThePlayer, line 697: Uhhh considering this massive copy-paste, is `orig_LoadLevel` patched to call this method? I'd really prefer not to have this massive switch in two places at once

Celeste.Mod.mm/Mod/Everest/Everest.cs
```diff
@@ -463,7 +463,7 @@ public static partial class Everest {
             }).Register();
 
             LuaLoader.Initialize();
-
+            TypeHelper.entityDataNameToType = new();
```
JaThePlayer: This should just be done inside of BakeVanillaEntityData (or maybe `TypeHelper.Load()` or something simillar)

JaThePlayer (changes requested):
Very nice PR feature-wise, but I would definitely like documentation for all newly added public methods/fields/getters.

Celeste.Mod.mm/Mod/Everest/Everest.cs
JaThePlayer: Actually, this line completely cancels out `BakeVanillaEntityData`, making it impossible to use TypeHelper on vanilla entities

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
```diff
@@ -0,0 +1,316 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Celeste.Mod.Helpers {
+    public static class TypeHelper {
+
+        public static Dictionary<string, Type> typeFullNameToTypeCache;
```
JaThePlayer: This field is never set and is always null, makes `typefinder` crash when provided a c# type

Celeste.Mod.mm/Patches/Session.cs
vivianlonging: This was just added because it crashed when I was building. No actual reason other than, it's the same functionality and it makes the code work

Celeste.Mod.mm/Patches/Level.cs
vivianlonging: additional arguments were requested by coders

Celeste.Mod.mm/Patches/Level.cs
vivianlonging: No. LoadEntity is meant for code to add an entity to the Scene manually. It does not get called in orig_LoadLevel

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
vivianlonging: This check is not guaranteeable in this specific case. I'll add a note in comments in its stead though.
Situationally, this arises because you can have this situation
```cs
[CustomEntity("VivHelper/CustomDashBlock = Load")]
public class CustomDashBlock : DashBlock {
    public static Entity Load(Level level, LevelData levelData, Vector2 offset, EntityData entityData) => entityData.Bool("disableFallingBlocksBreak", false) ? new CustomDashBlockAlt(entityData, offset, new EntityID(levelData.Name, entityData.ID)) : new CustomDashBlock(entityData, offset, new EntityID(levelData.Name, entityData.ID));
    ...
}
```
Which class should it be described as?

Celeste.Mod.mm/Mod/Everest/Everest.cs
vivianlonging: that was meant to be the Type.FullName -> Type cache

Celeste.Mod.mm/Patches/Level.cs
vivianlonging: Update: This method will be removed in favor of implementing it directly into Entity:.ctor

Celeste.Mod.mm/Patches/Level.cs
vivianlonging: Also, there are technically differences between LoadEntity and the entity-loading method inside orig_LoadLevel, after investigating the internals of the method, there are so many mods that already hook into orig_LoadLevel for that switch that modifying it breaks too many things at once for me to make it worth it.

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
vivianlonging: Made them readonly, but I'm not sure why properties would be necessary here. As a static readonly object they're only constructed, and if there was getter/setter rules about adding data to the map then I could see a use case (and I believe there is a structure construction that allows for that, but we clearly don't need it here.

vivianlonging:
Some technical notes:
I put LoadVanillaEntity into its own method because:
1. There's the possibility of mods already hooking into orig_LoadLevel's massive switch statement and the nightmare of deconstructing that mess would suck
2. LoadVanillaEntity does a number of additional things to ensure that loading an Entity actually works, rather than the switch statement. Rewriting it for a boolean variable is a sufficient alternative.
3. LoadVanillaEntity loads both Entities and Triggers, which would end up slowing things down slightly.
In addition, I'm not really sure how hefty the space cost of having "2 big switch statements" (specifically, CIL hashtable switch statements, since that's how C# encodes it) in source actually is? If someone could enlighten me on that it'd be greatly helpful in understanding how this method will end up working on the low-level side of things. Given that, I would suspect that "LoadVanillaEntity" isn't exactly the most commonly used method I'd assume it would be okay to do so, but maybe I'm mistaken.

I moved all of the stuff in Everest.Boot over to the static class constructor because 
1. It resolves the `readonly` accessor issue that would have come up later
2. after investigating it, turns out this way could be very fast under the hood ? Might not end up being fast due to a potential Core dll compatibility, and might have misunderstood the information I saw, but it looked like by instantiating the Dictionary, under the CLR, it might be able to instantiate the Dictionary's Hashtable in a faster way?

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
```diff
@@ -213,37 +208,85 @@ public static class TypeHelper {
...
+        /// <summary>
+        /// "Safe" variant of GetType. Attempts to get the type (class) based on the *full name* of the Type provided. Use this in your `if` statements, instead of GetType
+        /// </summary>
+        /// <param name="fullname">The Full Name (Name including namespaces) of the Type you are trying to find. Example: Celeste's Level class is 'Celeste.Level'</param>
+        /// <param name="type">The type as a class found. Returns null if not found</param>
+        /// <param name="cache">Set this to false if you have no intention of obtaining this type with this method after your first use.</param>
+        /// <returns>Whether or not a type (class) was able to be found given the provided FullName</returns>
+        public static bool TryGetType(string fullname, out Type type, bool cache = true) {
+            if (FullName_to_Type.ContainsKey(fullname)) {
+                type = FullName_to_Type[fullname];
```
JaThePlayer: `TryGetValue` to avoid doing 2 dictionary lookups here

cso.dat
```diff
@@ -0,0 +1,8 @@
+V1
+2:PatchEntityCtor,v
+2:context,u
+2:attrib,Q
+2:f_EntityData,B
+2:f_Level_temporaryEntityData,M
+2:cursor,G
+2:instr,g
```
JaThePlayer: ?

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -615,8 +617,8 @@ public static class Loader {
                                     (e as patch_Entity).EntityData = entityData;
                                     return e;
                                 };
-                                if (!TypeHelper.entityDataNameToType.ContainsKey(id))
-                                    TypeHelper.entityDataNameToType[id] = type;
+                                if (!TypeHelper.EntityDataName_to_Type.ContainsKey(id))
+                                    TypeHelper.EntityDataName_to_Type[id] = type;
```
JaThePlayer: the `ContainsKey` here is redundant (applies to other places in this method as well)

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
```diff
@@ -213,37 +208,85 @@ public static class TypeHelper {
...
+        /// Appends a Dictionary describing a relationship between an entity Data Name to an object Type. This should *only* be used in situations where you are using a Load method in your CustomEntityAttribute,
+        /// or where you do not use a CustomEntityAttribute at all, but passing the name will always return an object of the same class. 
+        /// </summary>
+        /// <param name="entityDataNameToType">The Dictionary featuring the relationship between names and types.</param>
+        /// <param name="overwrite">Whether or not you want to override previous data added to the Dictionary.</param>
+        public static void Link_DataNames_To_Types(Dictionary<string, Type> entityDataNameToType, bool overwrite = false) {
             foreach(KeyValuePair<string,Type> kvp in entityDataNameToType) {
                 if(overwrite || !entityDataNameToType.ContainsKey(kvp.Key))
                     entityDataNameToType[kvp.Key] = kvp.Value;
             }
```
JaThePlayer: This function seems to do nothing? I think the last line was supposed to write to a different dictionary

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
```diff
@@ -213,37 +208,85 @@ public static class TypeHelper {
             };
         }
 
-        public static void Add_DataNameTypeMap(string name, Type type, bool overwrite = false) {
-            if (overwrite || !entityDataNameToType.ContainsKey(name))
-                entityDataNameToType[name] = type;
+        public readonly static Dictionary<string, Type> FullName_to_Type;
+
+        public readonly static Dictionary<string, Type> EntityDataName_to_Type;
```
JaThePlayer: Using underscores in the middle of names is not the C# naming scheme. `FullNameToType` would be better. (same goes for all methods in this cass)

I still don't think this needs to be public, especially not publicly mutable - `CheckEntityByDataName` and `Link_DataName_To_Type` exists already for `EntityDataName_to_Type`, `FullName_to_Type` is just a cache for `TryGetType`.
Only publicly exposing a `IReadOnlyDictionary` view for `EntityDataName_to_Type` would be much more reasonable (albeit only useful if you need to enumerate through it, for whatever reason) and would leave more space for refactoring later. `FullName_to_Type` being public seems useless, imo.

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
JaThePlayer: nvm, it is slightly different. `TryAdd` would be equivalent I think

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
JaThePlayer: right, I misread that a bit, oops

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
vivianlonging: At first I thought TryGetValue and ContainsKey used two different lookup methods but that's apparently either just wrong or some .NET FW jank so yeah I'll update all of these scenarios

cso.dat
vivianlonging: This was done by Visual Studio and honestly I have no fucking clue.

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
vivianlonging: Was unaware that was a method lol

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
vivianlonging: In practice this code is supposed to do this, so I'll rewrite it as this.
```cs
if(overwrite) entityDataNameToType[kvp.Key] = kvp.Value;
else entityDataNameToType.TryAdd(kvp.Key, kvp.Value);
```
(if the key already exists dont add it to the list unless `overwrite` is true)

Celeste.Mod.mm/Mod/Helpers/TypeHelper.cs
vivianlonging: I agree with the nomenclature fix, but I was initially trying to make it more legible, since some words like EntityData use capitals anyways which makes it a bit more annoying to parse for non-English speakers.

JaThePlayer (approved):

Kalobi:
The random changes to Everest.cs and ClutterBlockGenerator.cs should be reverted (usually I'd say with an amend -> force push to prevent the history from being cluttered, but this is gonna have to be squash merged anyway).

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -600,31 +600,57 @@ public static class Loader {
...
+                                    return e;
+                                };
+                                // You cannot consistently determine the Type from the Method construction, so you can't map EntityDataName to Type from this method.
+// example: Load(Level level, LevelData levelData, Vector2 offset, EntityData entityData) => entityData.Bool("legacy") ? new ExampleClassOld(entityData, offset) : new ExampleClassNew(entityData, offset);
                                 goto RegisterEntityLoader;
                             }
 
                             ctor = type.GetConstructor(new Type[] { typeof(EntityData), typeof(Vector2), typeof(EntityID) });
                             if (ctor != null) {
-                                loader = (level, levelData, offset, entityData) => (Entity) ctor.Invoke(new object[] { entityData, offset, new EntityID(levelData.Name, entityData.ID + (patch_Level._isLoadingTriggers ? 10000000 : 0)) });
```
Kalobi, line 610: You're losing the _isLoadingTriggers patch here.

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -736,7 +1536,6 @@ static partial class MonoModRules {
 
             // Reset to apply entity patches
             cursor.Index = 0;
-
             // Patch the winged golden berry so it counts golden deaths as a valid restart
```
Kalobi: Please readd all these blank lines you keep removing for no reason.

Kalobi (changes requested):

vivianlonging:
UPDATE - This rewrite edits a lot of the previous changes to focus more on the relationships between Entity, EntityData, and EntityID. This is going to just be a writeup of all the changes made from what was previously added, inclusive of all the previous changes, to help contextualize the code (since there's a lot to read)

Changes:
Entity now gets a reference to the EntityData that constructed it `Entity.EntityData`. This covers all of the confirmable cases where an EntityData is used to construct an Entity.
EntityIDs are now implemented as a field of EntityData and modified within LevelData.CreateEntityData. The reason this is useful is explained more in-depth [here](https://docs.google.com/document/d/1kxbmGAh3ICYhCaYJ6mssORw-YgUqaU_Jsj9ahYlI3cY)
All EntityIDs used for loading entities use `EntityData.EntityID` within Everest's source code. It's not possible to edit external code sources doing things manually, so this is as good as we can get.
Trigger ID Offset is now managed in LevelData.CreateEntityData, since EntityID is a field of EntityData
LevelExt has a method to load any entity, vanilla or modded, from EntityData `level.LoadEntity(EntityData, ?LevelData, ?Vector2?)`
Added TypeHelper, which contains:
- Type Cache so that mods don't need to contain their own individual dictionaries and can look up values from the same cache
- Map between EntityData Name and Type (covering all values that can be confirmed)
- Methods for checking an Entity via content from its EntityData

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
vivianlonging: _isLoadingTriggers is now refactored to LevelData._isRegisteringTriggers 
We already make the assumption that LevelData.Triggers is always registering triggers, and now that EntityData contains its own ID per Data, we can assign the 10000000 offset to the ID in EntityData

maddie480:
Status unclear after some messages on Discord

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -709,31 +1524,85 @@ static partial class MonoModRules {
...
+
             ILLabel continueLabel = null;
             cursor.GotoNext(MoveType.After, instr => instr.MatchBrtrue(out continueLabel));
             // add
             // || _IsInDoNotLoadIncreased(levelData, trigger)
             // to if condition for continue to handle triggers that already add 10000000 to their DoNotLoad entry
             cursor.EmitLdarg0();
-            cursor.EmitLdloc(v_levelData);
-            cursor.EmitLdloc(v_entityData);
+            cursor.EmitLdloc(3); // levelData
```
Kalobi, line 1607: Why are you making the patch less readable by replacing variable references with magic numbers?

Celeste.Mod.mm/MonoModRules.Mod.cs
```diff
@@ -38,5 +48,17 @@ static partial class MonoModRules {
...
+        public static void PatchModSpawnEntityData(ILContext context, CustomAttribute attrib) {
+            FieldReference f_Entity_EntityData = MonoModRule.Modder.Module.GetType("Monocle.Entity").FindField("EntityData");
+            ILCursor cursor = new ILCursor(context);
+
+            cursor.GotoNext(MoveType.Before, i => i.MatchCall<Scene>("Add"));
+            cursor.Emit(OpCodes.Dup);
+            cursor.Emit(OpCodes.Ldarg_0);
+            cursor.Emit(OpCodes.Ldfld, f_Entity_EntityData);
+            cursor.Emit(OpCodes.Stfld, f_Entity_EntityData);
+        }
```
maddie480, line 62: I think we put the IL patches directly in the same file as the one they are used in now :thinking:

maddie480: ah it's used at multiple places, I see

Celeste.Mod.mm/Patches/FireBall.cs
```diff
@@ -15,6 +21,10 @@ public patch_FireBall(Vector2[] nodes, int amount, int index, float offset, floa
             // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
         }
 
+        [MonoModIgnore]
+        [PatchModSpawnEntityData] // handled in MonoModRules.Mod.cs
+        public override extern void Added(Scene scene);
```
maddie480, line 26: does this mean there are new requirements on how you initialize entities? :cold_sweat:

vivianlonging: Not really?
In effect, all this Patch does is, for vanilla entities where one entity spawns a series of other entities, the EntityData from the one entity is passed to all the other entities

vivianlonging:
Explanation for Everest PR [#689](https://github.com/EverestAPI/Everest/pull/689):

The goal of the PR, very simply, is to make object matching a lot easier and more consistent, by ensuring:
1. Coders should have a consistent way to determine if an Entity has specific EntityData information about it, rather than relying on code-deterministic variables.
2. Every EntityData should have a unique EntityID. This seems like it already exists, but it actually doesn’t. Currently, every Entity with an EntityID will have a resulting ID of
`[Name of room loaded into when spawned]:[integer ID relative to LevelData the global entity is from, regardless of if that is the loaded room]`, and this isn’t unique, in the scenario where any Entity from a different LevelData is loaded than the Session.LevelData currently set. This appears to be a vanilla oversight that doesn’t actually cause any problems in vanilla, since there’s never a time you load something from a different room, but for mods we have to pay attention to it. (This actually means that sometimes you can stop the loading of a completely different Entity as well, because loading entity with ID 1 from room `b` globally when you are in room `a` means you disable `a:1` instead of `b:1`.)
3. We should also expose various features for tracking the relationship between EntityDataName and Entity, since a few mods currently use equivalent systems, or would benefit from using this. (VivHelper, SSMHelper, EeveeHelper, FrostHelper, just to name a few) Conglomerating the List into a single list merely reduces memory bloat.

My solutions for these problems:
1.  Every Entity should have the EntityData that created it attached to it, *if* an EntityData was used to create it. This is relevant for a number of “object-generic” situations, i.e. “filter all colliding entities for the Type of the created entity’s Full Name matching a specified string” → “filter all colliding entities for the EntityDataName matching a specified string,” as well as tracking objects for things like CelesteTAS
1a. There are a few ways to attach the EntityData to the Entity that cover various numbers of cases. 
	Vanilla entities - this requires a manual patch for every vanilla entity, but I just wrote a metaprogramming script that does it and cleaned it up so it always works.
	EntityLoader overwrite: each EntityLoader in Level has a string [EntityDataName] and a Func<Level, LevelData, Vector2, EntityData, Entity> [EntityLoader]. By setting up the function as the original contents of the delegate, followed by `entity.EntityData = entityData;` we can add the EntityData to the Entity.
	Level.LoadEntity - this case is an unsolvable scenario, in that there is no way to predetermine whether or not 1 or more entities are loaded from Level.LoadEntity relate to the EntityData being passed, unless…
	Hook *every constructor from a class extending Entity with an EntityData parameter* to postprocess add the EntityData to the Entity - this is complete overkill and I’m choosing to not do this.
	Spawning Entities - For things like Fireball, where One Entity is created and then that spawns a set number of other Entity copies of itself, the reference for the Entity should be preserved. This is done with the PatchSpawnEntityData patch 
2. EntityIDs should be registered at the LevelData level, rather than registering as they are loaded into a level. This ensures that the EntityID will be of the form 
3. The EntityLoaders in Level already has a relationship for EntityDataName → EntityLoader, so if the Loader constructor does not have a delegated relationship (e.g. `[CustomEntity(“VivHelper/RainbowSpikesUp=LoadUp”)]`) we know that the class constructor is directly related to some specific EntityDataName string, meaning we can add it to the list. This actually covers every case besides vanilla, so simply creating the list with all vanilla, will just work here.

maddie480:
It has been agreed on Discord that this PR should be split into smaller ones

maddie480:
since this PR will be split up, I'm closing this one

PR #694, opened 2023-11-14
Fix swapped A and B on certain Nintendo Switch Pro Controllers

### LINKED ISSUE #675: Switch Pro Controller buttons are swapped on Linux

Yes

Yes

### Describe the bug

Button inputs from Switch Pro Controllers are switched around like this:
- A -> B
- B -> A

### Steps to reproduce

1. Connect a Switch Pro Controller and launch Celeste with Everest
2. Navigate UI, play a map, or try to change controller bindings. Inputs will not match the pressed buttons.

### Expected behavior

The detected input should match the button pressed on the controller.

### Operating System

Linux (kernel 6.5.7 with no special controller drivers)

### Everest Version

Bug introduced in 4231

None

### Additional context

_No response_

Chedski:
Fixes #675 by switching the button IDs for this specific model of pro controller (`030056fb7e0500000920000010026803` in `gamecontrollerdb.txt`) on Linux to the correct numbers


Popax21:
I'm not sure how I feel about editing `gamecontrollerdb.txt` directly in Everest - if this truly is an issue like that and not just a controller configuration problem it should be fixed in upstream SDL2 instead of in Everest.

microlith57:
according to [this issue](https://github.com/libsdl-org/SDL/issues/5151) this isn't considered a bug, and the appropriate solution is to use the controller's config to swap them

microlith57:
if that's our consensus too, is this ok to close?

Chedski:
> according to [this issue](https://github.com/libsdl-org/SDL/issues/5151) this isn't considered a bug, and the appropriate solution is to use the controller's config to swap them

This controller doesn't have any configuration options, though? It's a first-party Switch Pro Controller.

Chedski:
Also, that issue references both A/B and X/Y being switched, and on MacOS. This issue is only present on Linux and for this *specific* controller, and it only swaps A/B, not X/Y.

microlith57:
> This controller doesn't have any configuration options, though? It's a first-party Switch Pro Controller.

in that case i believe the bug would be the controller being misidentified as a third-party one

Chedski:
> in that case i believe the bug would be the controller being misidentified as a third-party one

Are you sure? The controller DB identifies it as "Nintendo Switch Pro Controller", and third-party controllers seem to be identified with the brand name.

microlith57:
could it be the case that the 1st and 3rd part controllers share an ID, and because the 3rd party ones are rebindable the mapping was defined incorrectly in SDL?
if so that would be a valid SDL bug IMO.

Popax21:
Closing since this isn't a bug in Everest itself, and the `core` branch has been merged into `dev` and will soon be deleted.

PR #697, opened 2023-11-26
Various load time improvements

JaThePlayer:
Shaves off a few seconds from loading times by optimizing various frequently called methods.

Most of the timesave is in `BinaryPacker`, achieved by providing starting capacities for lists and dictionaries so that they don't need to be resized all the time.

`RunLengthEncoding.Decode` got optimized to use a shared buffer instead of a `StringBuilder`, and now also accepts a `ReadOnlySpan<byte>` as an argument. From my testing, the shared buffer doesn't exceed 1MB, so it probably isn't worth it to clear it after map loading is finished. In the future, it might be worth it to make an internal class storing various shared buffers to be able to reuse them.

`Tracker.Initialize` ends up calling `FakeAssembly.GetTypesSafe` several times, so this PR introduces a temporary cache inside the `Tracker` to store the results of that call, saving ~200ms. Might be worth it to figure out caching `FakeAssembly.GetTypesSafe` directly, though that would require figuring out when the cache needs to be cleared...

`StrawberryRegistry`'s methods get called for each entity in each map in everest's `CoreMapProcessor`, taking a lot of time. A cache using `HashSet`s for them got introduced.

`BinaryPacker.Element.Attr*` methods were written *horribly* previously, always doing 2 dictionary lookups and a `string.ToLower`, even if that was unnecessary. They got rewritten now to return early.

A few other places got optimized slightly with spans, string interpolation instead of `string.Format`, and `Array.Empty<T>`, nothing major though.

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/Tracker.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/Tracker.cs
@@ -34,16 +34,40 @@ namespace Monocle {
     }
 
     class patch_Tracker : Tracker {
-        [MonoModIgnore]
-        public static extern List<Type> GetSubclasses(Type type);
+        // A temporary cache to store the results of FakeAssembly.GetFakeEntryAssemblies
+        // Set to null outside of Initialize.
+        private static Type[] _temporaryAllTypes;
+
+        private static Type[] GetAllTypesUncached() => FakeAssembly.GetFakeEntryAssembly().GetTypesSafe();
+        
+        [MonoModReplace]
+        private static List<Type> GetSubclasses(Type type) {
+            bool shouldNullOutCache = _temporaryAllTypes is null;
+            _temporaryAllTypes ??= GetAllTypesUncached();
+            
+            List<Type> subclasses = new();
+            foreach (Type otherType in _temporaryAllTypes)
+            {
+                if (type != otherType && type.IsAssignableFrom(otherType))
+                    subclasses.Add(otherType);
+            }
+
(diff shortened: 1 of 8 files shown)


Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -34,16 +34,40 @@ public class TrackedAsAttribute : Attribute {
...
+        private static List<Type> GetSubclasses(Type type) {
+            bool shouldNullOutCache = _temporaryAllTypes is null;
+            _temporaryAllTypes ??= GetAllTypesUncached();
+            
+            List<Type> subclasses = new();
+            foreach (Type otherType in _temporaryAllTypes)
+            {
+                if (type != otherType && type.IsAssignableFrom(otherType))
+                    subclasses.Add(otherType);
+            }
```
swoolcock, line 53: Could this be done with LINQ instead?  I'm not sure what the difference in allocations or stack size would be.

```cs
List<Type> subclasses = _temporaryAllTypes
    .Where(otherType => type != otherType && type.IsAssignableFrom(otherType)
    .ToList();
```

swoolcock (commented):
LGTM, just one potential change.

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
JaThePlayer: Linq is always going to be slower for this, unfortunately

PR #698, opened 2023-12-05
add attachToSolid attr to spikes

microlith57:
spinners have it, why not spikes
allows you to put them near clouds etc and not have them get picked up

### DIFF
--- a/Celeste.Mod.mm/Patches/Spikes.cs
+++ b/Celeste.Mod.mm/Patches/Spikes.cs
@@ -1,6 +1,4 @@
-﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
-
-using Celeste.Mod.Helpers;
+﻿using Celeste.Mod.Helpers;
 using Microsoft.Xna.Framework;
 using Mono.Cecil;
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/Spikes.cs
```diff
@@ -9,11 +11,21 @@
...
             : base(position, size, direction, type) {
             // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
         }
 
+        public extern void orig_ctor(EntityData data, Vector2 offset, Directions dir);
+
+        [MonoModConstructor]
+        public void ctor(EntityData data, Vector2 offset, Directions dir) {
+            orig_ctor(data, offset, dir);
+            if (data.Has("attachToSolid") && !data.Bool("attachToSolid")) {
```
JaThePlayer: This causes two dictionary lookups, you can make use of the 2nd parameter to Bool instead.
Imo it would also be nice to inject this condition directly into the ctor, both for speed and also to not introduce another orig_ method, but its not the end of the world

JaThePlayer (approved):

PR #700, opened 2023-12-13
Add DecalRegistryHandler and refactor all everest handlers to use it

JaThePlayer:
Currently, decal registry handlers perform XML object parsing for every decal instance, which is inefficient. This also makes it harder to cache things which need a bit more complex handling.

This PR adds a new way of creating decal registry handlers, by extending from this class and calling `DecalRegistry.AddPropertyHandler<T>()`:
```cs
    public abstract class DecalRegistryHandler {
        /// <summary>
        /// The name of this handler, which should match the xml tag name which creates this handler.
        /// This field will be accessed in <see cref="DecalRegistry.AddPropertyHandler{T}()"/>, as well as in error handling.
        /// </summary>
        public abstract string Name { get; }

        /// <summary>
        /// Parses the xml entry for an instance of this decal registry property.
        /// This function gets called once per class instance, while parsing the xml file.
        /// Use it to cache attributes from xml into fields on this handler instance.
        /// </summary>
        public abstract void Parse(XmlAttributeCollection xml);

        /// <summary>
        /// Applies this handler to the given decal.
        /// This function gets called for each decal, so make sure to do as much work as possible in <see cref="Parse"/> instead.
        /// </summary>
        public abstract void ApplyTo(Decal decal);
}
```

This class also exposes a variety of helper methods for easier parsing of the `XmlAttributeCollection`

### DIFF
--- a/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
+++ b/Celeste.Mod.mm/Mod/Registry/DecalRegistry.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Xna.Framework;
+﻿using Celeste.Mod.Registry.DecalRegistryHandlers;
+using Microsoft.Xna.Framework;
 using Monocle;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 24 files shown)


Kalobi:
Closing and reopening to rerun CI

PR #702, opened 2023-12-17
Fix Open Log File Folder not handling paths with spaces

swoolcock:
Attempting to open a log file path with spaces (for example, `~/Library/Application Support/itch/apps/celeste/Celeste.app/...`) would do nothing due to the space in "Application Support".

Using `ArgumentList` means that .NET will automatically escape arguments if required.

Tested on macOS, should work fine on Windows and Linux but untested.


swoolcock:
~~I'm going to go through and check for more instances of this, because I think it might also fix Everest sometimes not restarting itself.~~
Looks like the other places `Arguments` is used are all escaped, so this is the only place that needs fixing.

Popax21:
LGTM

PR #712, opened 2023-12-29
Generate a publicized vanilla DLL while keeping Everest untouched

psyGamer:
This creates a new `Celeste.public.dll`, which has all vanilla types/fields/methods publicizied, but keep Everest stuff untouched. This has the benefit of not requiring the publicizier in mods and it avoids accidently publicizing Everest internal stuff.

I'm not sure how to properly include this new DLL inside the lib-stripped release, but that would need to be done to access it inside a CI run.

### DIFF
--- a/MiniInstaller/MiniInstaller.csproj
+++ b/MiniInstaller/MiniInstaller.csproj
@@ -21,6 +21,8 @@
   <!-- Set up Piton AppHosts -->
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.HostModel" Version="3.1.16" />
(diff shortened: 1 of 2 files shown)


JaThePlayer:
Might be worth adding a `[OrigVisibility(...)]` attribute to all changed types (like the bepinex publicizer), so that you can easily see the real visibility of a given type/field/method from the IDE's decompiler?

psyGamer:
Closed in favor of doing this as a MSBuild step at the mod level. See [EverestAPI/CelestePublicizer](https://github.com/EverestAPI/CelestePublicizer).

PR #715, opened 2024-01-02
CassetteListener component

### LINKED ISSUE #475: Component-based CassetteBlockManager updates

Yes

### Describe your request

`CassetteBlockManager` currently gets all the `CassetteBlock`s through the tracker and executes methods on them in time with the rhythm, etc.  Code mods should be able to react to these events without hooking `CassetteBlockManager`.

Ideally you'd add a tracked component to your entity that has actions triggered by the cassette block manager, as well as a cassette index and activation property.

### Additional context

Rough example:

```cs
[Tracked]
public class CassetteListener : Component {
    public int Index;
    public bool Activated;

    public Action Finish;
    public Action WillToggle;
    public Action<bool> ActivateSilently;
    public Action<bool> SetActivated;

    public CassetteListener(int index) : base(false, false) {
        Index = index;
    }
}

// add component to your entity
Add(new CassetteListener(cassetteIndex) {
    WillToggle = () => {
        // called from CassetteBlockManager in the same fashion as CassetteBlock.WillToggle()
    }
});
```

swoolcock:
Resolves https://github.com/EverestAPI/Everest/issues/475

This is a fairly simple component for code modders to create cassette-based entities.  The component can be assigned delegates that are called in a similar fashion to `CassetteBlock` methods, as if the component were actually a cassette block.

Example entity:

```cs
public class SampleEntity : Entity {
    private Sprite sprite;

    public SampleEntity(EntityData data, Vector2 offset) : base(data.Position + offset) {
        Color activeColor = Color.White;
        Color toggleColor = activeColor * 0.75f;
        Color inactiveColor = activeColor * 0.3f;

        Add(sprite = GFX.SpriteBank.Create("sampleEntity"));

        Add(new CassetteListener(new EntityID(data.Level.Name, data.ID), data.Int("cassetteIndex"), data.Float("tempo", 1f)) {
            OnStart = activated => sprite.Color = activated ? activeColor : inactiveColor,
            OnActivated = () => sprite.Color = activeColor,
            OnDeactivated = () => sprite.Color = inactiveColor,
            OnWillActivate = () => sprite.Color = toggleColor,
            OnWillDeactivate = () => sprite.Color = toggleColor,
            OnFinish = () => sprite.Color = inactiveColor,
        });
    }
}
```
<br>

The `EntityID` is optional, and is used to prevent `OnStart` being called multiple times, in the same way that `CassetteBlock.SetActivatedSilently` is handled.

`EntityAdded` will attempt to create a `CassetteBlockManager` if required, using the same logic as `Level.LoadLevel`.

https://github.com/EverestAPI/Everest/assets/2331297/2a9bfdf4-c449-4f70-9e8e-1a1e3544bfb8

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/CassetteListener.cs
+++ b/Celeste.Mod.mm/Mod/Entities/CassetteListener.cs
@@ -54,8 +54,17 @@ namespace Celeste.Mod.Entities {
         /// </summary>
         public Modes Mode;
 
-        public CassetteListener(int index) : base(false, false) {
+        /// <summary>
+        /// Matches the functionality of <see cref="CassetteBlock.ID"/>.
+        /// </summary>
+        public EntityID ID;
+
+        public CassetteListener(int index) : this(index, EntityID.None) {
+        }
+
+        public CassetteListener(int index, EntityID id) : base(false, false) {
             Index = index;
+            ID = id;
         }
 
         /// <summary>
--- a/Celeste.Mod.mm/Mod/Entities/CassetteListener.cs
+++ b/Celeste.Mod.mm/Mod/Entities/CassetteListener.cs
@@ -1,0 +1,119 @@
+using Monocle;
+using System;
+
+namespace Celeste.Mod.Entities {
+    [Tracked]
+    public class CassetteListener : Component {
+        /// <summary>
+        /// Called by <see cref="Level.LoadLevel"/> when loading a room.
+        /// The parameter indicates whether this component will be the active one.
+        /// </summary>
+        public Action<bool> OnStart;
+
+        /// <summary>
+        /// Called by <see cref="CassetteBlockManager.StopBlocks"/> after collecting a cassette.
+        /// </summary>
+        public Action OnFinish;
+
+        /// <summary>
+        /// Called by <see cref="CassetteBlockManager.SetWillActivate"/> if <see cref="Index"/> is about to become current.
+        /// For a <see cref="CassetteBlock"/>, this is represented by a non-collidable block moving one pixel up.
+        /// </summary>
+        public Action OnWillActivate;
+
+        /// <summary>
+        /// Called by <see cref="CassetteBlockManager.SetWillActivate"/> if <see cref="Index"/> is about to no longer be current.
+        /// For a <see cref="CassetteBlock"/>, this is represented by a collidable block moving one pixel down.
+        /// </summary>
+        public Action OnWillDeactivate;
+
+        /// <summary>
+        /// Called by <see cref="CassetteBlockManager.SetActiveIndex"/> if <see cref="Activated"/> was changed from false to true.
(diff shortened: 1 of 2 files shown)


l-Luna:
entities should be able to listen for cassette beats of all indexes without registering multiple components (after the CBM has been added, too, to get the actual max index). i think ideally, the delegates would just gain an `int` index parameter, maybe with a "simple" version of the component for entities that don't need that.

they should also be able to *listen* for cassette beats without actually forcing cassette music; maybe by giving them an index of -1, and ignoring those components when setting up the CMB.

both of ^ could be grouped together by triggering the component on every beat if the index is -1, and letting it access the actual current beat via the tracker, but that's less clean & makes it more annoying to react to multiple beats *and* start the music & would be incompatible with custom CBMs that don't inherit from the vanilla one (which would be good to have).

swoolcock:
The way I'd originally done this is with a struct that contains the entire state of the `CassetteBlockManager` at that tick, such as the current/next/previous index, beat, sixteenth, etc. and have it pass that to actions based on what kind of change happened at the CBM level.

This was a lot harder using hooks since it needed to check all the CBM fields every frame, whereas making it part of Everest will be much simpler.

It'd be something like:
```cs
public struct CassetteBlockManagerState {
    // initial configuration
    public int BeatsPerTick;
    public int TicksPerSwap;
    public float Tempo;

    // changes as the music progresses
    public int CurrentBeat;
    public int CurrentIndex;
    public int CurrentTick;
    public int CurrentSixteenth;
}

public class CassetteListener {
    Action<CassetteBlockManagerState> OnBeat;
    Action<CassetteBlockManagerState> OnSwap;
    Action<CassetteBlockManagerState> OnWillToggle;
    Action<CassetteBlockManagerState> OnTick;
    // potentially more actions
```
<br>
We could include filtering by index etc. as this PR progresses and we decide exactly how we want it to work.

swoolcock:
Copypasta from Discord:
> luna deer, @ for GOD TIER — Today at 06:47
I'm v busy at the moment, but my initial thoughts is that passing everything about the CBM to listeners on every event feels very wasteful, especially for initial configuration options; for those in particular, it feels like any custom CBM that tries to change them over time will conflict with any listener that actually tries to use them, so either they should actually be made available only at initialization, or just not included here.
on initialization, might be good having an explicit OnStart event that gets run when e.g. entering a room or respawning, where vanilla would just use a silent update?

I'm getting the vibe that the ultimate goal is to refactor and retrofit vanilla CBM so that players can do whatever they want with it.  I know @Vexatos was working on something similar.  Do you want to weigh in here?

mayromr:
hey, is this going to get merged soon? This will be very useful for an entity I want to make.

swoolcock:
I'm happy to continue with this but I'm not sure of the direction we're taking it.  Personally I think this component is *enough* for now (with some tweaks), and we should look at replacing/retrofitting the vanilla system in the future.
I also would like to get *something* in as soon as possible, since it's blocking me from porting the paintbrushes.

PR #721, opened 2024-01-08
Autotiler cleanup and performance improvements

JaThePlayer:
Started out as optimisation, ended up with bugfixes and cleanup as well 😅

- Added the ability to disable automatic sorting of `<set>` elements, as it can be unintuitive and makes it harder to create complex tilesets.
```xml
<Tileset id="A" path="..." sortMode="none">
```

- All code paths of `TileHandler` now use the Everest version of autotiler code - improves performance, simplifies the code and makes it much easier to implement new optimizations.
- Introduced constants which explain the meanings for magic numbers within the `byte[] Masked.Mask`
- Fixed a potential bug when using non-ASCII chars in `GetByteLookup`, used in `<define>` elements - now throws an exception instead of potentially not working or overwriting other custom masks in hard to predict ways. This should be fine, as I honestly don't know if anybody used this feature at all ☃️
- Fixed the everest `TileHandler` being completely broken (it used to ignore the `ignores="..."` xml tag...
- Big performance improvements for maps with tons of rooms - f5 in 9d went from ~4.5s to ~2.8s for me
  * Introduced a cache for tileset masks
  * Stopped using `CheckForSameLevel` potentially several times per tile, as it's O(n) complexity (n - number of rooms in the map). Instead, we calculate which room the currently checked tile comes from (still O(n), but only once per tile), and simply check if the surrounding tiles belong to the same room (O(1))
  * Various micro-optimisations combined with code cleanup

### DIFF
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -112,18 +112,18 @@ namespace Celeste {
                                 foreach (char c in text) {
                                     switch (c) {
                                         case '0':
-                                            masked.Mask[i++] = 0; // No tile
+                                            masked.Mask[i++] = patch_Masked.TileEmptyMask;
                                             break;
                                         case '1':
-                                            masked.Mask[i++] = 1; // Tile
+                                            masked.Mask[i++] = patch_Masked.TilePresentMask;
                                             break;
                                         case 'x':
                                         case 'X':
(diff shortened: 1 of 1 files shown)


SnipUndercover:
Due to a lack of activity, we decided to close this PR. Please reopen the PR or make a new one if these fixes and features will be picked up again in the future.

PR #722, opened 2024-01-08
Bump MonoMod submodule

Kalobi:
Need to test whether this works with the CI build and my local build is just broken or whether there's actual issues with this


Kalobi:
Latest Mono.Cecil version seems to break stuff

PR #723, opened 2024-01-08
Add DoNotLoad check with base EntityID for triggers

Kalobi:
Trigger entity IDs are increased by 10000000 in vanilla, but Everest's custom entity handling does not do this, making DoNotLoad not work properly. Patching it into the custom entity handling would be very difficult with the way entity loaders work, and some entities already rely on the current behavior, so this PR adds a secondary DoNotLoad check that uses the non-increased EntityID.


Kalobi:
Changed to instead give the increased entity ID to triggers and check the data ID + 10000000 and + 20000000 in the DoNotLoad check. This avoids breaking any maps with duplicate entity IDs shared between entities and triggers. ShroomHelper will have to be adjusted because it manually constructs its EntityIDs, but it's currently broken anyway, so that's not a big deal.

PR #737, opened 2024-01-28
Add Player.GetStarFlyTimer()

piggynl:
Hi,

I'm trying to add a meter for starFlyTimer if Madeline is star flying in [Stamina Meter](https://gamebanana.com/mods/34280) mod, but I found `player.starFlyTimer` is private. This PR adds a public function `Player.GetStarFlyTimer()` to expose that value. I have no experience in C# and MonoMod, so please kindly tell me if there is another better way to expose a private field.

Thanks!

### DIFF
--- a/Celeste.Mod.mm/Patches/Player.cs
+++ b/Celeste.Mod.mm/Patches/Player.cs
@@ -21,6 +21,7 @@ namespace Celeste {
 
         // We're effectively in Player, but still need to "expose" private fields to our mod.
         private bool wasDashB;
+        private float starFlyTimer;
         private HashSet<Trigger> triggersInside;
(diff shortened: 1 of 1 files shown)


Kalobi:
There's a lot of private fields in vanilla that modders might want to access. Generally you should use reflection, DynamicData, or an assembly publicizer to access those. See the wiki entry on [accessing private fields and methods](https://github.com/EverestAPI/Resources/wiki/Making-Code-Mods#accessing-private-fieldspropertiesmethods) for one of those options. While Everest can certainly provide public ways to access them if necessary, I don't think that's required here.

piggynl:
Thank you for your detailed explanation! I'll use these methods instead.

PR #739, opened 2024-02-05
Add progress bar to splash

solstice23:
image

Possible enhancement: adding a numeric indicator showing `(loaded/total)`

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -196,7 +196,7 @@ namespace Celeste.Mod {
                     LoadZip(Path.Combine(PathMods, file));
                 }
                 foreach (string dir in dirs) {
-                    LoadZip(Path.Combine(PathMods, dir));
+                    LoadDir(Path.Combine(PathMods, dir));
                 }
 
                 enforceOptionalDependencies = false;
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -176,27 +176,34 @@ namespace Celeste.Mod {
 
                 enforceOptionalDependencies = true;
 
-                string[] files = Directory.GetFiles(PathMods);
-                Array.Sort(files); //Prevent inode loading jank
-                for (int i = 0; i < files.Length; i++) {
-                    string file = Path.GetFileName(files[i]);
-                    if (!file.EndsWith(".zip") || !ShouldLoadFile(file))
-                        continue;
+                string[] files = Directory
+                    .GetFiles(PathMods)
+                    .OrderBy(f => f) //Prevent inode loading jank
+                    .Select(Path.GetFileName)
+                    .Where(file => file.EndsWith(".zip") && ShouldLoadFile(file))
+                    .ToArray();
+                   
+                string[] dirs = Directory
+                    .GetDirectories(PathMods)
+                    .OrderBy(f => f) //Prevent inode loading jank
+                    .Select(Path.GetFileName)
+                    .Where(file => file != "Cache" && ShouldLoadFile(file))
+                    .ToArray();
+
+                EverestSplashHandler.SetSplashLoadingModCount(files.Length + dirs.Length);
+
+                foreach (string file in files) {
                     LoadZip(Path.Combine(PathMods, file));
                 }
-
-                files = Directory.GetDirectories(PathMods);
-                Array.Sort(files); //Prevent inode loading jank
-                for (int i = 0; i < files.Length; i++) {
-                    string file = Path.GetFileName(files[i]);
-                    if (file == "Cache" || !ShouldLoadFile(file))
-                        continue;
-                    LoadDir(Path.Combine(PathMods, file));
+                foreach (string dir in dirs) {
+                    LoadZip(Path.Combine(PathMods, dir));
                 }
 
                 enforceOptionalDependencies = false;
                 Logger.Log(LogLevel.Info, "loader", "Loading mods with unsatisfied optional dependencies (if any)");
                 Everest.CheckDependenciesOfDelayedMods();
+
+                EverestSplashHandler.AllModsLoaded();
 
                 watch.Stop();
                 Logger.Log(LogLevel.Verbose, "loader", $"ALL MODS LOADED IN {watch.ElapsedMilliseconds}ms");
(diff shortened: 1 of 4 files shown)


Wartori54:
Good addition, looks solid!, tho i have to say one thing: you forgot to remove a `Stream.ReadLine()` on line 169, thus causing the splash to timeout on each boot. Otherwise this LGTM.
As a side note: if i were you I'd try to stop your IDE from messing with the white spaces everywhere, it makes it really hard to see when a line was actually changed. And if you did that intentionally, please do it as two separate commits :).
About the text rendering, i already have planned that, alongside a bunch of other changes and features so don't worry about that.

solstice23:
Fixed it! And sorry for the space change, it was not intended, should I revert it?

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -176,28 +176,35 @@ public static class Loader {
 
                 enforceOptionalDependencies = true;
 
-                string[] files = Directory.GetFiles(PathMods);
-                Array.Sort(files); //Prevent inode loading jank
-                for (int i = 0; i < files.Length; i++) {
-                    string file = Path.GetFileName(files[i]);
-                    if (!file.EndsWith(".zip") || !ShouldLoadFile(file))
-                        continue;
+                string[] files = Directory
```
Kalobi, line 179: Did you profile these LINQ refactors? LINQ tends to allocate a bunch more objects than loop-based approaches, including intermediate enumerators and an entire second array at the end here, which you then also iterate through a second time because the loading now happens in a secondary pass after filtering.

solstice23: To obtain the accurate total number of mods to be loaded, we need to iterate it first before loading the files. I don't think this can be a performance bottleneck for just preprocessing an array of strings.

Also LINQ is even somehow faster than the for method.

image

<details>
  <summary>The test code</summary>

```
using System;
using System.Linq;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using Microsoft.Diagnostics.Runtime;


namespace MyBenchmarks
{

    public class LINQVsFor
    {
        private const int N = 10000;
        private readonly byte[] data;

        private string[] Whitelist = Array.Empty<string>(), Blacklist = Array.Empty<string>(), TemporaryBlacklist = Array.Empty<string>();

        private bool ShouldLoadFile(string file)
        {
            return Whitelist != null ? Whitelist.Contains(file) : (!Blacklist.Contains(file) && (TemporaryBlacklist == null || !TemporaryBlacklist.Contains(file)));
        }

        private const string PathMods = "D:\\SteamLibrary\\steamapps\\common\\Celeste\\Mods";
        private string[] _files;
        private string tmp;


        public LINQVsFor()
        {
            _files = Directory.GetFiles(PathMods);
        }

        private void LoadMod(string file)
        {
            tmp = file;
        }

        [Benchmark]
        public void LINQ()
        {
            string[] files = _files
                    .OrderBy(f => f)
                    .Select(Path.GetFileName)
                    .Where(file => file.EndsWith(".zip") && ShouldLoadFile(file))
                    .ToArray();
            int len = files.Length;
            foreach (string file in files)
            {
                LoadMod(Path.Combine(PathMods, file));
            }
        }


        [Benchmark]
        public void For()
        {
            string[] files = _files;
            Array.Sort(files);
            int len = 0;
            for (int i = 0; i < files.Length; i++)
            {
                string file = Path.GetFileName(files[i]);
                if (!file.EndsWith(".zip") || !ShouldLoadFile(file))
                    continue;
                len++;
            }
            for (int i = 0; i < files.Length; i++)
            {
                string file = Path.GetFileName(files[i]);
                if (!file.EndsWith(".zip") || !ShouldLoadFile(file))
                    continue;
            }
        }


        [Benchmark]
        public void For2()
        {
            string[] files = _files;
            bool[] shouldLoad = new bool[files.Length];
            Array.Sort(files);
            int len = 0;
            for (int i = 0; i < files.Length; i++)
            {
                string file = Path.GetFileName(files[i]);
                shouldLoad[i] = file.EndsWith(".zip") && ShouldLoadFile(file);
                if (shouldLoad[i])
                    len++;
            }
            for (int i = 0; i < files.Length; i++)
            {
                if (shouldLoad[i])
                    LoadMod(files[i]);
            }
        }
    }

    public class Program
    {
        public static void Main(string[] args)
        {
            var summary = BenchmarkRunner.Run<LINQVsFor>();
        }
    }
}
```
</details>

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -176,28 +176,35 @@ public static class Loader {
...
-
-                files = Directory.GetDirectories(PathMods);
-                Array.Sort(files); //Prevent inode loading jank
-                for (int i = 0; i < files.Length; i++) {
-                    string file = Path.GetFileName(files[i]);
-                    if (file == "Cache" || !ShouldLoadFile(file))
-                        continue;
-                    LoadDir(Path.Combine(PathMods, file));
+                foreach (string dir in dirs) {
+                    LoadZip(Path.Combine(PathMods, dir));
```
Wartori54: Just noticed now, typo with catastrophic effects, this should be `LoadDir`.

solstice23:
I was so dumb... Fixed.
Also, is adding a progress bar to the taskbar icon a good idea? It could indicate that Celeste is still loading and the loading progress without clicking to open the splash, but it can also be distracting.

Wartori54:
[I just went ahead](https://github.com/Wartori54/Everest/tree/splash-progress-updates) and continued your idea (which i had also planned :P) and added dynamic font rendering to render the total mod count and as well as the current loading mod.
image
Sent this to the everest team on discord and they liked it so its okay.

Also it is not distracting in any way, that was the initial goal of the splash, give feedback to the user while the game window is not visible, but to get the splash out quicker I decided to split its implementation into two PRs (since the font rendering was not trivial).

Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
JaThePlayer: Note: You should be careful with Linq microbenchmarks on modern .NET, as Profile Guided Optimisation can make Linq seem much much better in benchmarks than in real code - the JIT will just emit a fastpath directly into linq methods which inline everything, which it can't do in real apps due to how many distinct arguments linq methods receive.

Either way, perf here is realistically not a major concern, there are much bigger issues in loading than this :p

PR #740, opened 2024-02-06
Add `TextMenuExt.FloatSlider`

SnipUndercover:
I think having an `IntSlider`, but for floats, for options like percentages or multipliers (something with like 2 digits of precision) would be useful.  
It behaves almost identical to `IntSlider`, but also allows the user to input a number directly by going to `OuiNumberInput`.

At the moment only the class exists and can be used via something like this:
```cs
public float MySetting { get; set; }

public void CreateMySettingEntry(TextMenu menu, bool inGame) {
  menu.Add(
    new TextMenuExt.FloatSlider(Dialog.Clean("..."), 0, 100, MySetting, 1)
      .Change(value => MySetting = value)
  );
}
```

I'm planning on adding support for automatically creating `FloatSlider`s if a property type is `float`, as well as customizing its parameters.

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
+++ b/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
@@ -1,5 +1,6 @@
 ﻿using Celeste.Mod;
 using Celeste.Mod.Core;
+using Celeste.Mod.UI;
 using Microsoft.Xna.Framework;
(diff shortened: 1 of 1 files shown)


SnipUndercover:
Closing this PR due to a lack of activity. Furthermore, I don't think there's a lot of need for a FloatSlider.

PR #747, opened 2024-02-20
Create relative instead of absolute symlinks

### LINKED ISSUE #742: (Linux) Use relative rather than absolute symlinks

Yes

Yes

### Describe the bug

Upon installing Everest in a clean Celeste folder, then moving the folder, the game no longer launches because of missing library files. This seems to be because the symlinked libraries use absolute rather than relative symlinks, so when you move the game folder those symlinks are broken. Changing these to relative symlinks resolves the issue:

```bash
cd ./lib64-linux
ln -sr --force libFAudio.so libFAudio.so.0
ln -sr --force libFNA3D.so libFNA3D.so.0
ln -sr --force libSDL2-2.0.so libSDL2-2.0.so.0
ln -sr --force libSDL2_image-2.0.so libSDL2_image-2.0.so.0
ln -sr --force libfmod.so libfmod.so.10
ln -sr --force libfmodstudio.so libfmodstudio.so.10
```

### Steps to reproduce

1. Install stable Everest
2. Move the game folder to a different location
3. Attempt to launch the game

### Expected behavior

The game should launch after the directory is moved.

### Operating System

Arch Linux

### Everest Version

4607

_No response_

### Additional context

_No response_

Kalobi:
MiniInstaller now sets up relative symlinks. This should hopefully allow moving game installs without breaking symlinks and fix #742 .
Still needs to be tested on Linux and macOS.

### DIFF
--- a/MiniInstaller/Program.cs
+++ b/MiniInstaller/Program.cs
@@ -209,7 +209,7 @@ namespace MiniInstaller {
                 Directory.CreateDirectory(Path.Combine(PathGame, "Mods"));
             }
 
-            // Can't check for platform as some morons^Wuninformed people could be running MiniInstaller via wine.
+            // Can't check for platform as some people could be running MiniInstaller via wine.
             if (PathGame.Replace(Path.DirectorySeparatorChar, '/').Trim('/').EndsWith(".app/Contents/Resources")) {
                 PathOSXExecDir = Path.Combine(Path.GetDirectoryName(PathGame), "MacOS");
                 if (!Directory.Exists(PathOSXExecDir))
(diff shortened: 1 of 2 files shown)


Kalobi:
Seems to work fine on Linux, just needs macOS testing now.

swoolcock:
Relative symlink is created correctly but running Everest fails on macOS because Piton doesn't understand how to handle relative symlinks. @Popax21 

image

maddie480:
After updating Piton with some help from @Popax21, I can confirm the symlinks are created correctly with relative paths, and that the game runs (Everest or vanilla) on Windows x64, macOS and Linux

PR #749, opened 2024-02-29
Use streams while relinking asm files for the most time

Saplonily:
This could be a fix for those empty pdb files as monomod wouldn't load pdb files if there's no matching pdb file beside the dll file, though currently dll files is actually in temp folders.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Relinker.cs
@@ -8,8 +8,12 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


maddie480:
what's the status on this again? :sweat_smile:

Saplonily:
> what's the status on this again? 😅

I've been on it for weeks and no issues encountered, but I'd close this since it conflicts with #754 and I have not much spare time these times for it.

Kalobi:
@Saplonily Have you been using code mod hot reloading while testing this? IIRC that was the main concern.

Saplonily:
> @Saplonily Have you been using code mod hot reloading while testing this? IIRC that was the main concern.

Yes I updated 3 of my mods on this with hot reloading working

PR #753, opened 2024-03-12
Remove unhelpful "Everest in your Everest" error log message

SnipUndercover:
While nowadays it's rare to see this error log message showing up, it can still do so, confusing the end user.
This generally ends up with a modding helper asking the person for `log.txt`, causing unnecessary delay.

This replaces said message with what everything else would do - a stacktrace.


maddie480:
Rest in peace ever rest

maddie480:
We do get a lot of #modding_help threads just starting with a log saying "whoops everest ever rest the game crashed" requiring to ask the user to go fetch log.txt (if they can find it, didn't run the game again since, etc), so yeah

PR #758, opened 2024-03-31
Show a postcard when a map has no rooms

SnipUndercover:
Instead of crashing, show a postcard when a map has no rooms, or when the starting room defined in map metadata does not exist.

Below are screenshots of this PR in action.
Postcard saying "Oops! This map has no rooms! Please make sure the map has at least one room, and that you've saved it since its creation."
Postcard saying "Oops! The starting room a-00 does not exist! Please make sure the room exists, and that you've saved it since its creation."


maddie480:
(I force pushed with no changes because I want Azure to give me a build to see wtf the MonoModRule gives once applied)

maddie480:
## Before

image

## After

image

looks good :+1:

PR #770, opened 2024-05-24
fix(search-box): TextBox not consuming buffer

mayromr:
Fixes 2 issues pointed out on the celeste discord:
1. https://discord.com/channels/403698615446536203/1243118722068123669/1243118722068123669
2. https://discord.com/channels/403698615446536203/429775439423209472/1241578985759051817

The loop that consumes all the button inputs is kind of jank, it would be way better to just consume the button that was pressed, but the downside is small and it seems to work fine so w/e.

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
+++ b/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
@@ -6,6 +6,7 @@ using Monocle;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using static Celeste.patch_TextMenu;
 
 namespace Celeste {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
```diff
@@ -1480,6 +1481,7 @@ bool initiallyVisible
 
         public class TextBox : TextMenu.Item {
             private static readonly float DEFAULT_TEXT_SCALE = 1.10f;
+            private static readonly FieldInfo mInputVirtualInputsField = typeof(MInput).GetField("VirtualInputs", BindingFlags.NonPublic | BindingFlags.Static);
```
maddie480: `VirtualInputs` is `internal`, and since the code becomes part of Celeste when installed, you don't need reflection to access it ^^ you can access `patch_MInput.VirtualInputs` instead, which should be transformed into the real MInput on install

mayromr: Done.

Issue #575, opened 2023-01-27
Use a Set collection for the mod blacklist/whitelist


coloursofnoise:
Currently these use lists, but I don't really see any reason why we'd want any duplicates in them

src: https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs#L27-L57

mayromr:
Hey, I was looking into this issue, It seems like a good idea to move to `HashSet`, But I have tried playing around with this change but it seems that `HashSet` doesn't implement `AsReadOnly`.
After looking online for a solution I found out that in newer [versions of .NET](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1?view=netframework-4.5.2#remarks) HashSet Implements `IReadOnlyCollection<T>` which will solve our issue, are there any plans (or is it even possible) to upgrade the .NET versions of Everest?

Kalobi:
@nhruo123 Yes, we are in fact working on porting Everest to .NET 7.0 (see the `core` branch in this repo). A public beta should roll out soon.

microlith57:
now that some time has passed, @Popax21 thoughts?

PR #771, opened 2024-05-27
Make blacklist into hashset

Apllify:
Just changed the current List<str> representation of the blacklist into a HashSet<str> as per #575 now that we're using .NET 7.0. 

Ran some basic testing and had no issues (game opens, in-game blacklist changes work fine), but I would appreciate a quick second opinion on this.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -8,6 +8,7 @@ using Microsoft.Xna.Framework;
 using Monocle;
 using MonoMod.Utils;
 using System;
+using System.Collections.Immutable;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
```diff
@@ -32,11 +32,11 @@ public static class Loader {
             /// The path to the Everest /Mods/blacklist.txt file.
             /// </summary>
             public static string PathBlacklist { get; internal set; }
-            internal static List<string> _Blacklist = new List<string>();
+            internal static HashSet<string> _Blacklist = new HashSet<string>();
             /// <summary>
             /// The currently loaded mod blacklist.
             /// </summary>
-            public static ReadOnlyCollection<string> Blacklist => _Blacklist?.AsReadOnly();
+            public static IReadOnlyCollection<string> Blacklist => (IReadOnlyCollection<string>) _Blacklist ;
```
maddie480: instead of a simple cast, I would use `ToImmutableHashSet` to make it actually read-only :thinking: 

also the type change might be a breaking change, even though I don't think any mods would use `Everest.Loader.Blacklist` (unless there's a cursed mod out there that does stuff based on the blacklist :maddyS:)

PR #772, opened 2024-06-14
Fix incorrect Celeste.xml names

Brokemia:
Several of the names used in the Celeste.xml file are incorrect, this fixes them


Brokemia:
It's worth noting that `T:Monocle.VirtualAxis` and `M:Celeste.TextMenu.SubHeader.#ctor(System.String)` are in Celeste.xml but could just be documented in Everest instead, as they're not present in vanilla anymore and only exist in Everest for backwards compatibility

PR #780, opened 2024-06-19
Fixes EntityIDs to be consistent for globally generated entities

vivianlonging:
Split segment of #689.

Fixes EntityIDs to be consistent for globally generated entities. 
Currently, EntityIDs are constructed from the room Index "ID" (in *theory* this should be global indexed by Ahorn/Loenn, but isn't guaranteed, it is expected by vanilla to be indexed by room, so this is an assumption we keep for this situation) as well as the level that the player is *currently* in, regardless of the room that the index was constructed for.
This means that, for entities that load into the game on map startup ("global controllers") their IDs will be [whatever room the player started/restarted the Session from]:[the internal room indexed ID] which means that the EntityID is variable, and therefore can randomly disable entities from other rooms (with the global index system, this will only theoretically DoNotLoad a single entity, but this is still not intended behavior)

### DIFF
--- a/Celeste.Mod.mm/Patches/EntityData.cs
+++ b/Celeste.Mod.mm/Patches/EntityData.cs
@@ -14,8 +14,8 @@ namespace Celeste {
 
         public EntityID EntityID;
 
-        internal void SetEntityID(bool? isTrigger = null) {
-            EntityID = new EntityID(Level.Name ?? string.Empty, ID + ((isTrigger ?? false) || patch_LevelData._isRegisteringTriggers ? 10000000 : 0));
+        internal void InitializeEntityID() {
+            EntityID = new EntityID(Level.Name ?? string.Empty, ID + (patch_LevelData._isRegisteringTriggers ? 10000000 : 0));
         }
 
     }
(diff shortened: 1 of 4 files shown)


Celeste.Mod.mm/Patches/EntityData.cs
```diff
@@ -12,5 +12,11 @@ class patch_EntityData : EntityData {
             return orig_Has(key);
         }
 
+        public EntityID EntityID;
+
+        internal void SetEntityID(bool? isTrigger = null) {
```
maddie480: I find the name weird since it is not a setter for EntityID 🤔 maybe `ComputeEntityID` or `InitializeEntityID`?

maddie480: Is the parameter ever used?

vivianlonging: the parameter is not used... yet. it's added in for a future feature, assuming we infact get that far with this PR lmao
I'll change the name to InitializeEntityID

PR #782, opened 2024-06-20
Fix code hot reload accidentally loading a cached version

strandsofivy:
cache invalidation strikes again

The hash is never recomputed when an assembly changes, which makes Everest load a cached version even though the assembly has changed.

This change is yet untested


maddie480:
^ needs testing, more precisely

PR #786, opened 2024-06-29
EntityID fix: force resolve Level nullref on map reload

vivianlonging:
This forces the EntityID initializer to handle Level nullreferences with the string `none` when map reload occurs. This should never reasonably happen, but an error has come up somewhere. 

Since I can't actually *test* this error, I've decided to ensure there are 0 chances of having a NullRef in the first place.


maddie480:
Merge conflicts need to be resolved

Kalobi:
Presumably the merge conflicts are because the previous PR was reverted.

PR #788, opened 2024-07-02
Fixing EntityID duplication on Global objects

vivianlonging:
This is the same as the fixEntityID patch but i put it in a single commit

EntityID before was set by (level the player is in):(entity internal ID) which is not consistent for entities that spawn from EntityData globally, since (level the player is in) changed based on Save & Quit/Return To Map/Enter From Checkpoint

This sets EntityID to be relative to the LevelData that was used to house the EntityData, and also changes the patches in Level to use that EntityID without breaking mods hooking Level (which is why i didn't remove instructions)


maddie480:
Re-ran the Azure pipeline because NuGet had a moment and the nuget restore timed out after _1 hour_

maddie480:
revival of #780

PR #801, opened 2024-08-03
Make the Atlas warning output full path under mod

AAA1459:


maddie480:
I'd just like to see a screenshot before merging

maddie480:
The point is making it clearer which atlas the texture is from

PR #803, opened 2024-08-06
Fix TrailManager confusion when sprite are large or player hair is too long

AAA1459:
fix: 
7KMB{$SB6OGDW9OT($ Y88S

### DIFF
--- a/Celeste.Mod.mm/Patches/TrailManager.cs
+++ b/Celeste.Mod.mm/Patches/TrailManager.cs
@@ -1,17 +1,24 @@
-﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
-#pragma warning disable CS0649 // Field is never assigned to, and will always have its default value
-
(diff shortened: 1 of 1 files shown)


maddie480:
looks like there's a lot of `MonoModReplace` 🤔 how much of it is vanilla copypaste?

AAA1459:
Almost all, except `buffers` related

AAA1459:
Don't merge this now, wait me a sec

maddie480:
I'm wondering if you could make an IL patch instead, to avoid replacing code as much as possible 🤔

AAA1459:
> I'm wondering if you could make an IL patch instead, to avoid replacing code as much as possible 🤔

I am not familiar with IL only know how to emit a delegate. but i probably could, except `BeforeRender` that almost is rewritten

AAA1459:
@maddie480 oh thanks you, that gave me a very good reference.

PR #807, opened 2024-08-23
Make DeathEffect support transparent colors and custom texture

AAA1459:
Note: This is incompatible with [Fewer Visual Distractions](https://gamebanana.com/mods/491205) that its ilcode is not very safe

### DIFF
--- a/Celeste.Mod.mm/Patches/DeathEffect.cs
+++ b/Celeste.Mod.mm/Patches/DeathEffect.cs
@@ -1,15 +1,29 @@
 using System;
+using System.Xml;
 using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
+using Monocle;
 using MonoMod;
 using MonoMod.Cil;
+using MonoMod.InlineRT;
+using MonoMod.Utils;
 
 namespace Celeste {
     class patch_DeathEffect : DeathEffect {
 
+        // no-op. MonoMod ignores this - we only need this to make the compiler shut up.
         public patch_DeathEffect(Color color, Vector2 offset)
             : base(color, offset) { }
+
+
+        public MTexture Texture;
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/DeathEffect.cs
```diff
@@ -1,24 +1,50 @@
...
         public patch_DeathEffect(Color color, Vector2 offset)
             : base(color, offset) { }
 
+
+        public MTexture Texture;
+
+        /// <summary>
+        /// a temporary texture used to the next DeathEffect.Draw get call
+        /// </summary>
+        public static MTexture _texture;
```
JaThePlayer: This shouldn't be public

AAA1459: I had public it for do some special testing by other mods, I must have forgotten to change it back...

AAA1459:
new commit should fixed this PR conflict with FewerVisualDistractions

DashingCat (commented):
I tested the changes with the following command:
```c#
[Command("deatheffect", "deatheffect")]
public static void DeathEffect(string r, string g, string b, string a) {
    Player player = Engine.Scene.Tracker.GetEntity<Player>();
    player.Depth = -1000000000;
    player.Add(new DeathEffect(new Microsoft.Xna.Framework.Color(
        Convert.ToByte(r),
        Convert.ToByte(g),
        Convert.ToByte(b),
        Convert.ToByte(a)
    )));
}
```

The patch seems to give unexpected results with no alpha: the outline is not rendered while the filling color is; and with high alpha, the transparency is a bit hard to notice, as shown below:

https://github.com/user-attachments/assets/4ced7325-01dc-4535-99bf-376ecf4ce738

I don't consider this a blocker, because this feature is to be used explicitly by mods, and the rendering doesn't change with opaque colors, and I will approve the PR if this behavior is expected.

DashingCat (approved):
Turning the comment into an approval since the behavior shown is the one expected.

PR #811, opened 2024-08-30
Add option to show true LastArea in FileSelect

ErynGalen:
This adds an Everest option to show the SID of unavailable maps in the FileSelect menu, instead of showing the vanilla progression.


ErynGalen:
Example:

maddie480:
The "discussion" the 6-month-old label refers to is "should this really be an option? or should it just be... implemented without a switch?" I don't remember if the discussion really happened

maddie480:
:despaireline: the discussion never happened

microlith57:
i feel like there is no reason we wouldn't want this behaviour, it's kinda just obviously correct. is there a reason anyone would want to have this option turned off?

DashingCat (changes requested):
I agree with @microlith57, I believe this does not need to be an option.

However, the current implementation conflicts with the mod `CollabUtils2`.

The IL manipulation in this PR removes the reference to the field `Celeste.AreaData.Name`, which is used here in `modSelectSlotLevelSetDisplayName`: https://github.com/EverestAPI/CelesteCollabUtils2/blob/cc0609b38505bd5a0442c3933367bd3400634576/LobbyHelper.cs#L679-L698

With the changes in the PR applied, the collab name is no longer used, and instead the last area of the collab is shown (for example, for the `Strawberry Jam` collab, the name shown on the file select slot is `Grandmaster Lobby` instead of `Strawberry Jam Collab`).

SnipUndercover:
> With the changes in the PR applied, the collab name is no longer used, and instead the last area of the collab is shown (for example, for the Strawberry Jam collab, the name shown on the file select slot is Grandmaster Lobby instead of Strawberry Jam Collab).

@DashingCat Should we be worried about this? The collab will be unloaded anyway, making its collab name not present in the `HashSet<string> collabNames`.

Celeste.Mod.mm/Patches/OuiFileSelectSlot.cs
```diff
@@ -313,6 +300,34 @@ public override void Render() {
             }
         }
 
+        public string GetFurthestAreaName() {
+            if (!CoreModule.Settings.TrueLastAreaInFileSelect || string.IsNullOrEmpty(SaveData.TrueLastAreaSID)) {
+                return Dialog.Clean(AreaData.Areas[FurthestArea].Name);
+            }
+
+            float maxSize = 580;
```
SnipUndercover, line 308: Where does the value `580` come from?

Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
```diff
@@ -283,6 +282,10 @@ public bool ColorizedLogging {
             }
         }
 
+        [SettingSubText("MODOPTIONS_COREMODULE_TRUELASTAREAINFILESELECT_DESC")]
+        [SettingInGame(false)]
+        public bool TrueLastAreaInFileSelect { get; set; } = false;
```
SnipUndercover, line 287: As seen in the conversation in the comments, the new behavior seems more correct and there doesn't seem to be a reason to keep the old one around. The setting therefore is unnecessary.

SnipUndercover (changes requested):

Issue #439, opened 2022-04-19
Show SID during file select if the mod is disabled


Appels0:
### I am planning on implementing this myself

No

### Describe your request

On the file selection screen, show the SID of the map that the file will try to load if said mod is not enabled.

### Additional context

Over time I've ended up with quite a lot of files where I've saved and quit in a room with the intention of returning to them later, but have for one reason or another had to disable the mod before then. Now it's hard to keep track of which files are important and shouldn't be opened to avoid getting kicked out of the level because they all read "Forsaken  City" during file selection. This problem could be fixed if the files instead showed the SID of the map that the file will try to load, so you can at least have some idea what the file was for even if the mod in question is currently disabled.

PR #818, opened 2024-09-07
File session details on the file select screen

LuneAres:
## Overview
The goal was to be able to know if there was an ongoing session. You can see some informations about the ongoing session associated with the files before continuing them from the file select screen: 
- is there an ongoing session
- is the map available
- map name (SID if not available)
- map "chapter" (the text displayed above the area name on the banner)
    -> is it customizable in Everest? If it is, I need to make the session details compatible
- map title banner ("title" and "accent")
- map mode (nothing for A sides, an indicator for B or C side)
- session time
- deaths counter
- has the map already been completed on this save file
- berry counter
- berry list
    -> like the one at the bottom of the when in pause
    -> only if the map has already been completed
    -> scales to not take too much space

It addresses issue #439.

## Implementation
- adds a `SessionDetailsPage` entity that stores the relevant informations of a session to render them, and can be updated to switch sessions (the berry list is baked into a virtual render target)
- adds ref to a `SessionDetailsPage` instance in `OuiFileSelect`
- modifies `OuiFileSelect.Update()` to modify input capture
- modifies the IL patch on `OuiFileSelectSlot.Render` to change ticket rendering position based on a the `OuiFileSelectSlot.ticketRenderPosition` property (added by this change too) to make the ticket move depending on `OuiFileSelect.SessionDetailsEase`
- modifies the `OuiFileSelectSlot.IdlePosition` to shift the slots idle position to the left when showing session details
- modifies `Overworld.InputEntity` and  `Overworld.Update()` to show the input prompt for toggling session details in the bottom right of the screen with "Confirm" and "Cancel"

### AltSideHelper
I don't know what needs to be exposed and how.
For now I did put into (private) fields of `SessionDetailsPage` the things that I think are relevant for AltSide compat

### CollabUtils2
Maybe being able to see the icon and the name of the lobby the map is associated with would be great.
-> for example, Strawberry jam maps will display their sub-difficulty indicator as their icon

### Save slots manipulation
There is the idea of being adding the possibility to move save slots. I also thought about being able to erase a slot (so every slot below it get moved up by one).
We need to think of a way to cleanly allow those save slots operations without conflicts. For now there is a mod for moving slots (https://gamebanana.com/mods/539280), but it also uses the journal key (I actually didn't test the interaction between this mod and my patches)

## Issues / things to add
Things I just noticed I need to fix, or I think would be (more or less) relevant to consider doing/adding in continuation to the initial commit, at some point in time or never, by someone else or me, in Everest or in a mod.... I don't know, and I'm asking you about what you think (maybe that's not the right place to do that for some of them):

- allow for configuration of the keybind (currently hardcoded as the Journal key)
- rename `SessionDetailsPage` to `OuiFileSelectSessionDetailsPage` and move it to a seperate file
- add a public `bool ShowSessionDetailsUI` field to `Overworld` and update it from `OuiFileSelect` instead of harcoding the check in `Overworld.Update()`
- translation of the few new dialogs (currently there's only english and french)
- backpack icon on the card to tell if there is an ongoing session at first glance without looking at the details
- show collectables collected during the session (cassette / hearthgem)
- put the journal time icon next to the session time
- add a warning that if the map isn't loaded, then the session details may be different from what is espected
    -> apart from the name, chapter, and area icon, there is also: custom banner (title and accent), tab, mode icon (for B and C sides)
- show the mode icon for the A-side if it's a custom one?
- add an indicator of "position in the map" using the berry list
- a way to cache session details infos so we can show them even if the map isn't available (still indicate somewhere if the map is available or not)
- a few things about campaigns
    - being able to see all the campaigns you played for each savefile
    - being able to store up to one ongoing session per campaign for each file slot
    -> maybe add something LevelSetPicker-like between the ticket and session details page to move between played campaigns (use the savedata total playtime of the areas in the levelset to determine?) that can navigate between played campaigns when viewing session details
        - changing the rendered card, ticket and session details to match the selected campaign
        - being able to access an easier way to select the campaign, a bit like the map search that you can access from chapter select screen (maybe also add that to the LevelSetPicker)
            -> by pressing confirm?


maddie480:
well... it's still a draft 2 months later :thinking:

LuneAres:
I think I will remove the pull request.
Maybe I'll come back to this, but if I do I would probably rewrite it as a mod in first intention (maybe with compatibility with Aurora's per-map stored sessions, but at least with the ability to have one stored session per campaign)
And maybe it'll become a Everest pull request again one day (I'll try to write a better pull request then, both in terms of code and reviewability)
I also noticed a few minor graphical bugs to adress

PR #822, opened 2024-09-16
add positionMode attr to music fade trigger

microlith57:
when present, this overrides the vanilla direction param, which is less flexible


microlith57:
(cc'ing @maddie480 since helping hand has a related trigger)

PR #824, opened 2024-09-29
Sort Mod Options Alphabetically

notTamion:
I often find myself looking for the options of a specific mod i added but not remembering the full name of the mod. It would make finding these settings a lot easier if the options were sorted alphabetically as i often do remember the first letter or two. I would also be fine with adding an extra option to toggle between load order and alphabetical sort incase this is too big of a change which might confuse long time users that are used to their current order. (myers seems to add some random diff for some reason even though i didn't change anything in those lines)

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/OuiModOptions.cs
+++ b/Celeste.Mod.mm/Mod/UI/OuiModOptions.cs
@@ -117,8 +117,13 @@ namespace Celeste.Mod.UI {
                 }
             }
 
-            // reorder Mod Options according to the modoptionsorder.txt file.
             List<EverestModule> modules = new List<EverestModule>(Everest._Modules);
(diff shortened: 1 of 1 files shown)


RedFlames:
I guess because of `Everest.Boot()`, the CoreModule is currently always the first one in the list, but shouldn't you yank out `modules.FindIndex(module => module.Metadata.Name == "Everest")` rather than index `0` just in case? 😃 

Or possibly `modules.Remove(CoreModule.Instance)` and later `modules.Insert(0, CoreModule.Instance)` if that would work.

Also yeah I just randomly looked at this. Besides that as an average Everest user, as long as modoptionsorder.txt still works I'm not sure it'd make much of a difference to me if they're a jumble in load order or alphabetical x)

notTamion:
thx for the feedback.

PR #825, opened 2024-09-30
Do not hide the HUD when open the keys config menu in level and holds Journal key to clear config

AAA1459:


maddie480:
let's have a build to check what this does (if that spell still works :magic_wand:)
/azp run

microlith57 (approved):
lgtm :+1: 
@AAA1459 shall i merge this?

AAA1459:
@microlith57  of course i think

PR #832, opened 2024-10-20
Add ILCursor TryGotoNext/TryGotoPrev extensions

SnipUndercover:
Adds useful `GotoNextBestFit`/`GotoPrevBestFit` extensions for mods to simplify matching IL sequences which may have already been IL hooked by other mods.
These methods allow matching a set of instructions with a certain level of tolerance, `0x10` by default. Then, they pick the match with least instruction spread, which is closest to the cursor.

> [!NOTE]
> When using (`Try`)`GotoPrevBestFit` with `MoveType.After`, the index may have been moved forward instead of backward, if the predicates list is long enough and the first predicate matches close enough to the cursor.
> 
> For example:
> ```
> ldarg.0
> ldfld class Monocle.StateMachine Celeste.Player::StateMachine
> // cursor is here
> ldc.i4.0
> callvirt void Monocle.StateMachine::set_State(int32)
> ```
>
> When calling (`Try`)`GotoPrevBestFit` with `MoveType.After` and `predicates` matching the entire IL snippet, the cursor will be placed after the `callvirt`.

##### TODO list:
- [x] Add verbose logging to help debug when a match suddenly does not... match

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/ILCursorExtensions.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/ILCursorExtensions.cs
@@ -159,7 +159,7 @@ namespace Celeste.Mod.Helpers {
                     if (!cursor.TryGotoNext(MoveType.After, matcher))
                     {
                         Logger.Verbose(NextBestFitLogID,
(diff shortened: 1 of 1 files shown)


microlith57 (approved):
approved pending the verbose logging

microlith57 (approved):

PR #836, opened 2024-10-27
Add Everest Photosensitivity menu

GamationOnGithub:
Adds a submenu to the vanilla Options menu that allows for toggling the 5 parts of Photosensitive Mode (anxiety, glitch effects, screen flashes, lightning flashes, and text highlighting) on and off. Also creates publicly accessibly options for modders to tie their own effects to, e.g. `AllowFlash`, `AllowTextHighlighting`, etc.

PR #837, opened 2024-10-28
test pipeline

maddie480:


maddie480:
so the pipeline not triggering on PRs isn't limited to #836 huh :thinking:

maddie480:
https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers lied to me :landeline:

PR #844, opened 2024-11-02
Make `ModAsset.Deserialize` throw exceptions on failure

SnipUndercover:
`ModAsset.Deserialize<T>()` uses `ModAsset.TryDeserialize<T>(out T)`, which suppresses all exceptions when deserializing YAML.
This can hinder debugging if deserialization is failing for whatever reason, as the exception info is now lost.

This PR causes `ModAsset.Deserialize<T>()` to now throw exceptions, if deserialization fails.

###### *(Mods depending on `ModAsset.Deserialize<T>()` could break because of that, which is why this is a draft.)*


maddie480:
I guess we need to search for mods using ModAsset.Deserialize to pull this PR out of review hell?

maddie480:
I decompiled all mods and none of them seem to use `ModAsset.Deserialize`, so uh sounds good to me

maddie480:
nevermind, I grepped `.Deserialize` instead and found quite a few ones

maddie480:
- SkinModHelper `return skinConfigYaml.Deserialize<SkinModHelperConfig>();`
- Styline `return Everest.Content.Get("Content/" + data.Attr(attrName)).Deserialize<T>();`
- Eevee_Skinmod `paletteList.Add(child.PathVirtual.Substring(child.PathVirtual.LastIndexOf('/') + 1), child.Deserialize<Palette>());`
- AchievementHelper `List<Achievement> list = asset.Deserialize<List<Achievement>>();`
- SkinModHelperPlus `return asset.Deserialize<T>();`

SnipUndercover:
- SkinModHelper: The helper does not check for `null`s and will NRE if deserialization fails.
- Styline: Here it uses `Deserialize` twice, but I'm not sure whether it will NRE, probably best to poke @Popax21 about it?
- Eevee: The skinmod does not check for `null`s and will NRE if deserialization fails.
- AchievementHelper: The helper does not check for `null`s and will NRE if deserialization fails.
- SkinModHelperPlus: A move to `TryDeserialize` would help here.

An explicit exception seems like an overall benefit for SMH, the Eevee skinmod and AchievementHelper, so I'd leave them be.

SnipUndercover:
Should this PR be revived? Do we have any other cases of ModAsset.Deserialize? @maddie480

Issue #766, opened 2024-05-06
Saving to non-mod files in Mods/ messes up the main menu


underscoreevelyn:
### My mods and Everest install are up to date

Yes

### I have recreated the bug with only Everest OR a minimum number of mods enabled

Yes

### Describe the bug

Writing to non-mod files in Mods/, notably blacklist.txt and modoptionsorder.txt, duplicates the cursor on the main menu. This generally behaves poorly and can crash the game.

### Steps to reproduce

1. Navigate to the main menu. Other screens like the title screen and the mod options menu also work.
2. Write to an existing file in Mods/. If saving the file would create that file, it doesn't work for some reason. You can do it multiple times to get even more cursors.
3. Pressing confirm selects all the selected options at once, which can look weird or crash the game.

https://youtu.be/ZPZTfoG9Tls

### Expected behavior

Don't do that.

### Operating System

Linux

### Everest Version

4771

### Additional context

[log_20240506_164755.txt](https://github.com/EverestAPI/Everest/files/15226917/log_20240506_164755.txt)

underscoreevelyn:
This behavior is caused by creating two filesystem events in the same frame, which CAN happen when writing to files but doesn't ALWAYS happen. An easier way to trigger it is to just create multiple files at once, i.e. `touch Mods/{a,b}.txt`.

Handling two auto loads in a single frame on the main menu eventually calls `OuiMainMenu.CreateButtons` twice, which bugs it out and causes it to add a bunch of orphaned `MenuButtons` to the scene.

Please also enjoy this funny, slightly cursed one liner that fixes the problem from a mod:
```csharp
On.Celeste.OuiMainMenu.CreateButtons += static (orig, self) => { orig(self); self.Scene.OnEndOfFrame += () => { foreach (var button in self.Scene.Entities.FindAll<MenuButton>()) if (!self.Buttons.Contains(button)) button.RemoveSelf(); }; };
```

PR #845, opened 2024-11-03
Prevent leftovers MenuButtons when rebuilding menu

DashingCat:
Defer `RebuildMainAndTitle()` right before `Update()`, so `CreateButtons()` is only called once per frame.

When `CreateButtons()` is called more than once per frame, it creates orphaned buttons (not linked to the `OuiMainMenu`) which are not cleaned up properly, causing selection issues in the main menu.

The method `RebuildMainAndTitle()` is now `private` to prevent direct usage.

Closes issues #430 (and #766 which is a duplicate)


maddie480:
needs resolving conflicts (and I apparently can't do it myself)

DashingCat:
Thanks for the heads up, conflicts resolved.

PR #846, opened 2024-11-05
Add cassette fly check to Player.OnSquish

bigkahuna443:
The cassette fly state is intended to make the player invincible, but at the moment squish checks are still made against the player. This can cause deaths in modded situations when the return path crosses e.g. block entities.

This patch adds a check at the start of OnSquish to return immediately if the player is in the CassetteFly state.

Here is an example of the behavior [before the change](https://github.com/user-attachments/assets/f28b2689-3826-4885-9ed0-991d3a53587f) and [after the change](https://github.com/user-attachments/assets/7b350c28-b7bc-44f2-b18d-10338ddeadd4).

(This change was made manually in SJ and I thought it'd be helpful to add into Everest as a whole)

### DIFF
--- a/Celeste.Mod.mm/Patches/Player.cs
+++ b/Celeste.Mod.mm/Patches/Player.cs
@@ -329,6 +329,10 @@ namespace Celeste {
         [MonoModIgnore]
         [PatchPlayerApproachMaxMove]
         private extern int NormalUpdate();
+
+        [MonoModIgnore]
+        [PatchPlayerOnSquish]
+        protected extern override void OnSquish(CollisionData data);
     }
 
     public static class PlayerExt {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/Player.cs
```diff
@@ -601,5 +611,27 @@ public static void PatchPlayerApproachMaxMove(ILContext context, CustomAttribute
...
+            /*
+            Add a cassette fly check at the start of the method:
+            + if (this.StateMachine.State == StCassetteFly)
+            +    return;
+            */
+
+            cursor.EmitLdarg(0);
+            cursor.EmitLdfld(f_Player_StateMachine);
+            cursor.EmitCallvirt(m_StateMachine_get_State);
+            cursor.EmitLdcI4(21);
```
microlith57, line 631: minor nitpick:
```suggestion
            cursor.EmitLdcI4(21 /* StCassetteFly */);
```

microlith57 (approved):
looks good to me; unlikely to collide with the sj ilhook (though i assume you've tested that)

maddie480:
this is a 2-month-old draft, marking as dormant (ignore the pipeline failure)

PR #852, opened 2024-11-23
add session sliders

microlith57:
analogy: `Flags` : `bool` :: `Counters` : `int` :: `Sliders` : `float`

api developed in conversation with @JaThePlayer as this affects Session Expressions.

- `Slider`s are a class so that they can be cached
- there is an event for setting sliders, which also provides the previous value if it exists
  - keep in mind that getting a slider may create it, which is the only case when the previous value is `null`
  - also keep in mind that setting a slider by name when it previously did not exist will fire the changed event twice
- i spent so long on the serialization it was so painful. did you know `[Obselete]` has an undocumented interaction where it disables serialisation? did you know `SlidersSerializationOnly` has to be an array because it does some wacky stuff with lists that breaks things?

future work:
- make counters work this way too?
- some sort of triggers to interact with sliders/counters in some basic way?

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
@@ -182,9 +182,9 @@ namespace Celeste.Mod {
             }
 
             public static class Session {
-                public static event Action<patch_Session.Slider, float?> OnSliderChanged;
-                internal static void SliderChanged(patch_Session.Slider slider, float? previous)
-                    => OnSliderChanged?.Invoke(slider, previous);
+                public static event Action<patch_Session, patch_Session.Slider, float?> OnSliderChanged;
+                internal static void SliderChanged(patch_Session session, patch_Session.Slider slider, float? previous)
+                    => OnSliderChanged?.Invoke(session, slider, previous);
             }
 
             public static class Player {
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using _Decal = Celeste.Decal;
 using _EventTrigger = Celeste.EventTrigger;
 using _Level = Celeste.Level;
+using _Session = Celeste.Session;
 using _OuiJournal = Celeste.OuiJournal;
 using _OuiMainMenu = Celeste.OuiMainMenu;
 using _Player = Celeste.Player;
@@ -164,20 +165,26 @@ namespace Celeste.Mod {
                 internal static void LoadLevel(_Level level, _Player.IntroTypes playerIntro, bool isFromLoader)
                     => OnLoadLevel?.Invoke(level, playerIntro, isFromLoader);
 
-                public delegate void EnterHandler(Session session, bool fromSaveData);
+                public delegate void EnterHandler(_Session session, bool fromSaveData);
                 public static event EnterHandler OnEnter;
-                internal static void Enter(Session session, bool fromSaveData)
+                internal static void Enter(_Session session, bool fromSaveData)
                     => OnEnter?.Invoke(session, fromSaveData);
 
-                public delegate void ExitHandler(_Level level, LevelExit exit, LevelExit.Mode mode, Session session, HiresSnow snow);
+                public delegate void ExitHandler(_Level level, LevelExit exit, LevelExit.Mode mode, _Session session, HiresSnow snow);
                 public static event ExitHandler OnExit;
-                internal static void Exit(_Level level, LevelExit exit, LevelExit.Mode mode, Session session, HiresSnow snow)
+                internal static void Exit(_Level level, LevelExit exit, LevelExit.Mode mode, _Session session, HiresSnow snow)
                     => OnExit?.Invoke(level, exit, mode, session, snow);
 
                 public delegate void CompleteHandler(_Level level);
                 public static event CompleteHandler OnComplete;
                 internal static void Complete(_Level level)
                     => OnComplete?.Invoke(level);
+            }
+
+            public static class Session {
+                public static event Action<patch_Session.Slider, float?> OnSliderChanged;
+                internal static void SliderChanged(patch_Session.Slider slider, float? previous)
+                    => OnSliderChanged?.Invoke(slider, previous);
             }
 
             public static class Player {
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -1,17 +1,91 @@
...
+                [XmlAttribute]
+                public float Value;
+            }
+        }
+
+        [XmlIgnore]
+        private Dictionary<string, Slider> _Sliders;
+
+        [XmlIgnore]
+        public IReadOnlyDictionary<string, Slider> Sliders;
```
JaThePlayer: This should definitely be a `private set;` property, though typically stuff like this is done via `Sliders => _Sliders;`, because this is just a cast it shouldn't incur a performance penalty

Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -1,17 +1,91 @@
...
+
+        /// <summary>
+        /// Used internally for serialization; getting or setting this will cause issues.
+        /// Instead, use <see cref="Sliders" />.
+        /// </summary>
+        ///
+        /// NOTE: this cannot be private or [Obselete] because either of those would break serialization.
+        [XmlArray("Sliders")]
+        public SerializationOnly.Slider[] SlidersSerializationOnly {
```
JaThePlayer, line 67: `[EditorBrowsable(EditorBrowsableState.Never)]`? (Also on the SerializationOnly class)

Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -25,6 +99,40 @@ public void ctor(AreaKey area, string checkpoint = null, AreaStats oldStats = nu
...
+            => SetSlider(GetSliderObject(slider), value);
+
+        public void AddToSlider(Slider slider, float amount)
+            => SetSlider(slider, slider._Value + amount);
+
+        public void AddToSlider(string slider, float amount)
+            => AddToSlider(GetSliderObject(slider), amount);
+
+        public IEnumerator<Slider> EnumerateSliders()
+            => _Sliders.Values.GetEnumerator();
```
JaThePlayer: Do we need this, considering `Sliders` is exposed publicly already?

Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
```diff
@@ -180,6 +181,12 @@ internal static void Complete(_Level level)
                     => OnComplete?.Invoke(level);
             }
 
+            public static class Session {
+                public static event Action<patch_Session.Slider, float?> OnSliderChanged;
+                internal static void SliderChanged(patch_Session.Slider slider, float? previous)
+                    => OnSliderChanged?.Invoke(slider, previous);
```
JaThePlayer: I'd like this to give access to the `Session` object as well, either as an additional parameter or through `Slider`

Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -1,17 +1,91 @@
...
+using System.Collections.Generic;
+using System.Xml.Serialization;
 
 namespace Celeste {
     public class patch_Session {
+
+        public class Slider {
+            private patch_Session _Session;
+
+            public readonly string Name;
```
JaThePlayer: `public string Name { get; required init; }`? I think it's generally advised to expose properties over readonly fields

Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -1,17 +1,91 @@
...
 
+        public extern void orig_ctor();
+
+        [MonoModConstructor]
+        public void ctor() {
+            orig_ctor();
+
+            Sliders = _Sliders = new Dictionary<string, Slider>();
+        }
+
```
JaThePlayer: Hmm, there are some existing hooks for `Session..ctor`, I'm not sure whether this could break them due to Hookgen changing 🤔
```
On: System.Void Celeste.Mod.GhostModForTas.Recorder.GhostRecorder::OnSessionCtor(On.Celeste.Session+orig_ctor,Celeste.Session)
On: System.Void Celeste.Mod.CherryHelper.CherryHelper::handleSavedFlags(On.Celeste.Session+orig_ctor,Celeste.Session)
```
(https://jatheplayer.github.io/celeste/ilhookview/index.html?method=Celeste.Session..ctor%28%29)

Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -25,6 +99,40 @@ public void ctor(AreaKey area, string checkpoint = null, AreaStats oldStats = nu
...
+                _Sliders[slider] = obj = new(this, slider, 0f);
+                Everest.Events.Session.SliderChanged(obj, null);
+            }
+            return obj;
+        }
+
+        public void SetSlider(Slider slider, float value) {
+            float previous = slider._Value;
+            slider._Value = value;
+            Everest.Events.Session.SliderChanged(slider, previous);
```
JaThePlayer: This is a bit weird, maybe I'd move the event firing code to the Slider.Value setter instead, then this method doesn't even have to exist (If you have a Slider already, then you have the Value setter publicly available)

Celeste.Mod.mm/Patches/Session.cs
```diff
@@ -25,6 +99,40 @@ public void ctor(AreaKey area, string checkpoint = null, AreaStats oldStats = nu
...
+        }
+
+        public void SetSlider(string slider, float value)
+            => SetSlider(GetSliderObject(slider), value);
+
+        public void AddToSlider(Slider slider, float amount)
+            => SetSlider(slider, slider._Value + amount);
```
JaThePlayer: This overload seems a bit redundant, considering you can `slider.Value += amount;`

JaThePlayer (changes requested):

Celeste.Mod.mm/Patches/Session.cs
microlith57: oh true, forgot about that

Celeste.Mod.mm/Patches/Session.cs
microlith57: probably not tbh, will remove

Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
microlith57: leaning towards putting it in the event, but i guess i could do it through the slider instance. any preference?

Celeste.Mod.mm/Patches/Session.cs
microlith57: oh true, i guess i could monomodreplace it then

Celeste.Mod.mm/Patches/Session.cs
microlith57: true

JaThePlayer (approved):
lgtm

PR #858, opened 2024-11-30
Add Entity Collider Components to Entities

TheDavSmasher:
While making this to fulfill an idea for a single part of my mod, I instead generalized the idea and made it in such a way that having it in Everest would be useful for anyone.

The 2 components added, EntityCollider and EntityColliderByComponent, work in the same way a PlayerCollider or PufferCollider work in vanilla, except they are made so they can instead collide with any entity as defined with the generic type parameter. Entity Collider collides based on Entity, EntityColliderByComponent collides based on if an entity has the component of type.

Additionally, two methods were added to patch_Tracker: AddEntityToTracker and AddColliderToTracker. This allows for any type that is not normally tracked to become tracked and have its entities added to the Tracker. This is done in conjunction to the Entity Colliders due to performance, so it's not required to look through the Scene's Entities every single Update call. Some entities, like Springs, aren't tracked, but are a good candidate for something you'd want to collide with. This is why these methods were made. If anyone thinks this can be unsafe, anyone can find any entity anyways by going through Scene.Entities, so it does not add any vulnerability to normally untracked entities.

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -11,7 +11,7 @@ namespace Celeste.Mod.Entities {
 
         public Collider Collider;
 
-        public EntityCollider(Action<T> onEntityAction, Collider collider)
+        public EntityCollider(Action<T> onEntityAction, Collider collider = null)
             : base(active: true, visible: true) {
             OnEntityAction = onEntityAction;
             Collider = collider;
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -11,9 +11,9 @@ namespace Celeste.Mod.Entities {
     [Tracked(false)]
     public class EntityCollider<T> : Component where T : Entity {
         /// <summary>
-        /// Provides a simple way to know the Entity type of the specific Collider without Reflection
+        /// Provides a simple way to know the Entity type of the specific Collider
         /// </summary>
-        public readonly Type entityType = typeof(T);
+        public Type EntityType => typeof(T);
 
         /// <summary>
         /// The Action invoked on Collision, with the Component collided with passed as a parameter
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -1,0 +1,51 @@
+﻿using Monocle;
+using System;
+using Microsoft.Xna.Framework;
+
+namespace Celeste.Mod.Entities {
+    [Tracked(false)]
+    public class EntityCollider<T> : Component where T : Entity {
+        public readonly string entityType = typeof(T).Name;
+
+        public Action<T> OnEntityAction;
+
+        public Collider Collider;
+
+        public EntityCollider(Action<T> onEntityAction, Collider collider)
+            : base(active: true, visible: true) {
+            OnEntityAction = onEntityAction;
+            Collider = collider;
+        }
+
+        public override void Added(Entity entity) {
+            if (!Engine.Scene.Tracker.IsEntityTracked<T>()) {
+                patch_Tracker.AddEntityToTracker(typeof(T));
+            }
+            base.Added(entity);
+        }
+
+        public override void Update() {
+            if (OnEntityAction == null) {
+                return;
+            }
+
+            Collider collider = Entity.Collider;
+            if (Collider != null) {
+                Entity.Collider = Collider;
+            }
+
+            Entity.CollideDo(OnEntityAction);
+
+            Entity.Collider = collider;
+        }
+
+        public override void DebugRender(Camera camera) {
+            if (Collider != null) {
+                Collider collider = Entity.Collider;
+                Entity.Collider = Collider;
+                Collider.Render(camera, Color.HotPink);
+                Entity.Collider = collider;
+            }
+        }
+    }
+}
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -16,7 +16,7 @@ namespace Celeste.Mod.Entities {
         public Type EntityType => typeof(T);
 
         /// <summary>
-        /// The Action invoked on Collision, with the Component collided with passed as a parameter
+        /// The Action invoked on Collision, with the Entity collided with passed as a parameter
         /// </summary>
         public Action<T> OnEntityAction;
 
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -1,6 +1,7 @@
 ﻿using Monocle;
 using System;
 using Microsoft.Xna.Framework;
+using static System.Formats.Asn1.AsnWriter;
 
 namespace Celeste.Mod.Entities {
     /// <summary>
@@ -30,7 +31,13 @@ namespace Celeste.Mod.Entities {
 
         public override void Added(Entity entity) {
             base.Added(entity);
-            (Scene?.Tracker as patch_Tracker).Refresh();
+            //Only called if Component is added post Scene Begin and Entity Adding and Awake time.
+            if (Scene != null) {
+                if (!Scene.Tracker.IsEntityTracked<T>()) {
+                    patch_Tracker.AddTypeToTracker(typeof(T));
+                }
+                (Scene.Tracker as patch_Tracker).Refresh();
+            }
         }
 
         public override void EntityAdded(Scene scene) {
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -1,7 +1,6 @@
 ﻿using Monocle;
 using System;
 using Microsoft.Xna.Framework;
-using static System.Formats.Asn1.AsnWriter;
 
 namespace Celeste.Mod.Entities {
     /// <summary>
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -30,7 +30,7 @@ namespace Celeste.Mod.Entities {
 
         public override void EntityAdded(Scene scene) {
             if (!scene.Tracker.IsEntityTracked<T>()) {
-                (scene.Tracker as patch_Tracker).AddEntityToTracker(typeof(T));
+                patch_Tracker.AddTypeToTracker(typeof(T));
             }
             base.EntityAdded(scene);
         }
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -13,7 +13,7 @@ namespace Celeste.Mod.Entities {
         /// <summary>
         /// Provides a simple way to know the Entity type of the specific Collider without Reflection
         /// </summary>
-        public readonly string entityType = typeof(T).Name;
+        public readonly Type entityType = typeof(T);
 
         /// <summary>
         /// The Action invoked on Collision, with the Component collided with passed as a parameter
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -29,7 +29,7 @@ namespace Celeste.Mod.Entities {
         }
 
         public override void EntityAdded(Scene scene) {
-            if (scene.Tracker.IsEntityTracked<T>()) {
+            if (!scene.Tracker.IsEntityTracked<T>()) {
                 (scene.Tracker as patch_Tracker).AddEntityToTracker(typeof(T));
             }
             base.EntityAdded(scene);
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -17,11 +17,11 @@ namespace Celeste.Mod.Entities {
             Collider = collider;
         }
 
-        public override void Added(Entity entity) {
-            if (!Engine.Scene.Tracker.IsEntityTracked<T>()) {
-                patch_Tracker.AddEntityToTracker(typeof(T));
+        public override void EntityAdded(Scene scene) {
+            if (scene.Tracker.IsEntityTracked<T>()) {
+                (scene.Tracker as patch_Tracker).AddEntityToTracker(typeof(T));
             }
-            base.Added(entity);
+            base.EntityAdded(scene);
         }
 
         public override void Update() {
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -3,10 +3,21 @@ using System;
 using Microsoft.Xna.Framework;
 
 namespace Celeste.Mod.Entities {
+    /// <summary>
+    /// Allows for Collision with any type of entity in the game, similar to a PlayerCollider or PufferCollider.
+    /// Performs the Action provided on collision. 
+    /// </summary>
+    /// <typeparam name="T">The specific type of Entity this component should try to collide with</typeparam>
     [Tracked(false)]
     public class EntityCollider<T> : Component where T : Entity {
+        /// <summary>
+        /// Provides a simple way to know the Entity type of the specific Collider without Reflection
+        /// </summary>
         public readonly string entityType = typeof(T).Name;
 
+        /// <summary>
+        /// The Action invoked on Collision, with the Component collided with passed as a parameter
+        /// </summary>
         public Action<T> OnEntityAction;
 
         public Collider Collider;
--- a/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
+++ b/Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
@@ -28,11 +28,20 @@ namespace Celeste.Mod.Entities {
             Collider = collider;
         }
 
+        public override void Added(Entity entity) {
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Monocle;
+using System;
+using Microsoft.Xna.Framework;
+
+namespace Celeste.Mod.Entities {
+    [Tracked(false)]
+    public class EntityCollider<T> : Component where T : Entity {
+        public readonly string entityType = typeof(T).Name;
```
psyGamer: Whats the purpose of this `entityType`/`componentType` string?

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public Action<T> OnEntityAction;
+
+        public Collider Collider;
+
+        public EntityCollider(Action<T> onEntityAction, Collider collider)
```
psyGamer: Might want to proivde a default `Collider collider = null`, since that's what then uses the entity collider instead of a cusotm one.

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
```diff
@@ -0,0 +1,51 @@
...
+            : base(active: true, visible: true) {
+            OnEntityAction = onEntityAction;
+            Collider = collider;
+        }
+
+        public override void Added(Entity entity) {
+            if (!Engine.Scene.Tracker.IsEntityTracked<T>()) {
```
psyGamer: Prefer not using `Engine.Scene` in a context where you are in a scene already.
Rather use `entity.Scene` inside `EntityAdded`.

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,49 @@ private static List<Type> GetSubclasses(Type type) {
...
+                value.Add(type);
+            }
+            foreach (Type subType in subTypes) {
+                if (!TrackedEntityTypes.TryGetValue(subType, out List<Type> subvalue)) {
+                    TrackedEntityTypes[subType] = new List<Type> { type };
+                } else if (!subvalue.Contains(type)) {
+                    subvalue.Add(type);
+                }
+            }
+            Dictionary<Type, List<Entity>> entities = Engine.Scene?.Tracker.Entities;
```
psyGamer: Why is this method static when you access `Engine.Scene?.Tracker` anyway?

psyGamer (commented):
Left some of my thoughs. 
For the issue of a tracker in another scene technically having a type tracked, while that not acctually being the case, perhaps provide a method to refresh tracked entities/components and just let mods handle it themselves?

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            Entity.Collider = collider;
+        }
+
+        public override void DebugRender(Camera camera) {
+            if (Collider != null) {
+                Collider collider = Entity.Collider;
+                Entity.Collider = Collider;
+                Collider.Render(camera, Color.HotPink);
+                Entity.Collider = collider;
```
JaThePlayer, line 67: why is `Entity.Collider` getting set here?

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Monocle;
+using System;
+using Microsoft.Xna.Framework;
+
+namespace Celeste.Mod.Entities {
+    [Tracked(false)]
+    public class EntityCollider<T> : Component where T : Entity {
```
JaThePlayer, line 11: I'd like some documentation with `///` for new public classes.

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,49 @@ private static List<Type> GetSubclasses(Type type) {
...
+            if (entities != null && !entities.ContainsKey(type)) {
+                entities[type] = Engine.Scene.Entities.Where((e) => e.GetType() == type).ToList();
```
JaThePlayer: This doesn't handle inheritance, which could lead to weird edge cases between first room loads and subsequent ones

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,49 @@ private static List<Type> GetSubclasses(Type type) {
...
+                    TrackedComponentTypes[subType] = new List<Type> { type };
+                } else if (!subvalue.Contains(type)) {
+                    subvalue.Add(type);
+                }
+            }
+            Dictionary<Type, List<Component>> components = Engine.Scene?.Tracker.Components;
+            if (components != null && !components.ContainsKey(type)) {
+                List<Component> list = new();
+                foreach (Entity entity in Engine.Scene.Entities) {
+                    list.AddRange(entity.Components.Where((c) => c.GetType() == type));
```
JaThePlayer: (same as with entities, this should handle inheritance)

JaThePlayer (changes requested):

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
psyGamer: So that the collider has the absolute position instead of just entity-relative one.
`PlayerCollider.DebugRender` does the same.

JaThePlayer: oh, right

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
TheDavSmasher: Since all Entity Colliders are Tracked together, it makes it easier to know if a given Collider is of a given type, although I don't know if there's a more elegant solution to it, I'm willing to make it that if I get ideas.

Basically, since it's generic, it's just there for knowledge purposes

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
TheDavSmasher: Forgot about that, will do

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
TheDavSmasher: noted

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
TheDavSmasher: I kinda got this code from somewhere else so I should refactor it to not do that

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
psyGamer: So just for mods which type is actually tracked? In that case just provide the `Type` directly, since `type.Name` is not unique

TheDavSmasher: Ok I can do that

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,109 @@ private static List<Type> GetSubclasses(Type type) {
             // don't hold references to all the types anymore
             _temporaryAllTypes = null;
         }
+
+        public static void AddTypeToTracker(Type type, Type trackedAs = null, bool inheritAll = false) {
+            AddTypeToTracker(type, trackedAs, inheritAll ? GetSubclasses(type).ToArray() : Array.Empty<Type>());
+        }
+
+        public static void AddTypeToTracker(Type type, Type trackedAs = null, params Type[] subtypes) {
```
JaThePlayer, line 98: This method seems like a copy-paste of the everest patch for Initialize, would be nice to rewrite that patch to use this method, to not have so much duplicate logic that then needs to be updated twice if new tracker features get added.

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,109 @@ private static List<Type> GetSubclasses(Type type) {
...
+            if (typeof(Entity).IsAssignableFrom(type) && !Engine.Scene.Tracker.Entities.ContainsKey(type)) {
+                Engine.Scene.Tracker.Entities.Add(type, new List<Entity>());
+            }
+            else if (typeof(Component).IsAssignableFrom(type) && !Engine.Scene.Tracker.Components.ContainsKey(type)) {
+                Engine.Scene.Tracker.Components.Add(type, new List<Component>());
+            } else {
+                throw new Exception("Type '" + type.Name + "' cannot be TrackedAs because it does not derive from Entity or Component");
+            }
+            RefreshTrackerLists();
+        }
```
JaThePlayer: Is there any point to this method? The parameterless `RefreshTracker` method should already handle newly added types.
Also, that exception message definitely doesn't fit what this method does.

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,109 @@ private static List<Type> GetSubclasses(Type type) {
...
+
+        public static void RefreshTracker() {
```
JaThePlayer: Is there any point to this being public? Only case where you'd need to refresh the tracker is when calling `AddTypeToTracker` anyway 🤔

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,109 @@ private static List<Type> GetSubclasses(Type type) {
...
+
+        private static void RefreshTrackerLists() {
+            foreach (Entity entity in Engine.Scene.Entities) {
+                foreach (Component component in entity.Components) {
+                    Type componentType = component.GetType();
+                    if (!TrackedComponentTypes.TryGetValue(componentType, out List<Type> componentTypes)) {
+                        continue;
+                    }
+                    foreach (Type trackedType in componentTypes) {
+                        Engine.Scene.Tracker.Components[trackedType].Add(component);
```
JaThePlayer: Have you tested that this doesn't cause entities/components to get tracked twice?

JaThePlayer: (Safest bet would be to call .Clear on both the Entities and Components dictionaries)

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -123,5 +123,109 @@ private static List<Type> GetSubclasses(Type type) {
...
+            }
+            else {
+                throw new Exception("Type '" + type.Name + "' cannot be TrackedAs because it does not derive from Entity or Component");
+            }
+            RefreshTracker(type);
+        }
+
+        public static void RefreshTracker(Type type) {
+            if (typeof(Entity).IsAssignableFrom(type) && !Engine.Scene.Tracker.Entities.ContainsKey(type)) {
+                Engine.Scene.Tracker.Entities.Add(type, new List<Entity>());
```
JaThePlayer: I'd extract `Engine.Scene.Tracker` to a variable OR make this an instance method (applies to all places in this PR that do this)

JaThePlayer (changes requested):

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
psyGamer: Allows mods to manually refresh the tracker of another scene, since it wouldnt happen automatically.
However it might make sense to automatically refresh the tracker on scene change to avoid having to manually call it?

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// Allows for Collision with any type of entity in the game, similar to a PlayerCollider or PufferCollider.
+    /// Performs the Action provided on collision. 
+    /// </summary>
+    /// <typeparam name="T">The specific type of Entity this component should try to collide with</typeparam>
+    [Tracked(false)]
+    public class EntityCollider<T> : Component where T : Entity {
+        /// <summary>
+        /// Provides a simple way to know the Entity type of the specific Collider without Reflection
+        /// </summary>
+        public readonly Type entityType = typeof(T);
```
psyGamer: `public` fields are usually in `PascalCase` instead of `camelCase`.
Also the "without Reflection" sounds a bit weird, because `Type` _is_ reflection.
Maybe just "Type of the entity which is checked against." (with some better wording)?

Celeste.Mod.mm/Mod/Entities/EntityColliderByComponent.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Collision is done by component, as in, it will get all the components of the type and try to collide with their entities.
+    /// Performs the Action provided on collision. 
+    /// </summary>
+    /// <typeparam name="T">The specific type of Component this component should try to collide with</typeparam>
+    [Tracked(false)]
+    public class EntityColliderByComponent<T> : Component where T : Component {
+        /// <summary>
+        /// Provides a simple way to know the Component type of the specific Collider without Reflection
+        /// </summary>
+        public readonly Type componentType = typeof(T);
```
psyGamer: Same as with `entityType`

psyGamer (changes requested):

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
JaThePlayer: Well in that case this method should definitely be an instance method (called `Refresh()`), otherwise I don't see a sane way to use it on a different scene. This only seems useful in *very* niche cases, (creating a Level inside a Level??) but ig it doesn't hurt (With a comment with `///`)

psyGamer: Oh yeah. It definitly shouldnt be static

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
JaThePlayer: 1. This doesn't need to be a field, `=> typeof(T)` would be fine
2. If you have a variable of type `EntityCollider<T>`, then you already need to have known the `T` at compile-time to declare that variable. As far as I can tell, this is only useful when using reflection to access this field, at which point you might as well use reflection to get the generic parameter 🤔

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
TheDavSmasher: So, do you want me to have the Initialize method use this AddTypeToTracker method?

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
TheDavSmasher: 1) Only to not loop through all types StoredTypes
2) I copied the exception so I can fix the message

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
TheDavSmasher: I can do that

Celeste.Mod.mm/Mod/Entities/EntityCollider.cs
TheDavSmasher: My logic is for when getting the component through the tracker instead of instance

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
TheDavSmasher: My main question would be how do I know if there's a scene change at any time? Besides storing the "old" one in a variable and check against it every frame?

wuke32767:
don't think generic type can be tracked

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -71,57 +73,116 @@ private static List<Type> GetSubclasses(Type type) {
...
+            foreach (Type componentType in StoredComponentTypes) {
+                if (!Components.ContainsKey(componentType)) {
+                    Components.Add(componentType, new List<Component>());
+                }
+            }
+            foreach (Entity entity in Engine.Scene.Entities) {
+                foreach (Component component in entity.Components) {
+                    Type componentType = component.GetType();
+                    if (Components[componentType].Contains(component)
+                        || !TrackedComponentTypes.TryGetValue(componentType, out List<Type> componentTypes)) {
```
JaThePlayer: One final perf thing, I'm pretty sure `Components[componentType]` is a List, so doing a Contains check on it is slower than the dictionary check after the `||` operator. I think flipping the order of these 2 conditions will be a bit better for perf. (Same applies to entity code below)

JaThePlayer (approved):

TheDavSmasher:
USSRNAME (wuke32767) wants to provide a pull request to my changes so if changes come in, it's their pull request code and will be added to the conversation because it includes some MonoMod things I don't fully understand

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -56,72 +67,115 @@ private static List<Type> GetSubclasses(Type type) {
...
+                if (!Entities.ContainsKey(entityType)) {
+                    Entities.Add(entityType, new List<Entity>());
+                }
+            }
+            foreach (Type componentType in StoredComponentTypes) {
+                if (!Components.ContainsKey(componentType)) {
+                    Components.Add(componentType, new List<Component>());
+                }
+            }
+            foreach (Entity entity in Engine.Scene.Entities) {
```
psyGamer: You are using `Engine.Scene` on an instance method of a `Tracker`.
There is no guarantee that `Engine.Scene.Tracker == this`, which is bad for this usecase.
So the options are either adding a `Scene` field and setting that from the constructor, or having a `Scene` parameter which the user needs to pass in.

psyGamer (changes requested):

psyGamer (approved):

TheDavSmasher:
This is ready for final review / merging

PR #866, opened 2024-12-20
Added String method for non-empty string values

TheDavSmasher:
Entity Data objects when getting values out can be a little finicky when trying to get a pure string value out, especially if the value is an empty string. 

It is useful to have this method for one-line assignments for such cases where a string that isn't set is empty, but the Attr method would still return that empty string instead of using the default value

### DIFF
--- a/Celeste.Mod.mm/Patches/EntityData.cs
+++ b/Celeste.Mod.mm/Patches/EntityData.cs
@@ -21,8 +21,8 @@ namespace Celeste {
         /// <returns></returns>
         public string String(string key, string defaultValue = null) {
             if (Values != null && Values.TryGetValue(key, out object value)
-                && value is string val && val.Trim().Length > 0) {
-                    return val;
+                && value != null && value.ToString().Trim().Length > 0) {
+                    return value.ToString();
             }
             return defaultValue;
         }
--- a/Celeste.Mod.mm/Patches/EntityData.cs
+++ b/Celeste.Mod.mm/Patches/EntityData.cs
@@ -21,7 +21,7 @@ namespace Celeste {
         /// <returns></returns>
         public string String(string key, string defaultValue = null) {
             string valueStr;
-            if (Values != null && Values.TryGetValue(key, out object value) && !string.IsNullOrEmpty(valueStr = value?.ToString())) {
+            if (Values != null && Values.TryGetValue(key, out object value) && !string.IsNullOrWhiteSpace(valueStr = value?.ToString())) {
                     return valueStr;
             }
             return defaultValue;
--- a/Celeste.Mod.mm/Patches/EntityData.cs
+++ b/Celeste.Mod.mm/Patches/EntityData.cs
@@ -21,7 +21,7 @@ namespace Celeste {
         /// <returns></returns>
         public string String(string key, string defaultValue = null) {
             if (Values != null && Values.TryGetValue(key, out object value)
-                && value is string val && !string.IsNullOrWhiteSpace(val)) {
+                && value is string val && val.Trim().Length > 0) {
                     return val;
             }
             return defaultValue;
--- a/Celeste.Mod.mm/Patches/EntityData.cs
+++ b/Celeste.Mod.mm/Patches/EntityData.cs
@@ -20,8 +20,7 @@ namespace Celeste {
         /// <param name="defaultValue"></param>
         /// <returns></returns>
         public string String(string key, string defaultValue = null) {
-            if (Values != null && Values.TryGetValue(key, out object value)
-                && value != null && value.ToString().Trim().Length > 0) {
+            if (Values != null && Values.TryGetValue(key, out object value) && !string.IsNullOrEmpty(value?.ToString())) {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/EntityData.cs
```diff
@@ -12,5 +12,19 @@ class patch_EntityData : EntityData {
...
+        /// <summary>
+        /// Get the <see cref="T:System.String" /> value associated with a key.
+        /// An Empty or pure Whitespace value will result in the <paramref name="defaultValue"/> to be returned.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <param name="defaultValue"></param>
+        /// <returns></returns>
+        public string String(string key, string defaultValue = null) {
+            if (Values != null && Values.TryGetValue(key, out object value)) {
+                if (!string.IsNullOrWhiteSpace((string) value))
```
wuke32767: if `value` is not string, it would crash

TheDavSmasher: I'll add a type pattern to see if it's a string, would that work?

wuke32767: you need tostring i guess

TheDavSmasher: so `!string-IsNullOrWhiteSpace(value.ToString())` ?

TheDavSmasher: Or is there a way to do pattern matching

wuke32767:
as other methods always convert the result using `ToString` and `Parse`, i would suggest `ToString`

TheDavSmasher:
So, do a manual null check and then just do value.ToString()?

Celeste.Mod.mm/Patches/EntityData.cs
```diff
@@ -12,5 +12,19 @@ class patch_EntityData : EntityData {
...
+        /// <summary>
+        /// Get the <see cref="T:System.String" /> value associated with a key.
+        /// An Empty or pure Whitespace value will result in the <paramref name="defaultValue"/> to be returned.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <param name="defaultValue"></param>
+        /// <returns></returns>
+        public string String(string key, string defaultValue = null) {
+            if (Values != null && Values.TryGetValue(key, out object value)
+                && value != null && value.ToString().Trim().Length > 0) {
```
psyGamer: `value != null && value.ToString().Trim().Length > 0` can literally just be `!string.IsNullOrWhiteSpace(value?.ToString())`, which I feel convays intent more clearly.
Also it might reduce a string allocation by not calling `Trim()`

Also I saw you used it earlier, just because you already checked for null doesn't mean doing the check manually is nessecarly better.

psyGamer (changes requested):

Celeste.Mod.mm/Patches/EntityData.cs
TheDavSmasher: It is simpler but no string allocation simplicity because that's literally what IsNullOrWhiteSpace does to check white space

psyGamer: `string.IsNullOrWhiteSpace` loops over the string and does not allocate.
Also did you intend on using `string.IsNullOrEmpty`?

TheDavSmasher: No because essentially empty strings but only white space strings should be considered empty if you're trying to get a string out, I believe

Celeste.Mod.mm/Patches/EntityData.cs
```diff
@@ -12,5 +12,18 @@ class patch_EntityData : EntityData {
...
+        /// <summary>
+        /// Get the <see cref="T:System.String" /> value associated with a key.
+        /// An Empty or pure Whitespace value will result in the <paramref name="defaultValue"/> to be returned.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <param name="defaultValue"></param>
+        /// <returns></returns>
+        public string String(string key, string defaultValue = null) {
+            if (Values != null && Values.TryGetValue(key, out object value) && !string.IsNullOrEmpty(value?.ToString())) {
+                    return value.ToString();
```
wuke32767: value.ToString is called twice. better to reduce it to once.

wuke32767 (approved):
but im not a member of everest team

Celeste.Mod.mm/Patches/EntityData.cs
psyGamer: ..I don't understand what you're trying to say? 
White-space only strings should fall back to the default value, right? In that case `string.IsNullOrEmpty` doesn't cover that case, but `string.IsNullOrWhiteSpace` does.

TheDavSmasher: That's exactly what I said but phased way better

psyGamer (approved):

PR #869, opened 2024-12-24
Log warnings on unintended extensions

SnipUndercover:
This PR logs warnings when a file has a doubled file extension (like `example.png.png`) or a text-based file with an extra .txt extension (`everest.yaml.txt`).
Additionally rewritten to use `Span<T>`s and more modern syntax. I haven't run any benchmarks however.
Tested with several mods and example file names.

I put `example.txt.txt`, `example.yaml.txt` and `Audio/test.guids.txt.txt`, and the logs yield the following:
```
(12/24/2024 01:17:29) [Everest] [Warn] [Content] "example.txt.txt" has a doubled extension! It may not be handled correctly.
(12/24/2024 01:17:29) [Everest] [Warn] [Content] "example.yaml.txt" has an extra ".txt" extension! It may not be handled correctly.
(12/24/2024 01:17:29) [Everest] [Warn] [Content] "test.guids.txt.txt" has an extra ".txt" extension! It may not be handled correctly.
```

## Extra considerations
- Should the span helpers be public? They seem useful for mods to use.
- A sore point is that I can't find a sensible way to include which mod has those bad assets, and I can't just change the public API...

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
@@ -725,108 +725,316 @@ namespace Celeste.Mod {
             public static string GuessType(string file, out Type type, out string format) {
                 type = typeof(object);
                 format = Path.GetExtension(file) ?? "";
-                if (format.Length >= 1)
-                    format = format.Substring(1);
 
-                // Assign game asset types
-                if (format == "dll") {
+                if (format.Length < 1)
+                    return file;
+
+                format = format[1..];
+
+                ReadOnlySpan<char> fileSpan = file.AsSpan();
+                int fileSeparator = fileSpan.LastIndexOf('/') + 1;
+
+                // folder with / at the end
+                ReadOnlySpan<char> directorySpan = fileSpan[..fileSeparator]; // don't use Path.GetDirectoryName as it replaces '/' with '\' :catplush:
+                // file
+                ReadOnlySpan<char> fileNameSpan = fileSpan[fileSeparator..];
+                // file without extension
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
```diff
@@ -725,108 +725,316 @@ public static void Add(string path, ModAsset metadata)
...
+                int fileSeparator = fileSpan.LastIndexOf('/') + 1;
+
+                // folder with / at the end
+                ReadOnlySpan<char> directorySpan = fileSpan[..fileSeparator]; // don't use Path.GetDirectoryName as it replaces '/' with '\' :catplush:
+                // file
+                ReadOnlySpan<char> fileNameSpan = fileSpan[fileSeparator..];
+                // file without extension
+                ReadOnlySpan<char> fileNameOnlySpan = Path.GetFileNameWithoutExtension(fileNameSpan);
+
+                bool shouldSendWarning = true;
```
microlith57: this varname is confusing; i'd recommend something like `alreadyWarned` & invert the behaviour

microlith57 (approved):

Celeste.Mod.mm/Mod/Everest/Everest.Content.cs
SnipUndercover: Fair - also I noticed that the warning only logs the file name instead of its path, which is not the best.
My biggest gripe is that I can't access the information telling me what mod the file path belongs to.

microlith57: hmm yea, having the path would be nice. i wouldn't say it's a dealbreaker but would be good to look into

microlith57 (approved):

PR #872, opened 2025-01-09
Dispose the zip file after disposing its entries

SnipUndercover:
Apparently this fixes the issue where in-game mod updates just don't work for come mods (like CommunalHelper/BounceHelper), and from empirical testing this seems to fix this.

The true cause is not known however; see [this Celestecord message](https://discord.com/channels/403698615446536203/429775439423209472/1327036477166718996) for context.


SnipUndercover:
Turns out I just got lucky, because I just replicated the issue after the "fix".

PR #873, opened 2025-01-10
Fix Decal Registry code mod hot reload

JaThePlayer:
Fixes a bug where the Decal Registry doesn't get reloaded when a mod adds a new handler at runtime (mostly due to code mod hot reload), which would cause them to not work/use old behavior

Also added new `Everest.Events.AssetReload.OnAfterNextReload` and `OnBeforeNextReload` events, as they were useful here and I've wanted them in Frost Helper iirc, and they just seem useful in general.


microlith57 (approved):
lgtm :+1:

PR #874, opened 2025-01-29
Fix incorrect extension matcher

SnipUndercover:
I managed to miss an extension matcher for `.obj.export` files, which caused some mountain models (for mods like Into The Jungle) to not load correctly.
The fix has been tested and works correctly.


microlith57 (approved):
:+1:

PR #879, opened 2025-02-08
Entity.SourceData and SourceId, provide way to efficiently search by EntityData name

JaThePlayer:
Second take on the TypeHelper PR, considering the previous one is dead.

- Added `Entity.SourceData`, `Entity.SourceId`
- Added `Scene.FindEntitiesWithSid(string sid)`, which filters by entity data name, while still using the tracker.
- Added `Tracker.Get[Entities/Components]TrackIfNeeded`, which add the type to the tracker if needed, used by the `FindEntitiesWithSid` method, but is useful in general.
- Added a new `Celste.Mod.Registry.EntityRegistry` static class, with functions allowing for checking relations between c# types and EntityData names.
   These relations are generated as follows:
   - Vanilla entities are hardcoded
   - [CustomEntity] pointing to a ctor get registered automatically on mod load.
   - Dynamically as they occur, which guarantees they work even with the legacy `Everest.Events.Level.LoadEntity` event. This does mean that the sets returned by them are not guaranteed to be exhaustive - there could be relations that haven't occurred this session. However, that is fine realistically, because the main point of these APIs is to use them together with the tracker, in which case you don't care about searching for entity types that have never been instantiated anyway.
 - It seems like the PR which added `Tracker.Refresh/AddTypeToTracker` wasn't properly tested? In my testing, the function never did anything, and forcing a refresh caused a missing dictionary key crash for triggers. This PR fixes those bugs as well because these functions are crucial for `FindEntitiesWithSid`
 
 
image
image


JaThePlayer:
Tracker bugfix got moved to #883, which should be merged before this PR

psyGamer (approved):
Appears to be working when testing as a replacement for CelesteTAS' own system.
There are some differences, like the player not having an ID or a missing special-case for the cutscene-spawned feather at the start of 6A, but since they don't actually have any ID, I think it's good if Everest avoids special-cases and leaves that up to mods using it.
Didn't test the other parts intensivly, but from a quick glance, they seem to be working as expected.

maddie480 (approved):

PR #882, opened 2025-02-21
Updated Korean translation

lDMDiamondl:
Updated some translations, but I was not able to check if additional fonts are needed, sorry!


maddie480:
Needs updating fonts: the font generator came up with 26 characters and we currently have 20
https://maddie480.ovh/celeste/task-tracker/font-generate/edf97ee9-6c78-4586-8581-6ed8d1b8d872/download/0

PR #884, opened 2025-02-27
logs the total of loaded modules at launch

AAA1459:


Wartori54:
Could be merged with the following log, or otherwise the second "loader" capitalized.

Wartori54:
I meant the "loaded" in the second line, so that the sentence is capitalized, sorry.
Like `Logger.Info("loader", $"Loaded {Everest._Modules.Count} modules");`

You also seem to have converted all line endings from LF to CRLF accidentally.

AAA1459:
> You also seem to have converted all line endings from LF to CRLF accidentally.

hmm don't know why that happened, got rip of it.

does you mean like this now?

Wartori54 (approved):
LGTM
The line 1 change is due to the BOM characters being removed, but that's fine since there is no consistent pattern in Everest anyway.

maddie480 (approved):

PR #886, opened 2025-03-23
Add soundParam attribute for the Tileset XMLs + force a NoInlining on Player.DustParticleFromSurfaceIndex

SilverDorian46:
These changes are copied over from #867 to be merged in advance while the custom Particles implementation is being sorted out.
The additions include the following:
- **the `soundParam` attribute for `Tileset`:** 
  Makes this surface index play a different sound, by setting the `"surface_index"` parameter to the value. For example, setting a value of `"40"` makes it play the Glitch sound. The purpose of this is to make new surface indices play already existing sounds without having to set up a different event for them. This is also useful in cases where a different event is used with the `soundPath` attribute.
  Note: Defining a `sound` attribute is required to use this, and it's recommended to set the value for that `sound` attribute greater than the range of already defined vanilla sounds *(`50` is a good number to start from)*
- **forced `NoInlining` on `Player.DustParticleFromSurfaceIndex`:**
  Makes `Player.DustParticleFromSurfaceIndex` not inline, meaning it can be hooked by code mods without any problems. For the time being, this can be used by coders to implement their own dust particles


Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -70,13 +70,27 @@ private void ReadInto(patch_TerrainType data, Tileset tileset, XmlElement xml) {
...
 
-            if (xml.HasAttr("soundPath") && xml.HasAttr("sound")) { // Could accommodate for no sound attr, but requiring it should improve clarity on user's end
+            /*if (xml.HasAttr("soundPath") && xml.HasAttr("sound")) { // Could accommodate for no sound attr, but requiring it should improve clarity on user's end
                 SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
                 patch_SurfaceIndex.IndexToCustomPath[xml.AttrInt("sound")] = (xml.Attr("soundPath").StartsWith("event:/") ? "" : "event:/") + xml.Attr("soundPath");
             } else if (xml.HasAttr("sound")) {
                 SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
             } else if (!SurfaceIndex.TileToIndex.ContainsKey(xml.AttrChar("id"))) {
                 SurfaceIndex.TileToIndex[xml.AttrChar("id")] = 0; // fall back to no sound
+            }*/
```
Wartori54: If this is rewritten, is there any need to keep the old code commented?

Wartori54 (commented):
LGTM
(I'll approve once that commented out code is handled)

Celeste.Mod.mm/Patches/Autotiler.cs
SilverDorian46: I guess not

I'd remove it, but I don't have the project at the moment - I've had to get a new computer

maddie480: Someone could edit that out with the online code editor then

SnipUndercover: Let's see if I could do that then.

SnipUndercover: Looks like I didn't break everything, marking conversation as resolved.

Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -69,12 +69,16 @@ private void ReadInto(patch_TerrainType data, Tileset tileset, XmlElement xml) {
...
             }
-
-            if (xml.HasAttr("soundPath") && xml.HasAttr("sound")) { // Could accommodate for no sound attr, but requiring it should improve clarity on user's end
-                SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
-                patch_SurfaceIndex.IndexToCustomPath[xml.AttrInt("sound")] = (xml.Attr("soundPath").StartsWith("event:/") ? "" : "event:/") + xml.Attr("soundPath");
-            } else if (xml.HasAttr("sound")) {
+            
+            if (xml.HasAttr("sound")) {
                 SurfaceIndex.TileToIndex[xml.AttrChar("id")] = xml.AttrInt("sound");
+                // Could accommodate for no sound attr, but requiring it should improve clarity on user's end
```
maddie480: what does this comment refer to now?

maddie480 (approved):

Celeste.Mod.mm/Patches/Autotiler.cs
SnipUndercover: Seems like this comment was introduced in 5874a6c, where there was no `else if (xml.HasAttr("sound"))` branch, so I'd say this is irrelevant now. I'll remove it.

JaThePlayer (approved):

Wartori54 (approved):

PR #888, opened 2025-04-07
Add option to lock file to current levelset (#588)

### LINKED ISSUE #588: Option to lock save files to the first map you play them with

No

### Describe your request

It's surprisingly easy to lose track of which map you were playing a specific save file with.

 * If you delete a mod that contains a world map, the save file will automatically switch to a different map
 * If you accidentally toggle the world map switcher, you might forget which map you're playing on.

I'd like the option to disable changing the world map after beginning one of them. Switching away from the vanilla map would be allowed only if you haven't proceeded past Prologue.

### Additional context

The save files have a `LastArea_Safe` tag that contains the world map name, so I think it wouldn't be necessary to make any changes to save files to implement this. Just don't allow changing it (except after completing Prologue in vanilla).

DashingCat:
Adds an option to lock an existing save file to its current levelset (preventing the user from switching levelset).

This option can be toggled from the `FileSelectSlot` OUI, or using the `lock` command (like the `assist` or `variants` commands).

The lock status is saved in the `CoreModuleSaveData`.

I'm wondering if adding a button can causes issues in TASes, if they don't expect it.

Closes #588.

### DIFF
--- a/Celeste.Mod.mm/Content/Dialog/English.txt
+++ b/Celeste.Mod.mm/Content/Dialog/English.txt
@@ -48,6 +48,9 @@
 	MENU_TITLESCREEN_RESTART_VANILLA= Restarting into orig/Celeste.exe
 	MENU_TITLESCREEN_RESTART_VANILLA_SAVES_WARN= WARNING: Because of OS limitations, saves won't be shared!
 
+# Save file slot options
+	MENU_SAVEFILE_LOCKED= Locked
+
 # Extra Key Mapping
     KEY_CONFIG_ADDING= PRESS ADDITIONAL KEY FOR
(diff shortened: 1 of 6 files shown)


maddie480 (changes requested):
Hm, I'm not sure players would understand what an option just called "Locked" does, especially since the level set selector isn't there for existing files :thinking: "Lock to current mod"?
Also, what happens if the file is "locked" in a mod that has been disabled? You would probably end up opening vanilla instead

DashingCat:
Thanks for the feedback, I'm fine with changing the wording, but I'm not quite sure how to handle disabled mods, so I'm marking the PR as draft for now.

SnipUndercover:

PR #891, opened 2025-04-17
Install Mono for the build phase

maddie480:
I'm not 100% sure this will fix the build, but I need a PR to test the pipeline anyway.


maddie480:
this does check out with this little warning though

>  Some tools (e.g. Mono, NuGet, Terraform) are not available on the image. Therefore some tasks do not run or have reduced functionality.

maddie480:
wahoo it works

PR #892, opened 2025-04-17
Add fast paths to Monocle.Collide methods to handle Lists and arrays without allocations

JaThePlayer:
Because `Collide.Check`/`Collide.First` accept an `IEnumerable<Entity>` instead of a concrete type, there's always a heap allocation of an `IEnumerator<Entity>` and tons of virtual calls in these methods.

This PR adds new overloads and changes these methods to try automatically switching to a span-based loop, dodging that allocation and improving perf for the most common case of a `List<Entity>` (from the Tracker) being passed to these methods.

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/Collide.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/Collide.cs
@@ -15,19 +15,13 @@ public static class patch_Collide {
     
     [MonoModReplace]
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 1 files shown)


Wartori54 (approved):
Looks good, no hooks on any of the replaced methods I believe.
As a side note, I'm curious on how would performance fluctuate by removing all the `NoInlining` from `Monocle.Collide`'s other methods (and pushing it to `AggressiveInlining` maybe). There are known hooks in `Collide(Entity, Entity)` though.

maddie480 (approved):

PR #899, opened 2025-05-01
Consider missing vanilla savefile corrupted (#649)

### LINKED ISSUE #649: Events.FileSelectSlot.OnCreateButtons not applying settings for new saves

Yes

Yes

### Describe the bug

I am using https://github.com/EverestAPI/Resources/wiki/Everest-Events#fileselectslot for creating options on file slots. It works great for editing values on already existing saves. Unfortunately it looks like options set while creating a new save are not getting applied (or are getting erased when the user selects begin), which I'm guessing is a bug.

### Steps to reproduce

1. Create a mod with the wiki's example code: https://github.com/EverestAPI/Resources/wiki/Everest-Events#fileselectslot
2. Open Celeste and select a blank file slot.
3. Set Silhouette Mode to true then select "Begin".
4. Save and quit, and observe that the created save file has Silhouette Mode set to false.
5. At this point, setting it to true and selecting "Continue" will update the save data as expected.

### Expected behavior

4. The save file should have Silhouette Mode set to true.

### Operating System

Linux

### Everest Version

4107

_No response_

### Additional context

_No response_

DashingCat:
Commit https://github.com/EverestAPI/Everest/commit/fa9b476b16ef7d82c5c6ab8d3b50b79683f87922 always delete modded save data when starting a new game.

This PR changes this behavior:
- Modded save data is no longer deleted when starting a new save file.
- Instead, save slots without vanilla save data and with modded save data are considered corrupted, making the slot unusable until all save data (vanilla and modded) is deleted.

This allows the `Everest.Events.FileSelectSlot.OnCreateButtons` event to work properly on new save files, as modded save data is no longer deleted after being created.

Fixes #649.

### DIFF
--- a/Celeste.Mod.mm/Patches/OuiFileSelect.cs
+++ b/Celeste.Mod.mm/Patches/OuiFileSelect.cs
@@ -8,11 +8,29 @@ using System.Linq;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using MonoMod.Cil;
+using Celeste.Mod;
 
 namespace Celeste {
     class patch_OuiFileSelect : OuiFileSelect {
 
         internal bool startingNewFile;
+
+        public extern void orig_LoadThread();
+        public new void LoadThread() {
+            orig_LoadThread();
+
+            string saveFilePath = patch_UserIO.GetSaveFilePath();
+            if (!Directory.Exists(saveFilePath))
+                return;
+
+            for (int i = 0; i < Slots.Count() - 1; i++) { // - 1, last slot is always empty
+                if (!Slots[i].Exists && Directory.GetFiles(saveFilePath, $"{i}-mod*.celeste").Length > 0) {
+                    Logger.Warn("save", $"Save slot {i} have modded data but no save file, flagging as corrupted");
+                    Slots[i].Corrupted = true;
+                    Slots[i].Exists = true; // both `Corrupted` and `Exists` flags must be set
+                }
+            }
+        }
 
         [PatchOuiFileSelectSubmenuChecks] // we want to manipulate the orig method with MonoModRules
         public extern IEnumerator orig_Enter(Oui from);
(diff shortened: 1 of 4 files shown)


Wartori54:
I have a minor issue with this, which is: in the case where users have deleted their savefiles in vanilla, once they hop on Everest again, they would see like their deleted savefiles appear as corrupted. Which could lead to confusion, especially once this hits stable, where potentially multiple blank savefiles may become corrupted unexpectedly.
I think it could be better to convey to the user what's actually happening, given that the files aren't really corrupted, just partially deleted. So replacing the text when a savefile is partially deleted from "Corrupted" to something like "Partially deleted" would be much better.

DashingCat:
Thanks for the feedback, I clarified the text in the latest commit, here is a screenshot below:

image

Celeste.Mod.mm/Patches/OuiFileSelect.cs
```diff
@@ -8,12 +8,31 @@
...
+        public new void LoadThread() {
+            orig_LoadThread();
+
+            string saveFilePath = patch_UserIO.GetSaveFilePath();
+            if (!Directory.Exists(saveFilePath))
+                return;
+
+            for (int i = 0; i < Slots.Count() - 1; i++) { // - 1, last slot is always empty
+                if (!Slots[i].Exists && Directory.GetFiles(saveFilePath, $"{i}-mod*.celeste").Length > 0) {
+                    Logger.Warn("save", $"Save slot {i} have modded data but no save file, flagging as corrupted");
```
Wartori54: This sentence is a bit wrong, something like "Save slot {i} has modded data but no vanilla save data, flagging as corrupted" would be better.

Also using `save` as the log id seems a bit vague, `OuiFileSelect` or `SaveData` could be better.

Celeste.Mod.mm/Patches/SaveData.cs
```diff
@@ -297,8 +298,17 @@ public static void LoadModSaveData(int slot) {
...
-                return false;
+            bool vanillaExists = false;
+            string saveFilePath = patch_UserIO.GetSaveFilePath();
+            if (Directory.Exists(saveFilePath))
+                vanillaExists = File.Exists(Path.Combine(saveFilePath, $"{slot}.celeste"));
+
+            if (vanillaExists)
+                if (!orig_TryDelete(slot))
+                    return false;
+            else
```
Wartori54, line 311: Caution with brace-less statements, there is a bug here, the `else` statement adheres to the latest `if` instead of the first one (as indentation suggests). Please use braces here.

Celeste.Mod.mm/Patches/SaveData.cs
```diff
@@ -297,8 +298,17 @@ public static void LoadModSaveData(int slot) {
...
+                Logger.Warn("save", $"Deleting save data for slot {slot} which have no vanilla data");
```
Wartori54: Same as above for `save`. And here the "have" should be a "has".

Wartori54 (changes requested):
Tested it and it seems to work, it just needs a few logging changes.

DashingCat:
Thanks for the review, I fixed the `if` statement and log wording in the latest commit.

Wartori54 (approved):

microlith57 (approved):

PR #904, opened 2025-05-10
BOOT: When restarting the game, do not pass executable as command line argument

maddie480:
The doc says:

image

and now people's Steams are screaming about an unknown "Celeste.dll" argument being passed :destareline:


maddie480:
tested on Linux, needs testing on Windows

PR #906, opened 2025-05-12
Keep FakeWalls from hiding TalkComponentUI (#759)

### LINKED ISSUE #759: FakeWall hides TalkComponentUI

Yes

Yes

### Describe the bug

**Bug description**
If we have a NPC (which holds a TalkComponentUI), standing exactly on a FakeWall, and if the FakeWall is not uncovered, then the talk bubble will not show up
celeste

### Steps to reproduce

**Why we have this bug**
TalkComponentUI.Update will check if its Handler.Entity.Position (Position of NPC in this case) is covered by a FakeWall. If is, then its alpha will always be 0.
`
if (alpha < 1f && Handler.Entity != null && !base.Scene.CollideCheck<FakeWall>(Handler.Entity.Position)){
	alpha = Calc.Approach(alpha, 1f, 2f * Engine.DeltaTime);
}
`
Due to the minor differences between `Hitbox.Collide(Vector2 point)` vs `Hitbox.Collide(Hitbox hitbox)`,
`Scene.CollideCheck<FakeWall>(Handler.Entity.Position)` will return true here.

**How to patch this**
maybe just change it to
`Scene.CollideCheck<FakeWall>(Handler.Entity.Position - Vector2.UnitY)`
in `TalkComponentUI.Awake/Update`

**Why we should patch this**
Coz such kind of structure is not uncommon in mod maps

### Expected behavior

celeste2

### Operating System

Win 10

### Everest Version

4713

No

### Additional context

_No response_

DashingCat:
Prevent `Fakewalls` from unexpectedly hiding `TalkComponentUI`.

This PR uses the same collision as the `TalkComponent` entity.

Fixes #759.

### DIFF
--- a/Celeste.Mod.mm/Patches/TalkComponentUI.cs
+++ b/Celeste.Mod.mm/Patches/TalkComponentUI.cs
@@ -1,0 +1,48 @@
+#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
+
+using Microsoft.Xna.Framework;
+using Monocle;
+using System;
+using MonoMod;
+
+namespace Celeste {
+    class patch_TalkComponent : TalkComponent {
+
+        [MonoModConstructor]
+        [MonoModIgnore]
+        public patch_TalkComponent(Rectangle bounds, Vector2 drawAt, Action<Player> onTalk, HoverDisplay hoverDisplay = null)
+            : base(bounds, drawAt, onTalk, hoverDisplay) {
+            // no-op. MonoMod ignores this - we only need this to make the compiler happy.
+        }
+
+        public class patch_TalkComponentUI : TalkComponentUI {
+            private float alpha;
+            private float slide;
+            private float timer;
+
+            [MonoModConstructor]
+            [MonoModIgnore]
+            public patch_TalkComponentUI(TalkComponent handler) : base(handler) {
+                // no-op. MonoMod ignores this - we only need this to make the compiler happy.
+            }
+
+            public extern void orig_Awake(Scene scene);
+            public override void Awake(Scene scene) {
+                orig_Awake(scene);
+                if (Handler.Entity == null || Scene.CollideCheck<FakeWall>(new Rectangle((int)(Handler.Entity.X + Handler.Bounds.X), (int)(Handler.Entity.Y + Handler.Bounds.Y), Handler.Bounds.Width, Handler.Bounds.Height))) {
+                    alpha = 0f;
+                }
+            }
+
+            public extern void orig_Update();
+            public override void Update() {
+                timer += Engine.DeltaTime;
+                slide = Calc.Approach(slide, Display ? 1 : 0, Engine.DeltaTime * 4f);
+                if (alpha < 1f && Handler.Entity != null && !Scene.CollideCheck<FakeWall>(new Rectangle((int)(Handler.Entity.X + Handler.Bounds.X), (int)(Handler.Entity.Y + Handler.Bounds.Y), Handler.Bounds.Width, Handler.Bounds.Height))) {
+                    alpha = Calc.Approach(alpha, 1f, 2f * Engine.DeltaTime);
+                }
+                orig_Update();
+            }
+        }
+    }
+}
--- a/Celeste.Mod.mm/Patches/TalkComponent.cs
+++ b/Celeste.Mod.mm/Patches/TalkComponent.cs
@@ -1,5 +1,3 @@
-#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
-
 using Microsoft.Xna.Framework;
 using Monocle;
 using System;
@@ -16,6 +14,9 @@ namespace Celeste {
         }
 
         public class patch_TalkComponentUI : TalkComponentUI {
+            // use the same collision as the TalkComponent entity
+            // (see https://github.com/EverestAPI/Everest/issues/759)
+
             private float alpha;
             private float slide;
             private float timer;
@@ -26,22 +27,28 @@ namespace Celeste {
                 // no-op. MonoMod ignores this - we only need this to make the compiler happy.
             }
 
-            public extern void orig_Awake(Scene scene);
+            [MonoModLinkTo("Monocle.Entity", "Awake")]
+            [MonoModIgnore]
+            public extern void base_Awake(Scene scene);
+            [MonoModReplace]
             public override void Awake(Scene scene) {
-                orig_Awake(scene);
+                base_Awake(scene);
                 if (Handler.Entity == null || Scene.CollideCheck<FakeWall>(new Rectangle((int)(Handler.Entity.X + Handler.Bounds.X), (int)(Handler.Entity.Y + Handler.Bounds.Y), Handler.Bounds.Width, Handler.Bounds.Height))) {
                     alpha = 0f;
                 }
             }
 
-            public extern void orig_Update();
+            [MonoModLinkTo("Monocle.Entity", "Update")]
+            [MonoModIgnore]
+            public extern void base_Update();
+            [MonoModReplace]
             public override void Update() {
                 timer += Engine.DeltaTime;
                 slide = Calc.Approach(slide, Display ? 1 : 0, Engine.DeltaTime * 4f);
                 if (alpha < 1f && Handler.Entity != null && !Scene.CollideCheck<FakeWall>(new Rectangle((int)(Handler.Entity.X + Handler.Bounds.X), (int)(Handler.Entity.Y + Handler.Bounds.Y), Handler.Bounds.Width, Handler.Bounds.Height))) {
                     alpha = Calc.Approach(alpha, 1f, 2f * Engine.DeltaTime);
                 }
-                orig_Update();
+                base_Update();
             }
         }
     }
--- a/Celeste.Mod.mm/Patches/TalkComponent.cs
+++ b/Celeste.Mod.mm/Patches/TalkComponent.cs
@@ -14,8 +14,6 @@ namespace Celeste {
         }
 
         public class patch_TalkComponentUI : TalkComponentUI {
-            // use the same collision as the TalkComponent entity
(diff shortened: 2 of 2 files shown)


Celeste.Mod.mm/Patches/TalkComponentUI.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            [MonoModConstructor]
+            [MonoModIgnore]
+            public patch_TalkComponentUI(TalkComponent handler) : base(handler) {
+                // no-op. MonoMod ignores this - we only need this to make the compiler happy.
+            }
+
+            public extern void orig_Awake(Scene scene);
+            public override void Awake(Scene scene) {
+                orig_Awake(scene);
```
Wartori54: Should not call the original method when the intention is to overwrite the behaviour. An IL patch could be considered as well, fortunately here it seems like there are no mods that ilhook this method, so this is still safe.

The trick here would be to annotate this method with `[MonoModReplace]` and get rid of `orig_Awake` entirely, then, in the first line, you would do something similar to [this](https://github.com/EverestAPI/Everest/blob/f06b3da50c52053288bb71c8634a8fbdb82e4bd2/Celeste.Mod.mm/Patches/Puffer.cs#L11), in order to emit a `base.Awake()` call.

Celeste.Mod.mm/Patches/TalkComponentUI.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            public extern void orig_Update();
+            public override void Update() {
+                timer += Engine.DeltaTime;
+                slide = Calc.Approach(slide, Display ? 1 : 0, Engine.DeltaTime * 4f);
+                if (alpha < 1f && Handler.Entity != null && !Scene.CollideCheck<FakeWall>(new Rectangle((int)(Handler.Entity.X + Handler.Bounds.X), (int)(Handler.Entity.Y + Handler.Bounds.Y), Handler.Bounds.Width, Handler.Bounds.Height))) {
+                    alpha = Calc.Approach(alpha, 1f, 2f * Engine.DeltaTime);
+                }
+                orig_Update();
```
Wartori54: Above comment also applies here.

Wartori54 (changes requested):

DashingCat:
Thanks for the review, I implemented the requested changes in the latest commit.

Wartori54 (dismissed):
LGTM

SnipUndercover (commented):
LGTM, but I'd want to at least test this first.

Wartori54:
I did test it on my end, the glyph and overlay are synced now. 

Removing the quirk where the precise placement shown in the issue hides the UI may be beyond the scope of this PR, and of Everest, since it appears to be a quirk in the game's collision system itself. Although that's a guess.
I'd love to be proven wrong if anyone looks more deeply into this.

DashingCat:
I replicated the setup shown in the issue (`Lookout` on top of `Fakewall`), and here are screenshots without and with the changes applied:

- Without changes (`TalkComponentUI` not displayed, only the binding is shown when the player is close):
image
image

- With changes (`TalkComponentUI` displayed):
image
image

maddie480:
I'm going to approve then reject for a test, don't mind me

maddie480 (approved):

maddie480 (dismissed):
don't worry, this is a test

maddie480:
test

maddie480 (commented):
I would like to dismiss my change request GitHub please thanks

SnipUndercover (dismissed):
Looks good then.

JaThePlayer (approved):

Wartori54:
After some more rough testing looks like this doesn't work in all cases. First I want to apologize for claiming that it did not show up, I must have been tricked by a dark background, because the issue is still present with a Custom NPC from Everest.
Due to how massive the `Bounds` of the `TalkComponent` of that entity are (16x8 tiles using the default texture) you can easily partially cover that rectangle with a fake wall, and also enter it with the Player (with it not colliding with the `FakeWall` of course), causing the glyph to render while the overlay does not. You can also trigger this with a `Lookout`, see the images.
image

This is also why I saw the behaviour not change for Custom NPC, because their `Bounds` extends below ground.

The fix for this could be to strengthen the condition in `TalkComponent.Update` to not set `Highlighted` to `true` if `Bounds` is partially covered by a `FakeWall`, or, conversely, to also set `alpha` to 1f if the player manages to collide with `Bounds` while having it partially covered.
To potentially not "break" many maps, I think the first option is more promising, since, when the glyph renders but the overlay does not, it already hard enough to notice that if the glyph doesn't render anymore it won't be much noticeable.

Wartori54 (changes requested):
See my latest comment before this.

DashingCat:
Thanks for the feedback and extensive testing, I should have tested with the `CustomNPC` on my end.

About the suggested fixes and current state, here is a video with each proposal in a similar setup, and some comments:

https://github.com/user-attachments/assets/763208be-2a53-4f91-8595-6a57b4e98f9d


- current behavior (behavior 0; timestamp 00:08):

1. Player far: only the `Lookout` have a small overlay
2. Player in `Bounds` of `Lookout` near `FakeWall`: both glyph and overlay
3. Player in `Bounds` of `CustomNPC` above `FakeWall`: glyph but no overlay
4. Player in `Bounds` of `CustomNPC` inside `FakeWall`: glyph but no overlay

Behavior described in the issue.

- behavior with current changes applied (behavior 1; timestamp 00:22):

1. Player far: no entity have a small overlay
2. Player in `Bounds` of `Lookout` near `FakeWall`: glyph but no overlay
3. Player in `Bounds` of `CustomNPC` above `FakeWall`: glyph but no overlay
4. Player in `Bounds` of `CustomNPC` inside `FakeWall`: glyph but no overlay

Uncorrect behavior, overlay missing whenworking for entities with small Bounds (`Lookout`), but not for entites with large Bounds (`CustomNPC`).

- "strengthen the condition in TalkComponent.Update to not set Highlighted to true if Bounds is partially covered by a FakeWall" (in addition to current changes) (behavior 2; 00:35):

1. Player far: no entity have a small overlay
2. Player in `Bounds` of `Lookout` near `FakeWall`: no overlay, no glyph
3. Player in `Bounds` of `CustomNPC` above `FakeWall`: no overlay, no glyph
4. Player in `Bounds` of `CustomNPC` inside `FakeWall`: no overlay, no glyph

As far as I can tell, setting `Highlighted` to `false` prevent both glyph and overlays (small and glyph background) from being displayed.

- "set alpha to 1f if the player manages to collide with Bounds while having it partially covered" (replacing current changes) (behavior 3; timestamp 00:49):

1. Player far: no entity have a small overlay
2. Player in `Bounds` of `Lookout` near `FakeWall`: both overlay and glyph
3. Player in `Bounds` of `CustomNPC` above `FakeWall`: both overlay and glyph
4. Player in `Bounds` of `CustomNPC` inside `FakeWall`: both overlay and glyph

This ignores the `FakeWall` so it will always show the overlays.

I think the desired outcome is:
- Player in `Bounds`: show both overlay and glyph (player can interact)
- Player not in `Bounds`, `Bounds` not covered by `FakeWall`: show small overlay (hint that the player can interact)
- Player not in `Bounds`, `Bounds` covered by `FakeWall`: do not display anything (keep entity hidden), until the player reaches the bounds and can interact
- There should be no circumstances showing a glyph without overlay

This is up to discussion.

With that in mind, the latest commit (behavior 5; timestamp 01:17):
- uses the `Collider` instead of the `Bounds` for `FakeWall` detection, avoiding cases where the Entity is visible but its `Bounds` are still covered
- uses the `Highlighted` property to set the `alpha`, avoiding desyncs between overlay and glyph (caused by alpha difference)
- exhibits the desired outcome

(Behavior 4; timestamp 01:02 uses the `Collider` but ignores the `Highlighted` property)

Map and code used for testing in attached zip: [MapAndCode906.zip](https://github.com/user-attachments/files/20374367/MapAndCode906.zip)

Wartori54:
Wonderful analysis! Keep it up!

I agree that the solution here may be to use the entity's collider itself, since that's representative of its sprite. Note that here you are setting `alpha` to `1`, and never changing it back, which means that once `Highlighted` becomes `true`, thus the `hover/idle` sprite will render permanently (until the entity is reset) when the player is not in `Bounds`. I think that even though this was not a vanilla feature, its a nice to touch, and a good solution if we have to define some behaviour for such case.

One minor nitpick I have though is the direct assignment of `alpha` to `1` in `Update`, removing the fade in animation entirely. If you were to do something like:
```c#
private bool discovered = false;

public override void Update() {
    // ...
    if (Highlighted) discovered = true;
    if (alpha < 1f && (Handler.Entity != null && !Scene.CollideCheck<FakeWall>(Handler.Entity.Collider.Bounds) || discovered)) {
        alpha = Calc.Approach(alpha, 1f, 2f * Engine.DeltaTime);
    }
    base_Update();
}
```
you should be able to maintain the behaviour while having the animation back.
(Note that `Highlighted` can't be assigned to `true` before the first call to `Update` (if no mods interfere), so the check in `Awake` is redundant).

I will approve this PR once this minor change is made.

DashingCat:
I compared the behavior as of latest commit, and the suggested change below:

https://github.com/user-attachments/assets/5dacfe99-00e2-4bec-aedb-46f7f70d0094

- current behavior (timestamp 00:05):

1. When revealing `FakeWall`: the fade-in for the small overlay works properly
2. When interaction is possible while the `FakeWall` hides the entity: there is no fade-in and both overlay and glyph are shown instantly with full alpha

- suggested change (timestamp 00:21):

1. When revealing `FakeWall`: the fade-in for the small overlay works properly
2. When interaction is possible while the `FakeWall` hides the entity: the glyph appears instantly while the overlay fades-in, which looks a bit odd in my opinion

I'm not sure what is the best looking behavior, so I didn't make any change for now.

Please let me now if the current behavior if good as-is, or if I should include the suggested change.

Wartori54:
Missed the part where the stroke color of the glyph is hardcoded to black, LGTM!

Wartori54 (approved):

PR #913, opened 2025-06-03
Fix crash when TalkComponent Handler has no Collider

DashingCat:
Fix a `NullReferenceException` when the `TalkComponent` `Handler` entity has no `Collider` (such as the `OshiroLobbyBell` in "Celestial Resort").

Fallback to vanilla behavior in this case.

Crash introduced in https://github.com/EverestAPI/Everest/pull/906.

### DIFF
--- a/Celeste.Mod.mm/Patches/TalkComponent.cs
+++ b/Celeste.Mod.mm/Patches/TalkComponent.cs
@@ -34,7 +34,7 @@ namespace Celeste {
                 if (Handler.Entity?.Collider != null) {
                     // use Collider if available instead of Position, as Position behaves wrongly
                     // (see https://github.com/EverestAPI/Everest/issues/759)
-                    if (Handler.Entity == null || Scene.CollideCheck<FakeWall>(Handler.Entity.Collider.Bounds)) {
+                    if (Scene.CollideCheck<FakeWall>(Handler.Entity.Collider.Bounds)) {
                         alpha = 0f;
                     }
                 } else {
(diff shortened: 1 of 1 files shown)


maddie480 (approved):

Celeste.Mod.mm/Patches/TalkComponent.cs
```diff
@@ -31,10 +31,17 @@ public patch_TalkComponentUI(TalkComponent handler) : base(handler) {
...
             public override void Awake(Scene scene) {
                 base_Awake(scene);
-                // use Collider instead of Position, as Position behaves wrongly
-                // (see https://github.com/EverestAPI/Everest/issues/759)
-                if (Handler.Entity == null || Scene.CollideCheck<FakeWall>(Handler.Entity.Collider.Bounds)) {
-                    alpha = 0f;
+                if (Handler.Entity?.Collider != null) {
+                    // use Collider if available instead of Position, as Position behaves wrongly
+                    // (see https://github.com/EverestAPI/Everest/issues/759)
+                    if (Handler.Entity == null || Scene.CollideCheck<FakeWall>(Handler.Entity.Collider.Bounds)) {
```
AAA1459: just feel that `Handler.Entity == null` is redundant there and the line 62.

Wartori54 (approved):
100% TAS syncs and doesn't crash with this.

(We should definitely get that automated in CI through headless imo)

Celeste.Mod.mm/Patches/TalkComponent.cs
Kalobi: Agreed, those duplicated checks should be removed.

Kalobi (changes requested):

DashingCat:
Thanks for the feedback, the duplicated checks were removed in the latest commit.

SnipUndercover (approved):

Kalobi (approved):

PR #914, opened 2025-06-04
Run Celeste TASes on pull requests

maddie480:
expect a lot of build failures because I can feel an incoming mess with Actions secrets


maddie480:
... I think I'm expected to push on the Everest repository directly if I want the actions to actually run, so nevermind

PR #916, opened 2025-06-04
Run Celeste TASes on pull requests

maddie480:
Tested on this PR: https://github.com/EverestAPI/Everest/pull/915 because it needed to _target_ a branch that actually had the GitHub Actions yaml in order to run it.

The action:
- Waits for the build to be completed on Azure, polling every minute, filtering on the commit SHA
- Pulls the everest:vanilla Docker image, which is just Ubuntu with a bare-bones Celeste in it, and no Everest (despite the name). We can't pull an encrypted Celeste and decode it with a secret, because pull request pipelines _don't have access to secrets_, so here we go
- Downloads the build artifact from Azure and installs Everest from it
- Downloads CelesteTAS, the TAS files and the Strawberry Jam bundle :tm: from Banana Mirror (they're static, so a mod update breaking TASes won't break the build)
- Runs Celeste with special command line parameters that tell CelesteTAS to run the TAS, write the report and close the game
- Checks if the report says "success"

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/tas-check/1-get-build-url.sh
+++ b/.github/tas-check/1-get-build-url.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+# Waits for Azure to be finished building, and writes the download link to /tmp/everest-pr-tas-check/download-link.txt.
+# Parameter: commit SHA
+
+set -xeo pipefail
+
+get_build_id() {
+	curl 'https://dev.azure.com/EverestAPI/Everest/_apis/build/builds?definitions=3&statusFilter=completed' \
+		| jq -r ".value | map(select(.sourceVersion == \"${1}\")) | .[].id"
+}
+
+BUILD_ID=`get_build_id "$1"`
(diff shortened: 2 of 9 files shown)


maddie480:
This new shiny pipeline epicly failed because of an unrelated crash fixed in #913. 😅

Wartori54:
I think that fetching the tas files from the repo, and checking out a hard coded commit could be better. Same thing with the CelesteTAS version, if it can be fetched from gb or GitHub releases.
That would be good for transparency, and also adds the ability for people without celestemodupdater access to bump and inspect versions more easily.

For the required mods for each run (such as the SJ bundle), I can't think of a better alternative, so it's probably fine as is.

maddie480:
fixed, only the pinned SJ bundle remains on celestemodupdater

Wartori54 (approved):

JaThePlayer (approved):

psyGamer (approved):
Seems reasonable and should cover most obvious issues early. CelesteTAS also provides a utility program for simply executing the sync-check from a config file tho I suppose the features it provides aren't really useful to Everest (optionally installing latest version of an Everest branch, updating mods, better handling for running multiple TAS files, calculating a checksum to avoid repeated work) and therefore makes sense to use a custom script. For CelesteTAS a full sync-check is required before releasing (which takes 20min and therefore caching a nightly check is useful), but since Everest has structured it a bit differently, it seems reasonable.

Alright, enough rambling lmao

PR #919, opened 2025-06-10
Fix `Portraits.xml` reloading in unintended scenarios

SnipUndercover:
Introduced by #890, portraits are reloaded on any scene change. This causes the portraits to reload in unintended scenarios, like reloading the map via <kbd>F5</kbd>, opening the Debug Map, or hitting any of the hardcoded <kbd>F1</kbd>-<kbd>F3</kbd> binds.

This PR fixes this unintended behavior. Tested locally and it works.

### DIFF
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -585,11 +585,19 @@ namespace Celeste {
 
             Scene nextScene = patch_Engine.NextScene;
 
+            // reload the vanilla Portraits.xml when exiting; if a map overrides the Portraits.xml
+            // and doesn't have portrait_madeline defined, the game would crash when trying to load a save file
+            // (since it shows madeline's portrait)
+            //
+            // however we need to pay attention to the new scene, else we'll reload vanilla portraits too soon
+            // and make custom portraits stop working. therefore, we ignore these scenes as they don't actually
+            // exit the map:
+            // - MapEditor - debug map
+            // - AssetReloadHelper - f5 bind
+            // - LevelLoader - vanilla f1-f3 binds + everest softlock prevention
+            // - LevelExit (if Mode is Restart or GoldenBerryRestart) - restarts the level, duh
             if (nextScene is not (Editor.MapEditor or AssetReloadHelper or LevelLoader
                 or patch_LevelExit { Mode: LevelExit.Mode.Restart or LevelExit.Mode.GoldenBerryRestart })) {
-                // reload the vanilla Portraits.xml when exiting
-                // else, if a map overrides the Portraits.xml and doesn't have portrait_madeline defined,
-                // the game will crash when trying to load a save file (since it shows madeline's portrait)
                 GFX.PortraitsSpriteBank = new SpriteBank(GFX.Portraits, Path.Combine("Graphics", "Portraits.xml"));
             }
 
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -583,13 +583,18 @@ namespace Celeste {
         public override void End() {
             orig_End();
 
-            // reload the vanilla Portraits.xml in case it was overridden by a map
-            // else, if the Portraits.xml doesn't have portrait_madeline defined,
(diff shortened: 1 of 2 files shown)


Wartori54:
Since this is a detail that we all missed in the first pr, and requires a separate fix because it was not found until later, I believe it's worthy to be referenced in some way in the added code, specifically why those scenes should be excluded and for what reason.

SnipUndercover:
My first attempt with just `MapEditor` (Debug Map) and `AssetReloadHelper` (<kbd>F5</kbd>) apparently didn't work. I checked and the scene seems to also be changed to `LevelLoader` (hardcoded <kbd>F1</kbd>-<kbd>F3</kbd> binds, the softlock prevention, etc) or `LevelExit` (restart chapter / dying with a golden), and adding those two worked for both me and the person that experienced the issue.
I'll add the comments later today.

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -583,13 +583,26 @@ private static void _GCCollect() {
...
+            //
+            // however we need to pay attention to the new scene, else we'll reload vanilla portraits too soon
+            // and make custom portraits stop working. therefore, we ignore these scenes as they don't actually
+            // exit the map:
+            // - MapEditor - debug map
+            // - AssetReloadHelper - f5 bind
+            // - LevelLoader - vanilla f1-f3 binds + everest softlock prevention
+            // - LevelExit (if Mode is Restart or GoldenBerryRestart) - restarts the level, duh
+            if (nextScene is not (Editor.MapEditor or AssetReloadHelper or LevelLoader
+                or patch_LevelExit { Mode: LevelExit.Mode.Restart or LevelExit.Mode.GoldenBerryRestart })) {
```
Wartori54: `nextScene is not LevelLoader` is equivalent code here, just to handle the ARH since it works wildly differently from other scenes and will call `Scene.End` after the new `LevelLoader` instance has been created. Other scenes ultimately call `new LevelLoader` to switch back to the level, so they're redundant.
Also `Pico8.Emulator` and `OverworldReflectionsFall` do *not* make a `LevelLoader` to go back, so those should be excluded as well, since `Level.End()` is called.
```suggestion
            if (nextScene is not (LevelLoader or Pico8.Emulator or OverworldReflectionsFall)) {
```

This should be enough to handle all everest and vanilla scenes, how to handle modded scenes is still an open question.

Wartori54 (changes requested):

Celeste.Mod.mm/Patches/Level.cs
SnipUndercover: Suggestion applied. I guess when it comes to modded scenes, they should create a new `LevelLoader` scene which automatically reloads the portraits.

Wartori54 (approved):

Celeste.Mod.mm/Patches/LevelExit.cs
```diff
@@ -19,13 +19,19 @@ namespace Celeste {
...
         private XmlElement completeXml;
         private Atlas completeAtlas;
         private bool completeLoaded;
 
         private MapMetaCompleteScreen completeMeta;
 
+        // the mode field is private, but everest needs to access it
+        // however, making it public will break mods which use reflection to access the field...
+        // so we have no other option but to make a public property, oh well
+        public Mode Mode => mode;
```
DashingCat: I believe the changes to `LevelExit.cs` are no longer needed since the `Mode` property introduced is not used as of latest commit.

DashingCat (changes requested):
I tested the changes on my end, I confirm that before applying the changes, the unexpected reloading of portraits appears in <kbd>F1</kbd>-<kbd>F3</kbd> and <kbd>F5</kbd> scenarios, but I couldn't replicate it with the Debug Map.

This PR fixes the unexpected reloading in <kbd>F1</kbd>-<kbd>F3</kbd> and <kbd>F5</kbd> scenarios.

SnipUndercover:
Reverted the `LevelExit` changes.
I kinda wanted to leave them in considering that this seems useful to have public, even though publicizer exists.

microlith57 (approved):

SnipUndercover:
let's go

PR #921, opened 2025-06-12
Fix players can climb up to wall from the right or bottom of Triggered Spikes Original and die

AAA1459:
...from the right side of Triggered Spikes Original and die

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/TriggerSpikesOriginal.cs
+++ b/Celeste.Mod.mm/Mod/Entities/TriggerSpikesOriginal.cs
@@ -156,8 +156,8 @@ namespace Celeste.Mod.Entities {
         }
 
         private bool SideSafeBlockCheck(Player player) {
-            int top = (int) ((player.Top - Top) / 4f);
-            int bottom = (int) ((player.Bottom - Top) / 4f);
+            int top = (int) ((player.Top - Top) / 8f);
+            int bottom = (int) ((player.Bottom - Top) / 8f);
 
             if (bottom < 0 || top >= spikes.Length)
                 return false;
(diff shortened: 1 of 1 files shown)


Kalobi:
I'm not sure whether this is a change we can make at this point. Definitely needs to be looked into further.

AAA1459:
It's almost the same as this one in CH three years ago. https://github.com/CommunalHelper/CommunalHelper/commit/9a7bf7b87abaf0cbf8b62e5caa506da0d60d0dc0
but i agree

Wartori54:
Discord message with a video showcasing the issue: https://discord.com/channels/403698615446536203/1382575587054452818

AAA1459:
New commit is for fix this case.
The top one is Trigger Spikes Original. The bottom one is the vanilla one.
https://github.com/user-attachments/assets/0af5bcee-b8b1-411a-8192-2884769a7056

DashingCat:
I tested the changes on my end and it seems to be working properly as far as I can tell, it feels consistent with vanilla `TriggerSpikes`.

While this PR changes established behavior, I believe it is better to patch the inconsistency to match vanilla safe block checks (which prevent the player from climbing into spikes regardless of which side (left/right) you climb from).

microlith57 (approved):
IMO this fixes a bug (lack of parity with vanilla triggerspikes) so is ok to fix; though would be good to know if anyone relies on this behaviour

JaThePlayer:
I feel like it would be nice a add a `legacyClimbBehavior` (or something similar) EntityData option, which defaults to true, and lonn would default to false for newly placed spikes (and maybe even hide the field entirely?), that way we can be sure we won't break any maps with this

Wartori54 (changes requested):
Going to reiterate on what @JaThePlayer said, not really worth causing a possibly breaking change so let's make an opt-out toggle, with the old behaviour as default.

AAA1459:
> so let's make an opt-out toggle, with the old behaviour as default.

done. Do I need to worry about the config of new option as its everest entity? for loenn.

JaThePlayer (approved):
@Cruor heads up for a new `legacyBehavior` option being added

Wartori54 (approved):

PR #925, opened 2025-06-16
Handle 32-bit Unicode characters in dialog files

maddie480:
C# strings are encoded in UTF-16, and when encountering characters above U+FFFF, you get 2 chars ("surrogate pair") if you just iterate on the string. This PR adds a utility method to turn a string into an array of ints (with a small cache to avoid instantiating them on each frame) and patches a few methods with an IL hook that pretty much replaces `text` with `codePointArray`.

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/CacheHelper.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/CacheHelper.cs
@@ -1,0 +1,52 @@
+using System.Collections.Generic;
+
+namespace Celeste.Mod.Helpers {
+    /**
+     * A helper class that wraps a <code>compute</code> function, in order to keep the
+     * latest computed values in memory (holding up to <code>maxSize</code> values),
+     * to help with performance or with repeated memory allocations.
+     */
+    public abstract class CacheHelper<Key, Value> {
+        private readonly string name;
+        private readonly int maxSize;
+        private readonly Dictionary<Key, Value> cache = new Dictionary<Key, Value>();
+        private readonly LinkedList<Key> cacheOldestToNewest = new LinkedList<Key>();
+
+        public CacheHelper(string name, int maxSize) {
+            this.name = name;
+            this.maxSize = maxSize;
+        }
+
+        public Value GetCached(Key key) {
+            lock (cache) {
+                if (cache.TryGetValue(key, out Value value)) {
+                    LinkedListNode<Key> node = cacheOldestToNewest.Find(key);
+                    cacheOldestToNewest.Remove(node);
+                    cacheOldestToNewest.AddLast(node);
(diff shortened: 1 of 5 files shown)


JaThePlayer:
Any chance string.EnumerateRunes would work here? <https://learn.microsoft.com/en-us/dotnet/api/system.string.enumeraterunes>

maddie480:
Went for a capacity of 1000 for the cache because sitting on the mod options menu with Strawberry Jam enabled uses ~380 strings continuously

microlith57 (approved):
looks reasonable :+1:

JaThePlayer:
Thinking about this, I've known for a long time that a non-insignificant amount of per-frame memory allocations during gameplay happens due to `Emoji.Apply` being called each frame with the Speedrun Timer enabled. Maybe it would be nice to make use of this new cache to also cache the results of `Emoji.Apply`? (As in, a dictionary from the the pre-emojified string to a `ListInt` with emojis already applied)

maddie480:
resurrected the PR, I applied the same caching strat to `Emoji.Apply`

Celeste.Mod.mm/Mod/Helpers/CacheHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public Value GetCached(Key key) {
+            lock (cache) {
+                if (cache.TryGetValue(key, out Value value)) {
+                    LinkedListNode<Key> node = cacheOldestToNewest.Find(key);
+                    cacheOldestToNewest.Remove(node);
+                    cacheOldestToNewest.AddLast(node);
+                    return value;
+                }
```
JaThePlayer: That `cacheOldestToNewest.Find(key);` means that the happy path is doing an O(n) search here, maybe it'd be better to have `cache` be a `Dictionary<Key, (LinkedListNode<Key>, Value)>` to avoid this?

Celeste.Mod.mm/Mod/Helpers/CacheHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+
+                cache.Add(key, value);
+                cacheOldestToNewest.AddLast(key);
+                Logger.Verbose("CacheHelper", $"[{name}] Cached value for \"{key}\" => \"{value}\", cache size: {cache.Count}");
+                return value;
+            }
+        }
+
+        protected abstract Value compute(Key key);
```
JaThePlayer: Should be `Compute`

JaThePlayer (changes requested):
`CacheHelper` will definitely be useful for some code mods!

JaThePlayer (approved):

PR #926, opened 2025-06-17
Fix hardlock when leaving OuiMapSearch

DashingCat:
The game still processes inputs in the `OnTextInput` handler when leaving the `OuiMapSearch`.

This can cause a "hardlock" (game not processing any input), because the handler can disable the input engine.

The "hardlock" can be reproduced by entering some input when transitioning from the `OuiMapSearch` to the `OuiChapterSelect`.

This PR prevents the "hardlock" by removing the handler (which is done by setting `Searching` to `false`) right at the start of the `Leave` method, instead of waiting for the transition to end.

_(originally reported by `tobyaaa` on Discord)_

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs
+++ b/Celeste.Mod.mm/Mod/UI/OuiMapSearch.cs
@@ -463,6 +463,7 @@ namespace Celeste.Mod.UI {
             FromChapterSelect = false;
 
             MInput.Disabled = false;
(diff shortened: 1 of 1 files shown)


Wartori54 (approved):
I would not be surprised if there are some other bugs relating to the text input since those are quite messy in Celeste, but this LGTM.

microlith57 (approved):

PR #927, opened 2025-06-20
Animated Chapter Select Icons

ink-zoey:
This PR allows for chapter icons (`OuiChapterSelectIcon`) to be animated;
The system for this is similar to that of [Animated Parallax](https://github.com/EverestAPI/ModResources/wiki/Helping-Hand-Extra-Features#animated-parallax-stylegrounds), using a .meta.yaml file to provide extra data if needed.

More detail on usage can be found on my (presumably now temporary) gamebanana page for [AnimatedMapIconLib](https://gamebanana.com/tools/20056).

Changes include: 

- Modifying `OuiChapterSelectIcon`'s ctor to allow for the detection of animated icons, this is done by checking if the path to the icon cotains the keyword `"AnimatedMapIcons"`; e.g. `Graphics\Atlases\Gui\areas\MyCoolMod\AnimatedMapIcons`
- Modifying `OuiChapterSelectIcon.Update` to allow for updating both the front and back textures with the supplied animated textures.

Please feel free to inform me of any alternative implementation, or any incorrect practices, and or make any adjustments.

### DIFF
--- a/Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
+++ b/Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
@@ -8,7 +8,6 @@ using Monocle;
 using MonoMod;
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 
@@ -22,8 +21,6 @@ namespace Celeste {
             public string? FrontFrames { get; set; }
             public string? BackFrames { get; set; }
         }
-
-        private const string DetectionPath = "AnimatedMapIcons";
 
         private bool isAnimated;
 
@@ -62,20 +59,25 @@ namespace Celeste {
         public void ctor(int area, MTexture front, MTexture back) {
             orig_ctor(area, front, back);
 
-            // Don't bother if its not in the path, perhaps replace with an alternative check for a .meta.yaml file?
-            if (!front.AtlasPath.Contains(DetectionPath))
-                return;
-
-            isAnimated = true;
-
             // Remove the frame suffex.
             string frontPath = Regex.Replace(front.AtlasPath, "\\d+$", string.Empty);
             string backPath = frontPath + "_back";
 
-            FrontTextures = GFX.Gui.GetAtlasSubtextures(frontPath);
+            bool hasFront = ((patch_Atlas) GFX.Gui).HasAtlasSubtextures(frontPath);
+            bool hasBack = ((patch_Atlas) GFX.Gui).HasAtlasSubtextures(backPath);
+
+            if (hasFront || hasBack)
+                isAnimated = true;
+            else
+                return;
+
+            if (hasFront)
+                FrontTextures = GFX.Gui.GetAtlasSubtextures(frontPath);
+            else
+                FrontTextures = new List<MTexture>() { front };
 
             // Only grab the back textures if applicable. otherwise default to the front textures.
-            if (((patch_Atlas)GFX.Gui).HasAtlasSubtextures(backPath))
+            if (hasBack)
                 BackTextures = GFX.Gui.GetAtlasSubtextures(backPath);
             else if (GFX.Gui.Has(backPath))
                 BackTextures = new List<MTexture>() { GFX.Gui[backPath] };
@@ -91,14 +93,8 @@ namespace Celeste {
             FrontFPS = 12f;
             BackFPS = 12f;
 
-            // Read in data from a .meta.yaml file with the same name as the map icon (excluding the frame suffex).
-            if (!Everest.Content.Map.TryGetValue("Graphics/Atlases/Gui/" + frontPath + ".meta", out ModAsset value) || value.Type != typeof(AssetTypeYaml))
+            if (!(Everest.Content.Get<AssetTypeYaml>("Graphics/Atlases/Gui/" + frontPath + ".meta")?.TryDeserialize(out AnimatedMapIconMeta meta) ?? false))
                 return;
-
-            AnimatedMapIconMeta meta;
-
-            using (TextReader input = new StreamReader(value.Stream))
-                meta = YamlHelper.Deserializer.Deserialize<AnimatedMapIconMeta>(input);
 
             // Back sprite should always default to the front sprite.
             FrontFPS = meta.FrontFPS ?? FrontFPS;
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
```diff
@@ -18,6 +57,87 @@ public patch_OuiChapterSelectIcon(int area, MTexture front, MTexture back)
...
         }
 
+        public extern void orig_ctor(int area, MTexture front, MTexture back);
+        [MonoModConstructor]
+        public void ctor(int area, MTexture front, MTexture back) {
+            orig_ctor(area, front, back);
+
+            // Don't bother if its not in the path, perhaps replace with an alternative check for a .meta.yaml file?
+            if (!front.AtlasPath.Contains(DetectionPath))
+                return;
```
JaThePlayer: Is this additional subdirectory actually necessary? I'd much rather not have it necessary so you can just put animation frames into an existing directory and have it just work.

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
```diff
@@ -18,6 +57,87 @@ public patch_OuiChapterSelectIcon(int area, MTexture front, MTexture back)
...
+
+            isAnimated = true;
+
+            // Remove the frame suffex.
+            string frontPath = Regex.Replace(front.AtlasPath, "\\d+$", string.Empty);
+            string backPath = frontPath + "_back";
+
+            FrontTextures = GFX.Gui.GetAtlasSubtextures(frontPath);
```
JaThePlayer: So probably do the same thing here as with BackTextures and I think this code will work fine for un-animated icons as well.

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
```diff
@@ -18,6 +57,87 @@ public patch_OuiChapterSelectIcon(int area, MTexture front, MTexture back)
...
+            BackFPS = 12f;
+
+            // Read in data from a .meta.yaml file with the same name as the map icon (excluding the frame suffex).
+            if (!Everest.Content.Map.TryGetValue("Graphics/Atlases/Gui/" + frontPath + ".meta", out ModAsset value) || value.Type != typeof(AssetTypeYaml))
+                return;
+
+            AnimatedMapIconMeta meta;
+
+            using (TextReader input = new StreamReader(value.Stream))
+                meta = YamlHelper.Deserializer.Deserialize<AnimatedMapIconMeta>(input);
```
JaThePlayer: Probably worth using the built-in everest stuff for parsing yaml content:
```cs
if (!Everest.Content.Get<AssetTypeYaml>(..path..)?.TryDeserialize<AnimatedMapIconMeta>(out var meta) ?? false)
    return;
```

JaThePlayer (changes requested):

ink-zoey:
All requested changes have been made, 
I'm also allowing both the back and front icons to be animated independently.

(Perhaps I may want to add an `isAnimated` bool for both the front and back? Not sure if anyone is going to be just animating the back or something.)

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
```diff
@@ -1,11 +1,47 @@
 ﻿#pragma warning disable CS0649 // Field is never assigned to, and will always have its default value
+#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
+#pragma warning disable CS8632 // The annotation for nullable reference types should only be used in code within a #nullable annotations context
+#pragma warning disable SYSLIB1045 // Use GeneratedRegexAttribute to generate the regular expression implementation at compile time.
```
JaThePlayer, line 4: Well, we *should* be doing that

JaThePlayer (approved):
lgtm, switching over to generated regex for slightly better perf would be nice but its not the end of the world here.

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
```diff
@@ -18,6 +54,86 @@ public patch_OuiChapterSelectIcon(int area, MTexture front, MTexture back)
...
+                BackTextures.Count == FrontTextures.Count) // Make sure to see if the counts line up.
+                BackOrder = FrontOrder;
+        }
+
+        public extern void orig_Update();
+        public override void Update() {
+            orig_Update();
+
+            // Just return if hidden or if its not animated.
+            if (hidden || SaveData.Instance == null || !isAnimated)
```
Wartori54, line 119: Could you please elaborate on why is the `SaveData.Instance` check necessary here?

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
```diff
@@ -18,6 +54,86 @@ public patch_OuiChapterSelectIcon(int area, MTexture front, MTexture back)
...
+                return;
+
+            FrontFrameTimer -= Engine.DeltaTime;
+            if (FrontFrameTimer < 0f) {
+                FrontFrameTimer += 1f / FrontFPS;
```
Wartori54, line 124: This is a huge nitpick but this should either be an assignment or the `if` should become a `while`.
Same thing with the almost identical code down below.

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
ink-zoey: I noticed that `SaveData.Instance` was checked in `orig_Update`, and I assumed it was an extra safety precaution; I'm guessing that by this comment you're saying this check is irrelevant? (And should never really come up?).

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
ink-zoey: I can go in and make this small change if you'd like, I'm not sure how much it affects visuals though.

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
Wartori54: I see, you should specify that in the comment then.
I assume the vanilla game does that because it can actually happen, so it makes sense for you to also skip updates if that is null.

Celeste.Mod.mm/Patches/OuiChapterSelectIcon.cs
Wartori54: It does not affect visuals differently at all in the `while` case and makes swaps always happen at a fixed number of updates for assignment. It is nothing to ever worry about in celeste, but it could cause weird behaviour if `DeltaTime` were variable.

Wartori54 (approved):

PR #928, opened 2025-06-20
Make OnEndOfFrame thread safe

wuke32767:
Previously: Any `OnEndOfFrame` added during `OnEndOfFrame` invocation will be discarded.

Changes: They will be invoked in next frame.

Note: `OnEndOfFrame` can add new `OnEndOfFrame`s. Nobody should be doing this (otherwise they will notice it), but if it happens, this pr causes desync.

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/Scene.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/Scene.cs
@@ -3,6 +3,8 @@ using MonoMod;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


JaThePlayer (approved):
I wouldn't really say it's about thread-safety as its more about being able to set OnEndOfFrame on the same thread, but I'm for this either way. (With that said, Interlocked here isn't really necessary, but the code's cleaner than the non-Interlocked way, so it's fine imo)

wuke32767:
It's initially for thread safety, but the listed changes are more significant.

Wartori54 (approved):

PR #934, opened 2025-07-02
Ensure no dead players get loaded as part of a `Level.LoadOverride`

aonkeeper4:
This PR ensures that `Level.LoadOverride`s will not get used if their `NextLoadedPlayer` is dead. This can happen due to deaths caused by the state change in `AssetReloadHelper.ReloadLevel`: https://github.com/EverestAPI/Everest/blob/ecad4be5e89defe614118b8451b47f89cd43d800/Celeste.Mod.mm/Mod/Helpers/AssetReloadHelper.cs#L234 (for example, a death caused by `Player.StarFlyReturnToNormalHitbox` when reloading in feather state), which undesirably softlocks the player into having to reload the level again.


microlith57 (approved):
:+1:

JaThePlayer (approved):

PR #943, opened 2025-07-20
Add option to toggle music on/off

DashingCat:
This PR adds an option toggle the in-game music on or off, as well as a key bind for the toggle.


Kalobi:
The extra setting feels like menu clutter when there's already a way to turn music off in the vanilla settings, and tbh I don't see why a keybind for toggling music should be an Everest feature.

SnipUndercover:
I second this, this feels like it belongs to a mod more than Everest.

Wartori54:
I third this, there's no real reason for this to be in Everest, could be in any QOL mod.

PR #947, opened 2025-07-26
Fix: /tp and /reloadmap endpoints in debugrc not url decoding

Pandaptable:
This changes `/tp` and `/reloadmap` in debugrc to url decode the SID
Previously, if an SID contained a space, you could not use the endpoint for a given map.

Recreation: Download any map with an SID that has a space ([example](https://gamebanana.com/mods/602839)) and try to teleport or reload the map with the SID. It will return %20 instead of a proper space, leading it to fail.
(for the given example map it would be http://localhost:32270/tp?tp=Mochi/Puppygirl%20Adventures/Puppygirl%20Adventures&side=A&level=a-1&x=27&y=440)

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.DebugRC.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.DebugRC.cs
@@ -374,7 +374,7 @@ header {
                     Handle = c => {
                         NameValueCollection data = ParseQueryString(c.Request.RawUrl);
 
-                        string sid = data["area"];
+                        string sid =  WebUtility.UrlDecode(data["area"]);
                         if (string.IsNullOrEmpty(sid))
                             sid = (Engine.Scene as Level)?.Session.Area.GetSID();
                         if (string.IsNullOrEmpty(sid)) {
@@ -475,7 +475,7 @@ header {
                             return;
                         }
 
-                        string sid = data["area"];
+                        string sid =  WebUtility.UrlDecode(data["area"]);
                         if (string.IsNullOrEmpty(sid))
                             sid = session.Area.GetSID();
                         if (string.IsNullOrEmpty(sid)) {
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/Everest/Everest.DebugRC.cs
```diff
@@ -374,7 +374,7 @@ public static void Write(HttpListenerContext c, string str) {
                     Handle = c => {
                         NameValueCollection data = ParseQueryString(c.Request.RawUrl);
 
-                        string sid = data["area"];
+                        string sid =  WebUtility.UrlDecode(data["area"]);
```
microlith57: nitpick: double space could be a single space
```suggestion
                        string sid = WebUtility.UrlDecode(data["area"]);
```

Celeste.Mod.mm/Mod/Everest/Everest.DebugRC.cs
```diff
@@ -475,7 +475,7 @@ public static void Write(HttpListenerContext c, string str) {
                             return;
                         }
 
-                        string sid = data["area"];
+                        string sid =  WebUtility.UrlDecode(data["area"]);
```
microlith57:
```suggestion
                        string sid = WebUtility.UrlDecode(data["area"]);
```

microlith57 (approved):
approved with some tiny nitpicks :3

SnipUndercover (commented):
LGTM, will approve after those nitpicks are addressed.

Pandaptable:
what not using your normal editor does to a mf </3 (i normally highlight all whitespace so didn't notice when using rider) 👍

SnipUndercover (approved):

SnipUndercover:
> i normally highlight all whitespace so didn't notice when using rider

You're in luck.
image

PR #953, opened 2025-07-31
provides a method to safely enumerate SwapImmediately

wuke32767:
not fully tested, but looks good to me

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/SwapImmediatelyExtension.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/SwapImmediatelyExtension.cs
@@ -32,24 +32,46 @@ namespace Celeste.Mod {
         ///   }
         /// </code>
         /// </example>
-        public static IEnumerator SafeEnumerate(this IEnumerator self) {
-            Stack<IEnumerator> enums = new();
-            enums.Push(self);
+        public static Flattened SafeEnumerate(this IEnumerator self) {
+            if(self is Flattened f) {
+                return f;
+            }
+            return new(self);
+        }
 
-            while (enums.Count > 0) {
-                IEnumerator cur = enums.Peek();
+        public struct Flattened : IEnumerator {
+            private readonly Stack<IEnumerator> enums = new();
+            private object current = null;
 
-                if (cur.MoveNext()) {
-                    object obj = cur.Current;
+            public readonly object Current => current;
 
-                    if (obj is SwapImmediately swap) {
-                        enums.Push(swap.Inner);
+            public Flattened(IEnumerator from) {
+                enums.Push(from);
+            }
+
+            public bool MoveNext() {
+                while (enums.Count > 0) {
+                    IEnumerator cur = enums.Peek();
+
+                    if (cur.MoveNext()) {
+                        object obj = cur.Current;
+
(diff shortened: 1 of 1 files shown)


maddie480 (approved):

Wartori54:
I believe some sort of documentation for this method would be really beneficial, especially the intended usage of the method as well as its motivation.

SnipUndercover (changes requested):
I would want to see some documentation here, like what does this accomplish and what does it benefit.

microlith57 (approved):

Celeste.Mod.mm/Mod/Helpers/SwapImmediatelyExtension.cs
```diff
@@ -0,0 +1,41 @@
...
+    public static class SwapImmediatelyExtension {
+        /// <summary>
+        ///   when you want to enumerate a coroutine ienumerator,
+        ///   always keep in mind that the original sequence may have been wrapped into <see cref="SwapImmediately"/>.
+        ///   this method helps you to handle it correctly like <see cref="Coroutine"/>.<br/>
+        ///   also known as, flattening the given sequence.
+        /// </summary>
+        /// <returns>
+        ///   A new enumerator, but all <see cref="SwapImmediately"/> is safely handled.
+        /// </returns>
```
JaThePlayer:
```suggestion
        /// <summary>
        /// Flattens all <see cref="SwapImmediately"/> returned by the source IEnumerator, 
        /// that is, when the source IEnumerator would yield a <see cref="SwapImmediately"/>,
        /// elements from the wrapped IEnumerator get yielded one-by-one instead. 
        /// (Behaviour mimicks <see cref="Coroutine"/>'s handling of <see cref="SwapImmediately"/>.)
        /// When writing hooks on Coroutine-style methods, 
        /// this method is needed if you want to access values returned from the orig Coroutine,
        /// or rely on the source Coroutine yielding a specific amount of values.
        /// </summary>
        /// <returns>
        ///   A new enumerator, with all <see cref="SwapImmediately"/> flattened.
        /// </returns>
```

This should be a bit better imo, assuming I understand the reasoning behind this method existing.

JaThePlayer (changes requested):
The name `SafeEnumerate` implies some kind of exception handling imo, making this name very misleading, especially as it will appear in an IDE on any enumerator, not just ones used for coroutines. Maybe `FlattenSwapImmediately`?

The documentation also doesn't really explain what "handle it correctly" actually means, and when this method should be used (Is there potential for desyncs if you don't?)

SnipUndercover (commented):
I agree with Ja's suggestions. The only thing I'd add onto that is more `<see cref="..." />`s in XML docs.

JaThePlayer:
Side-note: Once this PR gets merged, documentation at <https://github.com/EverestAPI/Resources/wiki/Making-Code-Mods#on-hooks> should be adjusted to mention this method.

SnipUndercover (approved):

JaThePlayer:
Considering #955 exists, I assume this PR should not be merged anymore, considering that PR would introduce potentially binary-breaking changes (changing return type) to this method?

wuke32767:
you are right. i'll also apply the binary-breaking changes in this pr.

PR #955, opened 2025-08-02
Rework `SwapImmediately` handling

wuke32767:
Although #953 looks to be good, but i still noticed somethong:

What if code modders don't use it correctly?

According to https://github.com/EverestAPI/Resources/wiki/Making-Code-Mods#on-hooks, when hooking a coroutine with `SwapImmediately`, sometimes the method suffix will not run. However, #953 will ALWAYS run them. While nobody should be doing this, this could cause problems in the future:

- B flattens the sequence, just because they can.
- A adds a method suffix.
  - the suffix should not run, but...
- A runs before B.
  - their sequence will be flattened by B. so it runs.
- A is released with the bug.
- Some users don't enable B. For no reason.
- A can't work.

Actually there's no solution for this. so
This pr changes `Coroutine` to always run suffixes.

still not tested, but looks good to me.

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/SwapImmediatelyExtension.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/SwapImmediatelyExtension.cs
@@ -32,24 +32,46 @@ namespace Celeste.Mod {
         ///   }
         /// </code>
         /// </example>
-        public static IEnumerator SafeEnumerate(this IEnumerator self) {
-            Stack<IEnumerator> enums = new();
-            enums.Push(self);
+        public static Flattened SafeEnumerate(this IEnumerator self) {
+            if(self is Flattened f) {
+                return f;
+            }
+            return new(self);
+        }
 
-            while (enums.Count > 0) {
-                IEnumerator cur = enums.Peek();
+        public struct Flattened : IEnumerator {
+            private readonly Stack<IEnumerator> enums = new();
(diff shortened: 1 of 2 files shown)


Wartori54:
Looks good, I'd like to test it before giving my approval, which I will do at some other time.

Celeste.Mod.mm/Mod/Helpers/SwapImmediatelyExtension.cs
```diff
@@ -0,0 +1,78 @@
...
+            return new(self);
+        }
+
+        public struct Flattened : IEnumerator {
+            private readonly Stack<IEnumerator> enums = new();
+            private object current = null;
+
+            public readonly object Current => current;
+
+            public Flattened(IEnumerator from) {
```
microlith57, line 48: **nitpick:** maybe make this ctor `internal`? the only thing it lets you do that `SafeEnumerate` doesn't is `new Flattened(new Flattened(...))` (which seems pointless)
i don't see a huge downside to having this `public`, except that people might get confused

Wartori54 (approved):
Self-reminder to update the docs of `SwapImmediately` at the wiki once this is merged.
LGTM.

microlith57 (approved):
oops, looked at this a while ago but i guess i didn't actually press approve. lgtm :+1:

PR #956, opened 2025-08-03
Update Steamworks.NET

### LINKED ISSUE #860: Bundle an updated steamapi/Steamworks.NET

Yes

### Describe your request

I'm working on a mod to add Game Recording integration to Celeste, to make it easy to find and clip successful attempt at screens. However, this requires updating the steam_api and Steamworks.NET DLLs/so/dylibs to a recent build, as it relies on functionality that was added very recently.

Currently, to make it all work, I just take the latest version of the Steamworks Standalone release, copy paste Steamworks.NET.dll to the Celeste folder, and steam_api64.dll in the lib64-win-x64 folder. But this makes mod distribution rather complicated, as it requires manual steps for the users.

Would it be possible to update the steam_api64.dll and Steamworks.NET used by Everest to the latest version?

### Additional context

_No response_

roblabla:
This allows mods to make use of the newer features provided by Steamworks.NET. This is a requirement of my mod providing [Steam Game Recording Integration](https://github.com/roblabla/Celeste-SteamRecordingIntegration).

~~Needs https://github.com/EverestAPI/Everest-libs/pull/5~~ This PR is merged!

We're not updating to the _very_ latest Steamworks.NET as it contains a breaking change that breaks Celeste (same as https://github.com/rlabrecque/Steamworks.NET/issues/709). Fixing that would be relatively easy to do (just needs to nop out the call to `SteamUserStats.RequestCurrentStats()` in celeste's `Stats.MakeRequest`), but given my use-case doesn't need the newer functionality offered by the very latest Steamworks, I opted for the path of least resistance.

The largest part of the change is in the MiniInstaller. It does two things:

1. It copies the actual implementation dll of Steamworks.NET from lib-ext to the correct location to be picked up by Celeste.
2. It ignores Steamworks.NET in the `ConvertToNETCore` call. This is because otherwise, it would overwrite the previously copied Steamworks.NET with the version found in the orig folder, breaking everything.

I verified it works on Windows 64-bit and macOS (both Steam builds).

Fixes #860


roblabla:
Well, this is annoying. Because I added Steamworks.NET as a PackageReference, a reference assembly gets added to the output. This is not a huge problem for the everest build, as it gets overwritten with the correct assembly. What kinda sucks though is that this reference assembly will overwrite the original Steamworks.NET.dll, which means it will end up in the orig folder, breaking the unmodded game build.

I looked for a way to massage `dotnet publish` into not providing that reference assembly, but couldn't find one (to be clear, I may have overlooked something - I don't do a whole lot of C#).

I can see two ways to fix this:

1. Don't use nuget for the Steamworks.NET dependency. I could instead add Steamworks.NET.dll in `libs-ext` (along with steam_api). This way, it'd get copied by the MiniInstaller as necessary. This would require updating lib-stripped as well.
2. Provide the original Steamworks.NET in `libs-ext/lib-vanilla`. This way, the installer would overwrite the Reference Assembly with the original file.

For now, I think I'll try pursuing the first option.

roblabla:
All fixed. Steamworks.NET.dll is now copied from lib-ext, so there's no more shenanigans going on with the orig directory.

roblabla:
The everest-libs PR got merged, this is read for review.

SnipUndercover:
I'll rerun the pipeline since it errored out.

SnipUndercover:
@maddie480 It seems like the tas checks don't want to run? The action times out on waiting for the build to complete, even though it finished in 6 min.

Wartori54 (approved):
Changes are good, but not too happy since if Steamworks.NET.dll was platform and arch independent all of these would not be required at all. 
Patching Steamworks.NET.dll to be like that is something not really in the scope of this PR though. (But it should be looked into in the future.)

maddie480:
> @maddie480 It seems like the tas checks don't want to run? The action times out on waiting for the build to complete, even though it finished in 6 min.

it happens sometimes, Azure and GitHub Actions build "different merge commits", neither of which actually exist. GitHub builds https://github.com/EverestAPI/Everest/commit/ecc2f4d04b1fc2f13dc6a3a1339c98a58b382261, Azure builds https://github.com/EverestAPI/Everest/commit/11cb833978f9574ee0fb52b2d1d01d587bf8541c, because ????

maddie480:
(pushed an empty commit then force pushed to remove it, now both GitHub Actions and Azure Pipelines are working on commit https://github.com/EverestAPI/Everest/commit/a63dd5d1cc1b815d042f7a552c290e9476997304 so it should be fine)

SnipUndercover (approved):

SnipUndercover:
> (pushed an empty commit then force pushed to remove it, now both GitHub Actions and Azure Pipelines are working on commit https://github.com/EverestAPI/Everest/commit/a63dd5d1cc1b815d042f7a552c290e9476997304 so it should be fine)
> 
> 

Well for some reason the checks are still showing that they're canceled. Either GitHub is high or something's wrong, but I don't think this'll affect TASes by any means.
In any case, Let's Gamble, Try Merging.

maddie480:
The checks on https://github.com/EverestAPI/Everest/commit/e0d57e19404b37f1be1f00cef527379cbf6236e0 pass if you go check the commit directly rather than the "Checks" tab, so I vote for "GitHub is high"

PR #958, opened 2025-08-04
Add clearer errors on access failures

SnipUndercover:
Throws an error if the `Celeste.exe` or `Celeste.dll` have are marked as read-only and throws an error if so with instructions to remove the read-only attribute (Windows-only) or ensure the user has write permissions to the file.

The "Celeste not read-writeable - waiting" message now has a timeout of 1 minute instead of waiting indefinitely, throwing an exception if `Celeste.exe` or `Celeste.dll` are not accessible (including the source exception).

### DIFF
--- a/MiniInstaller/InGameUpdaterHelper.cs
+++ b/MiniInstaller/InGameUpdaterHelper.cs
@@ -59,10 +59,12 @@ public static class InGameUpdaterHelper {
         // let's wait for 1 minute
         const int WaitTimeSeconds = 60;
         const int MaxRetryCount = 12;
+
+        Exception error;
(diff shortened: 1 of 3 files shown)


microlith57 (approved):
nice :+1:

SnipUndercover:
A warning to readers: if you're thinking about making a quick fix on phone... don't!

this took me 3 attempts, please work

edit: let's go.

Wartori54 (approved):

SnipUndercover:
Wartori is evil for putting the "should be squashed" tag an hour before the pr was ready for merge :wholestare:

PR #966, opened 2025-08-14
Fix crash when trying to apply emoji to null

maddie480:
Apparently, that happens sometimes. And Emoji.Apply did handle it (by returning null directly). But the dictionary of the new Emoji.Apply cache doesn't like having null as a key...

see https://discord.com/channels/403698615446536203/1405650694647713882 / regression introduced in #925

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Emoji.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Emoji.cs
@@ -247,6 +247,7 @@ namespace Celeste.Mod {
         /// <param name="text"></param>
         /// <returns></returns>
         public static string Apply(string text) {
+            if (text == null) return null;
             return cachedApplies.GetCached(text);
         }
     }
--- a/Celeste.Mod.mm/Patches/Monocle/PixelFontSize.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/PixelFontSize.cs
@@ -116,6 +116,13 @@ namespace MonoMod {
     }
 
     static partial class MonoModRules {
+        private static void emitCallToCodePointList(ILCursor cursor, VariableDefinition v_listOfCodePoints) {
+            // List<int> listOfCodePoints = UnicodeStringHelper.ToCodePointList(text);
+            cursor.Emit(OpCodes.Ldarg_1);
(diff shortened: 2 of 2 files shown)


Wartori54 (dismissed):

DashingCat (changes requested):
There is a related crash which I believe should be fixed as well, one way to fix it would be to add a `null` check here:
https://github.com/EverestAPI/Everest/blob/92b6d88aa73e38970f4813b7e15c36d91dd0a565/Celeste.Mod.mm/Mod/Helpers/UnicodeStringHelper.cs#L26-L28
```diff
public static ListInt ToCodePointList(string text) {
+	if (text == null) return new ListInt { Elements = System.Array.Empty<int>() };
	return cacher.GetCached(text);
}
```

Because `Emoji.Apply` now returns `null`, the `null` should also be handled later by either `PixelFontSize.Measure` or `UnicodeStringHelper.ToCodePointList`; `cacher.GetCached(text)` in `ToCodePointList` does a dictionnary lookup with a `null` key and throws a `ArgumentNullException` (relevant snippets below).

https://github.com/EverestAPI/Everest/blob/92b6d88aa73e38970f4813b7e15c36d91dd0a565/Celeste.Mod.mm/Patches/Monocle/PixelFontSize.cs#L17-L21

https://github.com/EverestAPI/Everest/blob/92b6d88aa73e38970f4813b7e15c36d91dd0a565/Celeste.Mod.mm/Mod/Helpers/CacheHelper.cs#L25-L27

maddie480:
If this decompilation is to be believed, the orig method for `PixelFontSize.Measure` does handle nulls, and stops the execution before reaching the part where I patch `UnicodeHelper` in: https://github.com/BloodLantern/Celeste/blob/master/Monocle%2FPixelFontSize.cs#L74-L75

maddie480:
as discussed on Discord, my defense is invalid because I have an IL hook that changes the null check I was pointing to :speak_no_evil: 

if I inject my code after the null checks from the vanilla method, nulls won't reach my methods, so I won't need to handle nulls

maddie480:
done:

image

microlith57 (approved):
:+1:

JaThePlayer (approved):

DashingCat (approved):

PR #972, opened 2025-08-25
Deprecate TimeRate for Mods who should use TimeRateModifiers

TheDavSmasher:
TimeRate (and alongside it, TimeRateB) are a little confusing as to why there's two of them, so this PR adds an xmldoc to both of them.

However, we realized that mods should be using TimeRateModifier components instead of accessing and modifying the vanilla TimeRate and/or TimeRateB, therefore both of them were given the Obsolete attribute.

Additionally, a new field called EffectiveTimeRate was made, set to the value of `TimeRate * TimeRateB * GetTimeRateComponentMultiplier(this.scene)` which is then used to `DeltaTime = EffectiveTimeRate * RawDeltaTime`.

This is to encourage mods to use the Component dedicated to preventing conflicts while also providing the information useful to them. It still keeps both original fields untouched but encourages mods to leave them as they are unless they know what they're doing.


SnipUndercover:
Pushed a fixed version of the patch that doesn't throw a runtime error and follows code standards. Discussed in `#code_modding`.

SnipUndercover:
Updated the docs and obsoletion notices. I confirmed whether it's fine to push to the fork with the PR author.

SnipUndercover (dismissed):

Celeste.Mod.mm/Patches/Monocle/Engine.cs
```diff
@@ -149,38 +184,58 @@ public static Viewport Viewport {
...
     static partial class MonoModRules {
         public static void PatchEngineUpdate(ILContext context, CustomAttribute attrib) {
             TypeDefinition t_Engine = context.Method.DeclaringType;
             FieldReference f_scene = t_Engine.FindField("scene");
+            FieldReference sf_EffectiveTimeRate = t_Engine.FindField("EffectiveTimeRate");
             MethodReference m_GetTimeRateComponentMultiplier = t_Engine.FindMethod("GetTimeRateComponentMultiplier");
-            
+            MethodReference m_get_RawDeltaTime = t_Engine.FindMethod("get_RawDeltaTime");
+
+            // remove loading RawDeltaTime and multiplying it with TimeRate
```
Kalobi: Floating point multiplication is not necessarily associative. (RawDeltaTime * TimeRate) * TimeRateB might produce a slightly different result to (TimeRate * TimeRateB) * RawDeltaTime (same with the component modifier of course), which could cause a change of behavior in extreme cases and should be avoided.
Also, if we're making EffectiveTimeRate a field, that will cause it to be desynced from the actual time rate values (there is code that checks current TimeRate, and replicating that with EffectiveTimeRate would behave differently if the time rate changes within the frame because the check would be locked in at the start of the frame. On the other hand, GetTimeRateComponentMultiplier is not the cheapest operation, so this is probably the best approach. I'd slightly expand the documentation for the field with something like "multiplying ... together *at the beginning of the frame*." though.

Celeste.Mod.mm/Patches/Monocle/Engine.cs
```diff
@@ -43,15 +41,16 @@ public patch_Engine(int width, int height, int windowWidth, int windowHeight, st
         private static readonly FieldInfo f_Game_hasInitialized = typeof(Game).GetField("hasInitialized", BindingFlags.NonPublic | BindingFlags.Instance);
         private static readonly MethodInfo m_Game_RunLoop = typeof(Game).GetMethod("RunLoop", BindingFlags.NonPublic | BindingFlags.Instance);
         private static readonly MethodInfo m_Game_AfterLoop = typeof(Game).GetMethod("AfterLoop", BindingFlags.NonPublic | BindingFlags.Instance);
-        
+
```
Kalobi: Please avoid committing autoformatting changes in otherwise untouched code in a non-cleanup PR. This makes git blames confusing. This occurs throughout the file. We can consider doing a general formatting cleanup at some point, although I don't think we really have consistent coding guidelines.

Kalobi (dismissed):

Celeste.Mod.mm/Patches/Monocle/Engine.cs
SnipUndercover: Interesting catch. Maybe something like this would suffice more:
```cs
// before:
EffectiveTimeRate = TimeRate * TimeRateB * GetTimeRateComponentMultiplier();
DeltaTime = RawDeltaTime * EffectiveTimeRate;

// after:
float componentTimeRate = GetTimeRateComponentMultiplier();
EffectiveTimeRate = TimeRate * TimeRateB * componentTimeRate;
DeltaTime = RawDeltaTime * TimeRate * TimeRateB * componentTimeRate;
```
This would avoid potential inaccuracies since the order of the operations is the same as it was before. It's not like multiplications are particularly expensive.
I double checked whether the JIT won't try to reuse the intermediate `TimeRate * TimeRateB * componentTimeRate` value in this [Compiler Explorer](https://godbolt.org/z/3Ev8d1hME) snippet - it does not.

Celeste.Mod.mm/Patches/Monocle/Engine.cs
TheDavSmasher: Sorry, that was caused by my IDE most likely, though I'm not sure what to change.

Kalobi: Just go through the diff and revert every change that wasn't done by you

TheDavSmasher: Should be done unless you have more to say @Kalobi

Celeste.Mod.mm/Patches/Monocle/Engine.cs
TheDavSmasher: @Kalobi @SnipUndercover is this still in need of changes?

SnipUndercover: Yeah, the docs and emitted code would need to be updated as described above.

TheDavSmasher: Oh I thought the changes had already been made by you, I can do them and mention

SnipUndercover (approved):

microlith57 (approved):

SnipUndercover:
The PR is waiting for Kalobi's review, but given that the changes were addressed and the order of operations is now the same as in vanilla, I think it's safe to dismiss the stale review.

SnipUndercover:
I'll give this a day or two before merging just in case.

maddie480 (approved):
Well, another approval also pushes back the last call window

PR #974, opened 2025-08-31
Implement culling for LavaRect

JaThePlayer:
Also culls LavaRect.Update, as the changes it does are purely for visuals. Frost Helper's custom lava rects do the same thing and noone complained, so it should be fine.

### DIFF
--- a/Celeste.Mod.mm/Patches/LavaRect.cs
+++ b/Celeste.Mod.mm/Patches/LavaRect.cs
@@ -1,3 +1,4 @@
+using Celeste.Mod.Helpers;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using MonoMod;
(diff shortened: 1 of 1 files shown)


microlith57 (approved):

Wartori54:
The `Update` method contains a `Calc.Random` call: 
image
This could mess up the RNG in TASes if the updates are no-op'd when it is not visible. I have not investigated how impactful this is since calls to `Calc.Random` that affect gameplay in a substantial way are rare. (And CI TASes are not affected by this it seems.)

Kalobi:
Since Calc.Random is not seeded during normal gameplay and most TAS-relevant RNG occurs during level load where it is seeded, I don't think changing the number of Random calls in an Update method is a concern. @psyGamer Do you agree?

psyGamer:
Hmm I guess it's _somewhat_ a concern? 
During update it's not seeded but it is deterministic (when you leave out `Render` which CelesteTAS does), _but_ relying on `Calc.Random` over long periods of time after being seeded is a PITA while TASing anyway, so kinda a "purist" thing anyway.
CelesteTAS provides dedicated seeds which are a lot easier to work with and won't break by a change like this.
Lastly I doubt there exists any TAS which relies on `Calc.Random` **inside a room using lava**.
So this issue is mostly theoretical and not actually practical, so I would say the change is fine.

psyGamer (approved):

SnipUndercover:

PR #979, opened 2025-09-03
Add NoWarn attribute to DotNetZip Dependency

TheDavSmasher:
For a while I've found it annoying that Visual Studio tells me that the dependency is Vulnerable and since the dependency is currently deprecated and not updated, it'll remain that way, and this annotation is strictrly an IDE-only change to fix this minor annoyance.

### DIFF
--- a/Celeste.Mod.mm/Celeste.Mod.mm.csproj
+++ b/Celeste.Mod.mm/Celeste.Mod.mm.csproj
@@ -46,7 +46,13 @@
     <ProjectReference Include="..\external\NLua\build\net6.0\NLua.net6.0.csproj" />
     <PackageReference Include="MAB.DotIgnore" Version="3.0.2" />
 
-    <!-- Dependency not used by Everest itself, but kept for mod compatibility reasons -->
+    <!--
+        Dependency not used by Everest itself, but kept for mod compatibility reasons.
+
+        DotNetZip has a high-severity vulnerability (https://github.com/advisories/GHSA-xhg6-9j5j-w4vf),
+        but the dependency is no longer maintained, and this kind of vulnerability *isn't really a concern* for Celeste mods,
+        since they already let people run arbitrary code. It's fine to ignore the warning.
+    -->
     <PackageReference Include="DotNetZip" Version="1.16.0" NoWarn="NU1903" />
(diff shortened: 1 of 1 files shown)


microlith57 (approved):
seems reasonable

Kalobi (approved):
sounds good

SnipUndercover:

Celeste.Mod.mm/Celeste.Mod.mm.csproj
```diff
@@ -47,7 +47,7 @@
     <PackageReference Include="MAB.DotIgnore" Version="3.0.2" />
 
     <!-- Dependency not used by Everest itself, but kept for mod compatibility reasons -->
-    <PackageReference Include="DotNetZip" Version="1.16.0" />
+    <PackageReference Include="DotNetZip" Version="1.16.0" NoWarn="NU1903" />
```
SnipUndercover, line 56: I'd like a comment explaining why the `NoWarn` attribute is there just for clarity. LGTM otherwise.

SnipUndercover (changes requested):

Celeste.Mod.mm/Celeste.Mod.mm.csproj
SnipUndercover: I don't mind adding one myself if that's fine (@TheDavSmasher), since the main reason is that the dependency is deprecated and directory traversal attacks aren't *really* a concern for Celeste mods.

TheDavSmasher: @SnipUndercover all the NoWarn does is makes it so Visual Studio won't bug you about it being a vulnerability every time you open it to Everest. It's the entire content of the PR, too

SnipUndercover: Yeah, I know. I approve of the `NoWarn` being there, I just want a comment so that it's directly visible why it's there.
I don't mind putting it there myself if you agree with that.

TheDavSmasher: Sure go ahead

SnipUndercover (approved):

PR #980, opened 2025-09-04
fix wipedebug console command not doing anything

microlith57:
it was looking in -1.celeste for vanilla data; this is fixed by using GetFilename (which respects the debug files' non-numeric naming)

### DIFF
--- a/Celeste.Mod.mm/Patches/SaveData.cs
+++ b/Celeste.Mod.mm/Patches/SaveData.cs
@@ -300,8 +300,9 @@ namespace Celeste {
         public static new bool TryDelete(int slot) {
             bool vanillaExists = false;
             string saveFilePath = patch_UserIO.GetSaveFilePath();
+            string saveFileName = GetFilename(slot);
             if (Directory.Exists(saveFilePath))
-                vanillaExists = File.Exists(Path.Combine(saveFilePath, $"{slot}.celeste"));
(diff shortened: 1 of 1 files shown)


SnipUndercover (approved):
Looks fine. Do we want to also adjust the logger statements?

Celeste.Mod.mm/Patches/SaveData.cs
```diff
@@ -300,16 +300,17 @@ public static void LoadModSaveData(int slot) {
...
+                vanillaExists = File.Exists(Path.Combine(saveFilePath, $"{saveFileName}.celeste"));
 
             if (vanillaExists) {
                 if (!orig_TryDelete(slot)) {
                     return false;
                 }
             }
             else
-                Logger.Warn("SaveData", $"Deleting save data for slot {slot} which has no vanilla data");
+                Logger.Warn("SaveData", $"Deleting save data for slot {saveFileName} which has no vanilla data");
```
Kalobi, line 313: I think changing the logger warnings to be more human-readable is good, but we should change the one in line 415 too for consistency. Seems good otherwise

Kalobi (approved):

SnipUndercover:
This was considered severe enough to warrant bypassing the standard PR merging procedures, discussed in Celestecord.

PR #981, opened 2025-09-05
Add new fixWindBehavior attribute to Wires 

SnipUndercover:
Wind causes wires to go crazy because the `VisualWind` value is way too high. This patch lowers the wind amplitude if the attribute is set to `true`.

This should preferably default to `true` for new wire placements in map editors.

### DIFF
--- a/Celeste.Mod.mm/Patches/Wire.cs
+++ b/Celeste.Mod.mm/Patches/Wire.cs
@@ -1,7 +1,6 @@
 ﻿using Celeste.Mod.Helpers;
 using Microsoft.Xna.Framework;
 using Mono.Cecil;
-using Mono.Cecil.Cil;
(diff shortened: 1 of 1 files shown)


microlith57 (approved):

JaThePlayer (approved):

orbittwz:
Hi!
Old wires that were placed before the fix are not affected?
Because I still have glitches when wires are present in wind rooms...
https://streamable.com/guvbki

microlith57:
@orbittwz yes that's correct, there's an option that fixes it (`fixWindBehavior`) which is enabled by default for new wires, but disabled for old ones
it's up to the mapper to enable the option if they want to fix it

orbittwz:
I see... OK thanks for the quick answer! 😉

PR #985, opened 2025-09-06
add event for `GameLoader.LoadThread`

microlith57:
in analogy to the existing `LevelLoader.LoadingThread` event


Kalobi (approved):

SnipUndercover (changes requested):
LGTM, but I'd like some clarification that this event is invoked *right before* setting `loaded = true;`. 

Also, this event should probably be invoked *before* the `Console.WriteLine(...);` and `LoadTimer.Stop();` calls, so that mods' effect on the game load times can be measured.

SnipUndercover (approved):

PR #988, opened 2025-09-07
Set proper borderless windowed mode

TheDavSmasher:
NOTE: This PR is currently untested, mostly because I didn't know how to before but I'll notify once I do.

Something I noticed is that technically the game has a "windowed borderless" mode by setting the window size to 6, but the way the game is set up, this is a "fake borderless, and the window fitting the window properly isn't consistent, and it can also be moved with certain actions, leading it to become uncentered/offset.

This PR modifies the menu, adding an option between Fullscreen, Windowed, and the addition of Borderless, instead of having a boolean check for Fullscreen.

Internally it depends on SDL2 to handle the Borderless mode, in a similar way Celeste (or rather FNA) already uses SDL2 for fullscreen and windowed, so this _should_ work with the same built-in engine.


TheDavSmasher:
I have no idea why this failed

TheDavSmasher:
Cool

SnipUndercover:

PR #991, opened 2025-09-13
add dialog keys for the critical error handler

microlith57:
:3

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
+++ b/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
@@ -332,22 +332,22 @@ namespace Celeste.Mod.UI {
         private float crouchTimer;
         private bool isColonThree;
 
-        private string textOooops = "Oooops! :(";
-        private string textOooopsColonThree = "Oooops :3";
-        private string textCriticalError = "Celeste/Everest encountered a critical error.";
-        private string textDiscord = "Please report this in the Celeste discord!";
-        private string textDiscordLink = "discord.gg/celeste - channel #modding_help";
-        private string textLogFileBackedUp = "Your log file has been backed up; please attach it to your bug report.";
-        private string textAdditionalErrors = "Additional errors have occurred since the initial crash!";
-        private string textErrorDetails = "Error Details";
+        private string textOooops;
+        private string textOooopsColonThree;
+        private string textCriticalError;
+        private string textDiscord;
+        private string textDiscordLink;
+        private string textLogFileBackedUp;
+        private string textAdditionalErrors;
+        private string textErrorDetails;
 
-        private string textOpenLogFileFolder = "Open log file folder";
-        private string textRetryLevel = "Retry level";
-        private string textSaveQuit = "Save & Quit";
-        private string textSaveProgress = "Save current progress";
-        private string textReturnToMenu = "Return to main menu";
-        private string textExitGame = "Exit Game";
-        private string textRestartGame = "Restart Game";
+        private string textOpenLogFileFolder;
+        private string textRetryLevel;
+        private string textSaveQuit;
+        private string textSaveProgress;
+        private string textReturnToMenu;
+        private string textExitGame;
+        private string textRestartGame;
 
         private CriticalErrorHandler(ExceptionDispatchInfo error, string logFile, string logFileError) {
             Depth += 100; // Render below other overlays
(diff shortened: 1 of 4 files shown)


DashingCat (changes requested):
I agree that adding dialog keys is better than hard-coding text in UIs, and from my testing this PR works as intended, both in English and other languages (which fallback to English text).

However the changes in `CriticalErrorHandler.cs` could be simpler. 
The function `PopulateDialog` is not needed as `Dialog.Get` already uses the English language if the dialog keys are not translated in other languages (relevant code below).

https://github.com/EverestAPI/Everest/blob/33ad712541e60374c2c5e86f7ad931b96291b1bd/Celeste.Mod.mm/Patches/Dialog.cs#L184-L199

I will approve this PR once the changes in `CriticalErrorHandler.cs` are simplified.

DashingCat (approved):

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -348,12 +366,34 @@ private CriticalErrorHandler(ExceptionDispatchInfo error, string logFile, string
             LogFileError = string.IsNullOrEmpty(logFile) ? logFileError : null;
 
             playerShouldTeabag = CoreModule.Settings.CrashHandlerAlwaysTeabag || (!(Settings.Instance?.DisableFlashes ?? true) && new Random().Next(0, 10) == 0);
+            isColonThree = CoreModule.Settings.CrashHandlerAlwaysColonThree || new Random().Next(0, 10) == 0;
```
SnipUndercover:
```suggestion
            Random rng = new();
            playerShouldTeabag = CoreModule.Settings.CrashHandlerAlwaysTeabag || (!(Settings.Instance?.DisableFlashes ?? true) && rng.Next(0, 10) == 0);
            isColonThree = CoreModule.Settings.CrashHandlerAlwaysColonThree || rng.Next(0, 10) == 0;
```

Celeste.Mod.mm/Patches/Commands.cs
```diff
@@ -300,5 +300,10 @@ private static void CmdOpenLog() {
             else
                 Engine.Commands.Log($"{pathLog} does not exist");
         }
+
+        [Command("criticalerrorhandler", "open the critical error handler screen")]
+        private static void CmdCriticalErrorHandler() {
```
SnipUndercover, line 305: Do we want to allow changing the exception message?

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -348,12 +366,34 @@ private CriticalErrorHandler(ExceptionDispatchInfo error, string logFile, string
...
+
+            PopulateDialog();
 
             beforeRenderInterceptor = new BeforeRenderInterceptor(BeforeRender);
             Add(new Coroutine(Routine()));
             Logger.Info("crit-error-handler", $"Created critical error handler for exception {errorType}: {errorMessage}");
         }
 
+        private void PopulateDialog() {
```
SnipUndercover, line 386: Probably add a comment that we're doing this to not have to call `Dialog.Get` multiple times in case something's wrong with the dialog.

SnipUndercover (commented):
Looks good, I just want to test before approving.

SnipUndercover:
Tested locally and it works well. Will approve once the remaining concerns are addressed. May your eyes be blessed with this beauty:
image

SnipUndercover:
...i just realized i tested this but completely forgot to approve.

awesome

Though, one more review is necessary anyways so...

SnipUndercover (approved):

Wartori54:
Looks good, I'd like Snip's comments to get resolved before approving though.
Also, I hate to say it but, the addition of the colon three "Easter Egg" feels a tad bit unnecessary, especially in the crash handler.

Celeste.Mod.mm/Patches/Commands.cs
microlith57: i don't think this is necessary; besides, if this accepted a `string message` then the message could only be one word long

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -327,9 +327,29 @@ private set {
         private VirtualRenderTarget playerRenderTarget;
         private bool UsePlayerSprite => !disablePlayerSprite && State != DisplayState.BlueScreen;
 
+        private static bool canColonThree;
```
DashingCat, line 330: The name `canColonThree` is not very clear, another name like `alreadyCrashed` could be clearer.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
DashingCat: This conversation is marked as resolved but there is no comment in the code, is it intentional?

DashingCat (changes requested):

SnipUndercover (approved):

maddie480:
So... what was decided about the glorious :3 Easter egg?

microlith57:
converted back to draft to resolve conflict and address comments

PR #993, opened 2025-09-16
Remove unused PhotosensitiveMode setting

SnipUndercover:
The setting was a leftover from when #836 was being developed, and was forgotten about when moving it to the vanilla menu.

Its existence causes issues when saving the photosensitive mode preferences, as it takes precedence over vanilla. If the setting is changed in the vanilla options menu, modded settings must also be saved for the change to be actually remembered.

Removing the setting should prevent this from happeming.


Wartori54 (approved):

SnipUndercover:
The issue seems important enough to warrant merging early in my opinion, since this changes the behavior of a vanilla setting.

PR #996, opened 2025-09-21
Rewrite `QueuedTaskHelper`

SnipUndercover:
Simplifies the implementation of `QueuedTaskHelper` and adds documentation. It also *should* fix asset reload issues - often mappers would save their map and there was a small chance that Everest would randomly stop reloading their map until the game is restarted.

This is because the previous implementation did not safeguard against waiting for a negative `TimeSpan`. The new implementation avoids this by creating the delay `Task` immediately after instantiating the `QueuedTask`. This ensures that a negative `TimeSpan` is never awaited (and if it is, now it becomes user error).

Briefly tested this myself and I can confirm that asset reloads seem to work the same.

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
@@ -1,10 +1,13 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading;
 using System.Threading.Tasks;
 
+// the class has been part of the public API for years, so we can't just change the namespace
+// ReSharper disable CheckNamespace
 namespace Celeste.Mod {
+    [Obsolete("Use QueuedTaskHelperV2 instead.")]
     public static class QueuedTaskHelper {
 
         // Make sure to lock Timers and update both of those at the same time before unlocking!
@@ -13,6 +16,7 @@ namespace Celeste.Mod {
 
         public static readonly double DefaultDelay = 0.5D;
 
+        [Obsolete($"Queued tasks now support {nameof(CancellationToken)}s. Pass the token when creating the task and cancel it when necessary.")]
         public static void Cancel(object key) {
             lock (Timers) {
                 if (Timers.Remove(key, out Stopwatch timer)) {
@@ -23,8 +27,11 @@ namespace Celeste.Mod {
             }
         }
 
+        [Obsolete("Use QueuedTaskHelperV2.Do instead.")]
         public static Task Do(object key, Action a)
             => Do(key, DefaultDelay, a);
+
+        [Obsolete("Use QueuedTaskHelperV2.Do instead.")]
         public static Task Do(object key, double delay, Action a) {
             lock (Timers) {
                 if (!Timers.TryGetValue(key, out Stopwatch timer)) {
@@ -56,8 +63,11 @@ namespace Celeste.Mod {
             }
         }
 
+        [Obsolete("Use QueuedTaskHelperV2.Get instead.")]
         public static Task<T> Get<T>(object key, Func<T> f)
             => Get(key, DefaultDelay, f);
+
+        [Obsolete("Use QueuedTaskHelperV2.Get instead.")]
         public static Task<T> Get<T>(object key, double delay, Func<T> f) {
             lock (Timers) {
                 if (!Timers.TryGetValue(key, out Stopwatch timer)) {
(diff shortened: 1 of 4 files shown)


SnipUndercover:
I sent a test build of the PR in Celestecord and asked several people to check if this solves their issue 3 days ago - basically map or whatever it is they do and save files often to trigger the bug.
I've had one person thank me for the build as it's made their mapping experience way smoother, which gives me even more confidence that it should be fine. 

...though with a sample size of 1 that doesn't seem very reassuring.

Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
```diff
@@ -85,6 +95,321 @@ public static Task<T> Get<T>(object key, double delay, Func<T> f) {
...
+        internal static readonly Dictionary<object, QueuedTaskBase> RunningTasks = new();
+        public static readonly TimeSpan DefaultDelay = TimeSpan.FromSeconds(0.5);
+
+        /// <summary>
+        ///   Schedules the given <paramref name="delegate"/> to execute asynchronously after the <see cref="DefaultDelay"/>.
+        /// </summary>
+        /// <remarks>
+        ///   If a task with the same key is scheduled twice, the delay time spans will be overlaid.
+        ///   That is, when the first execution delays for <c>10</c> seconds, and a second execution that delays for <c>5</c> seconds
+        ///   is scheduled <c>7.5</c> seconds after the first one, the actual delay time becomes <c>12.5</c> seconds.
```
Wartori54: You should also note here (and in all the other xml docs) that a task gets overlaid as long as it has not started running, if it has, it won't get overlaid any more until it ends and a new task enters the waiting stage.

Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
```diff
@@ -85,6 +95,321 @@ public static Task<T> Get<T>(object key, double delay, Func<T> f) {
...
+            } while (!DelayTask.IsCompleted);
+        }
+
+        protected void LogError(Exception exception) {
+            Logger.Error(nameof(QueuedTask), $"A queued task with key hashcode {Key.GetHashCode()} failed.");
+            Logger.LogDetailed(exception);
+        }
+
+        protected void CompleteTask() {
+            QueuedTaskHelperV2.RunningTasks.Remove(Key);
```
Wartori54: This write should happen in a lock of `RunningTasks`, dict writes are not thread safe by default.
Furthermore we risk a race condition in `OverlayDelay` where the `Do` method sees there's an active task and calls `OverlayDelay` but the task gets removed and finishes in between the check and the `OverlayDelay` call.

Wartori54 (changes requested):
Looks mostly good, just a few details.

Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
SnipUndercover: Right, I forgor about the lock.

I guess we could remove the task from the dict before the delegate invocation? Alternatively change the check to `task.DelayTask.IsCompleted` in `Do`/`Get<T>`.

Wartori54: You allow for multiple callback for the same key to be running simultaneously that way, which the old implementation allowed. I think that's not a good feature though given that makes it harder to write safe callbacks which interact with a global state.
Adding the lock here should suffice to fix the race condition, if that was not clear.

SnipUndercover: Right 2.0, my concern was that OverlayDelay could be called after exiting the while loop, effectively being a nop. Though I guess that isn't a problem since the callback is about to be executed anyway.

Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
SnipUndercover: Does "running" here mean "delay over, executing delegate", or "the queued task is in state `TaskStatus.Running`"?

Wartori54: The former, `TaskStatus` would always be `Running` regardless of whether it is in the Delay Task or running the callback.

Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
```diff
@@ -324,17 +318,38 @@ public abstract class QueuedTaskBase {
...
+        ///     <item>
+        ///       <term>Awaited</term>
+        ///       <description>The delay is ignored, as the callback is already executing</description>
+        ///     </item>
+        ///   </list>
         /// </summary>
         /// <remarks>
-        ///   For example, when overlaying a <c>5</c>-second delay on top of a <c>10</c>-second delay,
+        ///   When overlaying a <c>5</c>-second delay on top of a <c>10</c>-second delay,
         ///   which starts <c>7.5</c> seconds after the first one, the total delay becomes <c>12.5</c> seconds.
```
microlith57: so this is like
```
0         5        10
+-+-+-+-+-+-+-+-+-+-+-+-+-
|==================>|      0.0 -> 10.0 (10.0 sec)
               |========>| 7.5 -> 12.5  (5.0 sec)
overlays to
|=======================>| 0.0 -> 12.5 (12.5 sec)
```
right? but the terminology of 'overlay' to me implies
```
0         5        10
+-+-+-+-+-+-+-+-+-+-+-+-+-
|==================>|      0.0 -> 10.0 (10.0 sec)
               |=>|        7.5 ->  9.0 ( 1.5 sec)
*overlays to
|==================>|      0.0 -> 10.0 (10.0 sec)
```
but actually this implementation does this: (from what i can see)
```
0         5        10
+-+-+-+-+-+-+-+-+-+-+-+-+-
|==================>|      0.0 -> 10.0 (10.0 sec)
               |=>|        7.5 ->  9.0 ( 1.5 sec)
overlays to
|================>|        0.0 ->  9.0 ( 9.0 sec)
```
which does not seem like overlaying to me! — i do think the behaviour as implemented is _correct_ (or at least sensible), i just feel like the documentation comment isn't representative of that behaviour.

microlith57 (approved):
implementation looks good, just a minor concern around wording :)

Celeste.Mod.mm/Mod/Helpers/QueuedTaskHelper.cs
SnipUndercover: Both the first and second demonstration are describing overlaying correctly. The third one however does not represent how the implementation behaves.

`Task.Delay(TimeSpan)` will begin the delay after the task has been created, not when it is awaited. This means that we can defer waiting for the delay and do initialization stuff in the meantime.
Hitting the `await DelayTask;` statement will retrieve the task reference from `DelayTask` *once* and pause `Queued` until its delay is over, entering `TaskStatus.WaitingForActivation`. Changing `DelayTask` at this point (by `UpdateDelay`) will not have any effect during the `await`, as the task is not running and the task reference in `DelayTask` is not read again.
When the delay is over, the condition inside the loop is re-evaluated, reading the reference again, and the branch back is taken if necessary.

Example:
- `QueuedTask` is instantiated with a delay of 10s and added to the dict
- `Queued` is initialized
- `Queued` yields and pauses
- `Queued` is activated again
- The task reference in `DelayTask` is retrieved
  - *Changing the delay before this point will overwrite it, as its reference has not yet been `await`ed*
- The reference is `await`ed, pausing `Queued`
- Assume `UpdateDelay` has been called with a delay of 5s, 7.5s after the instantiation of `QueuedTask`
- The 10s delay is over, `Queued` is 
activated
- The loop condition is evaluated
- The new task reference in `DelayTask` is retrieved, determining it's not completed yet
- The branch back is taken
- The new task reference in `DelayTask` is retrieved
- The reference is `await`ed, pausing `Queued`
- The 5s delay is over, `Queued` is 
activated *(total wait time is approx. 12.5s)*
- The loop condition is evaluated
  - *Changing the delay before this point will overlay it again, as the loop condition has not yet been evaluated*
  - *Changing the delay **after** this point will ignore it, as the wait is over*
- The branch back is not taken
- The delegate is called
- `QueuedTask` is removed from the dict

I don't know what other word to use instead of overlaying. The delays are taking place in parallel basically.

microlith57: ohhhh, i didn't realise that by the point that the delay is switched out, it's already being awaited... that makes sense. happy with this comment and implementation then :3

Wartori54 (approved):
LGTM

PR #1000, opened 2025-09-27
Added Move methods in Actor for both directions

TheDavSmasher:
It essentially takes the logic found in the MoveH, MoveV, etc methods and creates Vector2 and two float methods that moves in both directions at the same time. It could potentially also make the base methods simply link to these methods to reduce code in the Actor file but I only really added the methods for now. The seem useful.


TheDavSmasher:
I have no idea why the nowarn is part of this PR, I synced from base repo and created a clean new branch?

SnipUndercover:
Probably rebase the branch on `upstream/dev` and force push to the fork to clean the commit history up.

PR #1001, opened 2025-09-27
Added Move methods in Actor for both directions

TheDavSmasher:
It essentially takes the logic found in the MoveH, MoveV, etc methods and creates Vector2 and two float methods that moves in both directions at the same time. It could potentially also make the base methods simply link to these methods to reduce code in the Actor file but I only really added the methods for now. The seem useful.

### DIFF
--- a/Celeste.Mod.mm/Patches/Actor.cs
+++ b/Celeste.Mod.mm/Patches/Actor.cs
@@ -36,51 +36,55 @@ namespace Celeste {
             movementCounter = Vector2.Zero;
         }
 
-        private static Point VectorToPoint(Vector2 source) {
-            return new(
-                (int) Math.Round(source.X, MidpointRounding.ToEven),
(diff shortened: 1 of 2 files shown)


SnipUndercover:
There's lots of issues with the new methods.
- `Approach` does not behave like an approach if the x and y components of `maxMove` aren't equal.
- `Move` does not take the player's subpixels into account when calculating the amount of pixels to move.
- `Move` expects that all movement is perfectly diagonal; if either move vector component is less than `1`, the player won't move at all.
- `MoveExact` will exit early if one of the directions reaches its destination, but not both.
- All movement in Celeste is done per axis, so at this point there is no benefit to adding `Vector2`-based overloads which defy that assumption.
- Even if the logic is changed to move per-axis, this gives no benefit to just calling `MoveH` and `MoveV` manually.

Having discussed this with @Wartori54, we consider this sufficient enough reason to close this PR.

PR #1002, opened 2025-09-28
Maintain original entity / component order when refreshing tracker

psyGamer:
On room load, the entites / components in the tracker will be in their load order.
However when something causes the tracker to refresh, it will sort them in the order of the entity list.
This can cause gameplay differences with overlapping `PlayerCollider`s, because of a different order.

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/Tracker.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/Tracker.cs
@@ -178,6 +178,10 @@ namespace Monocle {
                 return;
             }
             tracker.currentVersion = TrackedTypeVersion;
+
+            var origEntities = tracker.Entities.ToDictionary(entry => entry.Key, entry => new List<Entity>(entry.Value));
+            var origComponents = tracker.Components.ToDictionary(entry => entry.Key, entry => new List<Component>(entry.Value));
+
             foreach (Type entityType in StoredEntityTypes) {
                 ref var list = ref CollectionsMarshal.GetValueRefOrAddDefault(tracker.Entities, entityType, out _);
                 list ??= new();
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -195,6 +199,37 @@ public static void Refresh(Scene scene = null, bool force = false) {
                 }
                 tracker.EntityAdded(entity);
             }
+
+            // Keep original order of existing instances
+            // Especially important for the PlayerCollider component for example
+            foreach (Type entityType in StoredEntityTypes) {
+                ref var newList = ref CollectionsMarshal.GetValueRefOrAddDefault(tracker.Entities, entityType, out _);
+                if (newList == null || !origEntities.TryGetValue(entityType, out var oldList)) {
```
Wartori54: How can `newList` be null?

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -195,6 +199,37 @@ public static void Refresh(Scene scene = null, bool force = false) {
                 }
                 tracker.EntityAdded(entity);
             }
+
+            // Keep original order of existing instances
+            // Especially important for the PlayerCollider component for example
+            foreach (Type entityType in StoredEntityTypes) {
+                ref var newList = ref CollectionsMarshal.GetValueRefOrAddDefault(tracker.Entities, entityType, out _);
```
Wartori54: Should probably use the safer variant here if only the list's contents are mutated.

Celeste.Mod.mm/Patches/Monocle/Tracker.cs
```diff
@@ -195,6 +199,37 @@ public static void Refresh(Scene scene = null, bool force = false) {
...
+                    continue;
+                }
+
+                newList.Sort((lhs, rhs) => {
+                    int lhsIdx = oldList.IndexOf(lhs);
+                    int rhsIdx = oldList.IndexOf(rhs);
+                    if (lhsIdx == -1) return  1;
+                    if (rhsIdx == -1) return -1;
+                    return lhsIdx - rhsIdx;
+                });
```
Wartori54, line 218: This comparison doesn't induce a consistent total order (if neither are on the `oldList` then `lhs > rhs` and `rhs > lhs`), adding `if (lhsIdx == rhsIdx) return 0;` before the first if check would make it consistent.

Wartori54 (changes requested):
Just a few tiny nitpicks. All three comments apply to both entity lists and component lists.

Wartori54 (approved):

SnipUndercover (approved):
This looks like it'll change the order of the items in the new list relative to each other, but it seems to not be a big deal.

SnipUndercover:
The PR is deemed important enough to warrant skipping the 3-day wait.

PR #1004, opened 2025-10-03
Fix weird behaviour and freeze with SubMenu with enterOnSelect and no hoverable items

ACherryJam:
Fixes three small bugs that happen when SubMenu has `enterOnSelect` flag and no hoverable items. I'll refer to these SubMenus as special. 
My first actual pull request so I might be a bit overly verbose. That's probably a good thing? Please let me know.

## 1. TextMenu selection moves one item down per special SubMenu and loses auto scroll when entering Mod Options
Expected behaviour: No selection movement, TextMenu auto scrolls

Why it happens:
1. SubMenu runs `FirstSelection()` when added for the first time
2. `FirstSelection()` calls `MoveNext(1, ...)`, trying to move the selection down
3. do while loop iterates over all elements but can't find any, `Selection >= Items.Count` triggers an exit from the submenu
4. `Exit()` is called that attempts to recover `Container.AutoScroll` state but `containerAutoScroll` was never set because `ConfirmPressed` wasn't called. `Container.AutoScroll` becomes false.
5. `Container.MoveSelection(direction, true)` moves TextMenu selection one item down

Fix: Remove `FirstSelection` calls in `Add` and `Insert`. It seems that these calls are redundant because `FirstSelection` and `LastSelection` methods are called in `ConfirmPressed` and that method is the only one that can get the menu focused.

## 2. TextMenu loses auto scroll when selecting special SubMenu from above
Expected behaviour: TextMenu auto scrolls

Why it happens:
The preservation of `Container.AutoScroll` state is after the Selection methods, so similarly, `Exit()` gets called and AutoScroll's corrupts.

Fix: Place `containerAutoScroll` assignment before Selection calculations 

## 3. Game freezes when selecting special SubMenu from below
Expected behaviour: SubMenu is skipped, the item above is selected

Why it happens:
`LastSelection` is called from `ConfirmPressed` and because there's no Hoverable items it returns 0. `MoveSelection(0, ...)` doesn't move `Selection` because `Math.Sign(0)` is 0. The game stuck in an infinite loop.
```c#
            public void LastSelection() {
                Selection = LastPossibleSelection;
                MoveSelection(0, false);
            }
```

Fix: Make `LastSelection` an inverse of `FirstSelection`, set `Selection` to the upper bound and move down.

A mod with a special SubMenu for replication: [LastSelectionSubMenuFreeze.zip](https://github.com/user-attachments/files/22689151/LastSelectionSubMenuFreeze.zip)

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
+++ b/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
@@ -446,7 +446,7 @@ namespace Celeste {
             /// <inheritdoc cref="TextMenu.Current"/>
             public TextMenu.Item Current {
                 get {
-                    if (Items.Count <= 0 || Selection < 0) {
+                    if (Items.Count <= 0 || Selection < 0 || Selection >= Items.Count) {
                         return null;
                     }
                     return Items[Selection];
@@ -640,8 +640,8 @@ namespace Celeste {
             /// Set the selection to the last possible <see cref="TextMenu.Item"/>.
             /// </summary>
             public void LastSelection() {
-                Selection = LastPossibleSelection;
-                MoveSelection(0, false);
+                Selection = Items.Count;
+                MoveSelection(-1, false);
             }
 
             /// <inheritdoc cref="TextMenu.MoveSelection(int, bool)"/>
--- a/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
+++ b/Celeste.Mod.mm/Mod/UI/TextMenuExt.cs
@@ -558,9 +558,6 @@ namespace Celeste {
                     Container.Add(item.ValueWiggler = Wiggler.Create(0.25f, 3f, null, false, false));
                     Container.Add(item.SelectWiggler = Wiggler.Create(0.25f, 3f, null, false, false));
                     item.ValueWiggler.UseRawDeltaTime = (item.SelectWiggler.UseRawDeltaTime = true);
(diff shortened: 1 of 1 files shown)


ACherryJam:
Previously the special SubMenu would get skipped when entry is attempted. This behaviour isn't intuitive so I changed it to actually entering the menu. When entering from the above or below the scroll is aligned to the first and last item respectfully.

Before (courtesy of rain): 

https://github.com/user-attachments/assets/bda7ee25-f452-428e-8699-4ebcb69c8bc5

After:

https://github.com/user-attachments/assets/11c6663b-bf20-4182-b7fb-5d95140c9970

Demo from the After vid: [LastSelectionSubMenuFreeze.zip](https://github.com/user-attachments/files/22711809/LastSelectionSubMenuFreeze.zip)

microlith57 (approved):
cool :+1:

DashingCat (approved):

Wartori54 (approved):
Extremely solid first PR, good job! Thank you so much for the test mod too, highly appreciated!
Code looks good and does indeed work, so this is a LGTM on my end.

The only small nitpick I have is with the removal of those lines of code in `Insert` and `Add` where if you somehow ended up with your selection being `-1` (or, more generally, simply out of bounds) and a hoverable item gets added you would not automatically snap to it (as it would make sense to do), yet I could not come up with any "sane" scenario for that to happen so lets roll with that.

PR #1008, opened 2025-10-07
LevelEnter.OnPlayVignette event

SilverDorian46:
This allows the user to conditionally play a custom vignette before the next scene is instantiated


JaThePlayer:
I think the event should return a `Scene?`, which `PlayCustomVignette` will set `Engine.Scene` to if not null. Sounds good otherwise.

SilverDorian46:
> I think the event should return a `Scene?`, which `PlayCustomVignette` will set `Engine.Scene` to if not null. Sounds good otherwise.

now that I think about it, yes. I'll try changing it in a moment

JaThePlayer (approved):

SnipUndercover:
Will approve once the TAS check syncs just in case, otherwise looks good.

SnipUndercover (approved):

PR #1011, opened 2025-10-16
Get page size dynamically (issue #1010)

### LINKED ISSUE #1010: Crash on 16/64k pagesize system

Yes

Yes

### Describe the bug

In [https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/Everest/BOOT.cs#L64](https://github.com/EverestAPI/Everest/blob/dev/Celeste.Mod.mm/Mod/Everest/BOOT.cs#L64), it only allocates 0x1000 (4096 decimal), which causes EINVAL errno on 16k/64k pagesize systems due to not being a multiple of the system pagesize since it's smaller than pagesize. See: https://man7.org/linux/man-pages/man2/mprotect.2.html

### Steps to reproduce

1. Run

### Expected behavior

Work

### Operating System

Arch Linux ARM

### Everest Version

self-compiled dev branch

_No response_

### Additional context

_No response_

DashingCat:
The memory page size was assumed to always be `4096`, which is not the case (like in #1010).

This PR uses the `getpagesize()` system call (available in `libc` like `mprotect`) to dynamically figure out the page size.

_Note: I was not able to test this on a system with a page size different than `4096`._

Fixes #1010.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/BOOT.cs
+++ b/Celeste.Mod.mm/Mod/Everest/BOOT.cs
@@ -53,19 +53,25 @@ namespace Celeste.Mod {
                 // So probe for it before continuing to boot on Linux
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) {
                     [DllImport("libc", SetLastError = true)]
+                    static extern nuint getpagesize();
+
+                    [DllImport("libc", SetLastError = true)]
                     static extern int mprotect(IntPtr ptr, nuint len, int prot);
                     const int PROT_READ = 1, PROT_WRITE = 2, PROT_EXEC = 4;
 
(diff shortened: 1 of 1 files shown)


microlith57 (commented):
seems to be failing the tas sync check

DashingCat:
Thanks @microlith57 for the heads up, I fixed the `heapPage` declaration in the latest commit, and now all checks have passed.

NoaHimesaka1873:
Tested on my MacBook Air M2 running Asahi ALARM with 16k pagesize and it works fine!

EDIT: with screenshot: image

Wartori54 (approved):

microlith57 (approved):

PR #1023, opened 2025-11-22
fix SeekerEffectsController overwriting music layer 3

microlith57:
since this behaviour is silly and inconsistent, and i can't think of any reason anyone would want it to happen on purpose, i think it's sensible to simply disable it globally including in vanilla maps.

tested and confirmed working in vanilla 5a. 

reported [here](https://discord.com/channels/403698615446536203/429775320720211968/1439856845362561134)

### DIFF
--- a/Celeste.Mod.mm/Patches/SeekerEffectsController.cs
+++ b/Celeste.Mod.mm/Patches/SeekerEffectsController.cs
@@ -1,0 +1,20 @@
+using Monocle;
+using MonoMod;
+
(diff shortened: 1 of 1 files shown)


SnipUndercover (approved):

SnipUndercover:
Linking [this Celestecord discussion](https://discord.com/channels/403698615446536203/429775320720211968/1439852107812438026) for context.

Wartori54 (approved):

PR #1025, opened 2025-11-23
Fix format parameter not being passed to handler

SnipUndercover:
Formatted values inside of interpolated strings will not have their format applied, as the `LogInterpolatedStringHandler` class (and its generic variant) did not pass the `format` argument in `AppendFormatted<T>` overloads that included it. This meant that the format would not be applied.
```cs
Logger.Debug("MyMod", $"{255:X4}"); // will output "255" instead of "00FF"
```
Tested with @2CM - they compiled Everest with this patch and it seemed to work.


microlith57 (approved):
yep

JaThePlayer (approved):

SnipUndercover:
Since this is a bug that has already gotten into stable, I think this warrants skipping the last-call period and merging directly to dev. Discussed in #modding_dev.

PR #1026, opened 2025-11-27
Show more verbose messages on duplicate tileset IDs

SnipUndercover:
Makes the `Autotiler` constructor explicitly throw when a duplicate tileset ID is encountered. It also shows a new postcard when applicable. The Unicode code point is also printed in case the tileset ID happens to be an unprintable character, as several people reported that high tileset IDs (such as `U+5337`) turn into `?` (`U+003F`) when viewed in `log.txt`.

### IL patch, decompiled
image

### Postcard
image

### Logged exception
image

### DIFF
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -491,6 +491,11 @@ namespace Celeste {
 }
 
 namespace MonoMod {
+
+    /// <summary>
+    ///   Patches the <see cref="Celeste.Autotiler"/> constructor to explicitly throw
+    ///   an error when encountering a duplicate tileset ID.
+    /// </summary>
     [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchAutotilerCtor))]
     class PatchAutotilerCtorAttribute : Attribute { }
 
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -466,3 +489,22 @@ private class patch_Masked {
 
     }
 }
+
+namespace MonoMod {
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchAutotilerCtor))]
+    class PatchAutotilerCtorAttribute : Attribute { }
+
+    static partial class MonoModRules {
+        public static void PatchAutotilerCtor(ILContext context, CustomAttribute attrib) {
```
microlith57, line 503: it'd be nice to have a comment showing what this patch does (preferably a diff of the il like on some of my patches), especially since this deals with locals by number

microlith57 (approved):

microlith57 (approved):

Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -406,6 +413,22 @@ public bool TryGetCustomDebrisImpactSfx(out string sfxEvent, char tiletype) {
             return !string.IsNullOrEmpty(sfxEvent = lookup.TryGetValue(tiletype, out patch_TerrainType t) ? t.DebrisImpactSfx : "");
         }
 
+        public void ThrowOnDuplicateId(char id, string xmlName) {
```
JaThePlayer: should be private

JaThePlayer (changes requested):
Just a tiny change needed, looks good otherwise

JaThePlayer (approved):

Issue #1029, opened 2025-11-29
Surface Sound Priority attribute for Solid Decals


SilverDorian46:
### I am planning on implementing this myself

Yes

### Describe your request

An extra attribute for Solid Decals via Decal Registry, called something like "priority", which takes an integer and assigns it to the solid's `SurfaceSoundPriority` variable

This is useful in cases where the player touches or stands on multiple platforms stacked on one another. For such a case, the player's code checks each of the platforms for its priority, and the one with the highest priority is returned, the sound of which is played

`IntroPavement` (added by `IntroCar`) is the only entity in the base game that has a non-zero surface sound priority, and no Everest feature as of yet has allowed mappers to set that variable

PR #1030, opened 2025-11-29
Add "priority" attribute for "solid" decals

SilverDorian46:
See #1029

### DIFF
--- a/Celeste.Mod.mm/Mod/Registry/DecalRegistryHandlers/SolidDecalRegistryHandler.cs
+++ b/Celeste.Mod.mm/Mod/Registry/DecalRegistryHandlers/SolidDecalRegistryHandler.cs
@@ -3,7 +3,7 @@
 namespace Celeste.Mod.Registry.DecalRegistryHandlers; 
 
 internal sealed class SolidDecalRegistryHandler : DecalRegistryHandler {
-    private int _x, _y, _width, _height, _index;
+    private int _x, _y, _width, _height, _index, _priority;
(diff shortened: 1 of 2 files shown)


microlith57 (approved):
i think `priority` is not a very descriptive name for the xml atrribute but otherwise this looks good

SilverDorian46:
the TAS checks were cancelled?

Wartori54:
> i think `priority` is not a very descriptive name for the xml atrribute but otherwise this looks good

What about `soundPriority`?

This PR LGTM if the naming complaint is addressed.

SilverDorian46:
> > i think `priority` is not a very descriptive name for the xml atrribute but otherwise this looks good
> 
> What about `soundPriority`?
> 
> This PR LGTM if the naming complaint is addressed.

I do believe `priority` makes it more consistent with `index`, plus it's less characters to type, and with that it sort of strikes a good balance

we can always update the article on Decal Registry later to clarify what it's for

Wartori54:
> > > i think `priority` is not a very descriptive name for the xml atrribute but otherwise this looks good
> > 
> > 
> > What about `soundPriority`?
> > This PR LGTM if the naming complaint is addressed.
> 
> I do believe `priority` makes it more consistent with `index`, plus it's less characters to type, and with that it sort of strikes a good balance
> 
> we can always update the article on Decal Registry later to clarify what it's for

Agreed.

Wartori54 (approved):

PR #1033, opened 2025-12-07
Remove dependency hot loading

SnipUndercover:
This feature caused a lot of problems for people than it was worth convenience-wise.
- Virtually no mod expects to be hot loaded in this way, and can likely crash even after the try/catches in `OuiDependencyDownloader`
- Crashes interrupt the download/update process, often making users stuck with a configuration of mods and Everest that kills the game on bootup until they update from outside the game
- Lots of people make `#modding_help` reports of crashes when downloading dependencies that are fixed by restarting
- The state of the game at the time of hot loading is not the same as the state of the game at the time of a cold boot, potentially making mods not fully initialize
- Take the following scenario:
  - The user has mod A enabled that optionally depends on mod C
  - The user installs mod B that requires C to be present
  - When downloading dependencies, C will be downloaded and loaded *after* A, even though A optionally depends on C and needs to load after C if present.
  
  This can cause unexpected behavior if the user did not restart after installing deps.

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/OuiDependencyDownloader.cs
+++ b/Celeste.Mod.mm/Mod/UI/OuiDependencyDownloader.cs
@@ -51,6 +51,20 @@ namespace Celeste.Mod.UI {
             }
         }
 
+        // rip hot loading. this feature caused a lot of problems for people than it was worth convenience-wise.
+        // - virtually no mod expects to be hot loaded in this way, and can likely crash even after the try/catches seen here
+        // - crashes interrupt the download/update process, often making users stuck with a configuration of mods and everest
+        //   that kills the game on bootup until they update from outside the game
+        // - lots of people make #modding_help reports of crashes when downloading dependencies that are fixed by restarting
+        // - the state of the game at the time of hot loading is not the same as the state of the game at the time of a cold boot,
+        //   potentially making mods not fully initialize
+        // - take this scenario: the user has mod A enabled that optionally depends on mod C. the user installs mod B that requires
+        //   C to be present. when downloading dependencies, C will be downloaded and loaded *after* A, even though A optionally depends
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/UI/OuiDependencyDownloader.cs
```diff
@@ -352,6 +370,7 @@ private bool unblacklistMods(HashSet<string> modFilenamesToUnblacklist) {
                     foreach (string line in currentBlacklist) {
                         if (modFilenamesToUnblacklist.Contains(line)) {
                             // comment this line to unblacklist this mod.
+                            LogLine(string.Format(Dialog.Get("DEPENDENCYDOWNLOADER_MOD_UNBLACKLIST"), line));
                             blacklistTxt.WriteLine("# " + line);
```
microlith57: **nitpick:**
```suggestion
                            LogLine(string.Format(Dialog.Get("DEPENDENCYDOWNLOADER_MOD_UNBLACKLIST"), line));
                            // comment this line to unblacklist this mod.
                            blacklistTxt.WriteLine("# " + line);
```

microlith57 (approved):
o7

SnipUndercover:
It just occurred to me that users could still encounter crashes after restarting.
- Extended Variant Mode is enabled
- Maddie's Helping Hand is disabled and out of date
- Some other mod X that depends on Maddie's Helping Hand is installed
- Installing dependencies for X will update Extended Variant Mode and unblacklist Maddie's Helping Hand
- Maddie's Helping Hand is not updated; this new configuration of old MHH and new ExtVars will cause a crash on boot

Should we be concerned about this?

**EDIT:** Actually, new ExtVars will probably depend on a newer version of MHH to prevent the crash, so this is probably a non-issue.

DashingCat (approved):

Wartori54 (approved):

SSM240:
good riddance

PR #1034, opened 2025-12-08
Actually avoid string interpolation in the Logger

wuke32767:
after i read the source code of DefaultInterpolatedStringHandler, i think current LogInterpolatedStringHandler can not avoid string interpolation.
https://github.com/dotnet/runtime/blob/74cf618d63c3d092eb91a9bb00ba8152cc2dfc76/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/DefaultInterpolatedStringHandler.cs#L145-L156

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Logger.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Logger.cs
@@ -346,7 +346,7 @@ namespace Celeste.Mod {
             }
             return false;
         }
-        
+
         /// <summary>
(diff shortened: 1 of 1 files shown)


JaThePlayer:
The `out bool shouldLog` is treated specially by the compiler and will completely bypass AppendLiteral/AppendFormatted calls if its false. Example from Everest.Boot decompiled:
image

wuke32767:
thanks, i did not investigated.

PR #1035, opened 2025-12-09
Change loglevel of ILCursorExtensions logging to Verbose

grog398:
Does exactly what the title says, ILCursorExtensions logging makes the logfile extremely cluttered, and is, well, extremely verbose. Changing the loglevel here would allow for LogLevel.Debug to be more useful and easier to read for codemodders.


microlith57 (approved):

SnipUndercover:
The motivation behind making these debug was to provide an easier at-a-glance way to tell whether cursor matching is not done correctly. However with more mods using these extensions it was made clear that the match failure logs could be a cause of unnecessary log spam, so I agree with making those (lines `172` and `423`) verbose.

I still want to keep the start/end logs in debug to give the user the option to see or hide the detailed matching steps by changing the log levels in `modsettings-Everest.celeste`.

SnipUndercover (approved):

PR #1037, opened 2025-12-10
Add workspace to MiniInstaller

### LINKED ISSUE #983: orig/Celeste.exe is coreified and does not run

Yes

Yes

### Describe the bug

If the installation of Everest fails after coreification but before the apphost setup, the next install will cause Olympus to delete the `orig` directory under the assumption that this is a vanilla install and safe to backup.
This causes the vanilla executable in `orig/Celeste.exe` to be coreified. Trying to run coreified vanilla will always crash with this message in the Event Viewer on Windows:
```
Application: Celeste.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: exception code e0434352, exception address 00007FF9899C7F7A
Stack:
```

### Steps to reproduce

1. Try to install Everest
2. Have the install fail after coreification but before the apphost setup (so before or while running MonoMod)
3. Install Everest from Olympus
4. `orig/Celeste.exe` is coreified

### Expected behavior

`orig/Celeste.exe` remains a .NET Framework assembly and runs normally.

### Operating System

N/A (OS-agnostic)

### Everest Version

5800, but likely applies to all core (4465+) versions

_No response_

### Additional context

Since `MiniInstaller` does not do any cleanup if installation fails, `Celeste.exe` is converted to .NET Core without applying any of the Everest patches. Then, when installing from Olympus, it finds that `Celeste.exe` is vanilla (unaware of the fact that it's coreified), and deletes `orig` under the oblivious assumption that this is a vanilla install and safe to back up.

The deletion logic can be found in Olympus' [CmdInstallEverest](https://github.com/EverestAPI/Olympus/blob/977e47bd2addb50ced3a81ee8a006bfc050e8afa/sharp/CmdInstallEverest.cs#L11-L21).

### LINKED ISSUE #984: Installing Everest with coreified vanilla always fails and deletes Celeste.dll

Yes

Yes

### Describe the bug

Installing Everest with a coreified `orig/Celeste.exe` will always fail and delete `Celeste.dll`, but not the apphost.
This is what causes the "entity not found: path/to/Celeste.dll" message box when trying to run Everest.
This is a side-effect of #983.

### Steps to reproduce

1. Install Everest with a coreified `orig/Celeste.exe` (see #983)
2. Wait until it fails
3. Try to run Everest

### Expected behavior

`Celeste.dll` does not get deleted / the installation succeeds.
This situation shouldn't be happening at all.

### Operating System

N/A (OS-agnostic)

### Everest Version

5580, but likely applies to all core (4465+) versions

_No response_

### Additional context

The issue arises because `MiniInstaller.DepCalls.ConvertToNETCore` returns early if it sees that the source assembly is already coreified, without copying it to the destination (line `103`).
https://github.com/EverestAPI/Everest/blob/676c509277754b95516af65a56853d00e4e1e855/MiniInstaller/DepCalls.cs#L92-L117

`MiniInstaller` tries to coreify vanilla, but since this method returns early, it does nothing. Because of that, the `Celeste.exe` apphost is not deleted.
Then, when running MonoMod, it tries to resolve the `Celeste` assembly, it finds the apphost instead and dies a brutal death.
https://github.com/EverestAPI/Everest/blob/676c509277754b95516af65a56853d00e4e1e855/MiniInstaller/Program.cs#L85-L90
Since `MiniInstaller` deletes `Celeste.dll` in the beginning and does not do any cleanup after installation fails, `Celeste.dll` is gone while the `Celeste.exe` apphost remains. Running the game at this point yields the "entity not found: path/to/Celeste.dll" message box that has been plaguing #modding_help for ages.

DashingCat:
This PR adds a workspace folder to the MiniInstaller installation process, so it doesn't overwrite `Celeste.exe` early (for example, during the step applying Everest patches, after coreification).

The workspace is used to store the `Celeste.exe`/`Celeste.dll` files, until after the step applying Everest patches, reducing the time window during which the process can be interrupted and left in a state which would allow `orig/Celeste.exe` to be coreified.

This is a work in progress, the following still needs to be done:
- [x] Testing on Windows
- [x] Testing on MacOS
- [x] _Applying the changes to MiniInstaller's "FastMode" (not needed)_

Fixes #983.
Fixes #984.
_As far as I understand, #984 occurs only because of #983_.


MiniInstaller/LibAndDepHandling.cs
```diff
@@ -268,6 +267,7 @@ public static void SetupAppHosts(string appExe, string appDll, string resDll = n
                 // Bind Linux apphost
                 Logger.LogLine($"Binding Linux apphost {Path.ChangeExtension(appExe, null)}");
                 HostWriter.CreateAppHost(Path.Combine(hostsDir, "linux"), Path.ChangeExtension(appExe, null), Path.GetRelativePath(Path.GetDirectoryName(appExe), appDll));
+                File.Delete(Globals.PathCelesteExe);
```
Wartori54: What's the motivation behind this deletion? `Celeste.exe` should not exist anymore on linux systems here when on the standard mode. Is this for the fast mode?

MiniInstaller/Program.cs
```diff
@@ -80,25 +80,43 @@ public static int StandardMode(string[] args) {
...
+                Directory.CreateDirectory(miniInstallerWorkspaceWIP);
+                Directory.CreateDirectory(miniInstallerWorkspaceDone);
+
+                string coreifiedCeleste = Path.Combine(miniInstallerWorkspaceWIP, "Celeste.exe");
+                string coreifiedModdedCeleste = Path.Combine(miniInstallerWorkspaceDone, "Celeste.dll");
+
                 DepCalls.LoadModders();
 
-                DepCalls.ConvertToNETCore(Path.Combine(Globals.PathOrig, "Celeste.exe"), Globals.PathCelesteExe);
+                DepCalls.ConvertToNETCore(Path.Combine(Globals.PathOrig, "Celeste.exe"), coreifiedCeleste);
```
Wartori54: In order to further prevent any issues similar to #983 we should be using an intermediary file name here, rather than reusing `Celeste.exe`. In fact, [fast mode already does this](https://github.com/EverestAPI/Everest/blob/dcfbe559a8b8cd5fa5f1f7c2aef4e9e083fa9637/MiniInstaller/Program.cs#L192), although it should not remain cached in this case.

MiniInstaller/LibAndDepHandling.cs
Wartori54, line 279: `Globals.PathCelesteExe` will remain in MacOS as well, it should be deleted just like in linux.

MiniInstaller/Program.cs
```diff
@@ -80,25 +80,43 @@ public static int StandardMode(string[] args) {
...
+                    Logger.LogLine("MiniInstaller workspace already exists, cleaning before continuing.");
+                    Directory.Delete(miniInstallerWorkspace, true);
+                }
+
+                Directory.CreateDirectory(miniInstallerWorkspaceWIP);
+                Directory.CreateDirectory(miniInstallerWorkspaceDone);
+
+                string coreifiedCeleste = Path.Combine(miniInstallerWorkspaceWIP, "Celeste.exe");
+                string coreifiedModdedCeleste = Path.Combine(miniInstallerWorkspaceDone, "Celeste.dll");
+
```
Wartori54, line 91: Using two subdirectories feels an overkill, what about using just one and changing the coreified executable name to not collide with `Celeste.exe` (see comment below)?

MiniInstaller/Program.cs
```diff
@@ -80,25 +80,43 @@ public static int StandardMode(string[] args) {
...
-                DepCalls.RunMonoMod(Globals.PathCelesteExe, dllPaths: mods);
+                DepCalls.RunMonoMod(coreifiedCeleste, coreifiedModdedCeleste, dllPaths: mods);
 
                 string hookGenOutput = Path.Combine(Globals.PathGame, "MMHOOK_" + Path.ChangeExtension(Path.GetFileName(Globals.PathCelesteExe), ".dll"));
-                DepCalls.RunHookGen(Globals.PathCelesteExe, Globals.PathCelesteExe);
+                DepCalls.RunHookGen(coreifiedModdedCeleste, Globals.PathCelesteExe);
                 DepCalls.RunMonoMod(hookGenOutput, dllPaths: mods); // We need to fix some MonoMod crimes, so relink it against the legacy MonoMod layer
 
-                MiscUtil.MoveExecutable(Globals.PathCelesteExe, Globals.PathEverestDLL);
+                MiscUtil.MoveExecutable(coreifiedModdedCeleste, Globals.PathEverestDLL);
```
Wartori54: Also since we are already moving temporary results into a subdirectory until they are finished we should also do everything in there, not just the main Celeste binary handling: patched FNA and hookgen should be thrown in the subdirectory before overwritting the existing files.

Wartori54 (changes requested):
Just some polish needed here and there.
About your fast mode question, I'd be fine with merging without any changes to fast mode, it is meant for advanced users only, so if the installation breaks the user will notice that and be able to fix it.

DashingCat:
Thanks @Wartori54 for the feedback, all comments are addressed in the latest commit.

MiniInstaller/DepCalls.cs
```diff
@@ -63,7 +63,7 @@ public static void RunMonoMod(string asmFrom, string asmTo = null, string[] dllP
         string asmTmp = Path.Combine(Globals.PathTmp, Path.GetFileName(asmTo));
         try {
             // We're lazy.
-            Environment.SetEnvironmentVariable("MONOMOD_DEPDIRS", Globals.PathGame);
+            Environment.SetEnvironmentVariable("MONOMOD_DEPDIRS", $"{Globals.PathMiniInstallerWorkspace}:{Globals.PathGame}"); // Prioritize workspace
```
Wartori54: While on Unix systems the path separator is always `:`, on windows [it is not](https://github.com/dotnet/dotnet/blob/a8faa5ebe301d5a258c5622975df78254e5395db/src/runtime/src/libraries/Common/src/System/IO/PathInternal.Windows.cs#L45), use `Path.PathSeparator` instead to ensure it working on all platforms.

MiniInstaller/DepCalls.cs
```diff
@@ -84,7 +84,7 @@ public static void RunMonoMod(string asmFrom, string asmTo = null, string[] dllP
     public static void RunHookGen(string asm, string targetName) {
         Logger.LogLine($"Running MonoMod.RuntimeDetour.HookGen for {asm}");
         // We're lazy.
-        Environment.SetEnvironmentVariable("MONOMOD_DEPDIRS", Globals.PathGame);
+        Environment.SetEnvironmentVariable("MONOMOD_DEPDIRS", $"{Globals.PathMiniInstallerWorkspace}:{Globals.PathGame}"); // Prioritize workspace
```
Wartori54: Same as above here.

Wartori54 (changes requested):
Other than this LGTM. Would appreciate some testing on other platforms before merging though.

DashingCat:
The latest commit uses `Path.PathSeparator` instead of `:`.

Wartori54 (approved):

MiniInstaller/MiscUtil.cs
```diff
@@ -72,18 +72,15 @@ public static bool IsSteamworksNet(string file) {
     }
     // This is not "pure" but I guess it also somewhat fits here
     public static void MoveExecutable(string srcPath, string dstPath) {
-        File.Delete(dstPath);
-        File.Move(srcPath, dstPath);
+        File.Move(srcPath, dstPath, true);
```
microlith57: nitpick: (for readability)
```suggestion
        File.Move(srcPath, dstPath, overwrite: true);
```
(& similarly for the other `File.Move` calls below)

MiniInstaller/LibAndDepHandling.cs
```diff
@@ -106,8 +106,7 @@ void CopyNativeLib(string src, string dst) {
 
         // Copy our Steamworks.NET.dll
         string steamworksLibDst = Path.Combine(Globals.PathGame, "Steamworks.NET.dll");
-        File.Delete(steamworksLibDst);
-        File.Copy(steamworksLibSrc, steamworksLibDst);
+        File.Copy(steamworksLibSrc, steamworksLibDst, true);
```
microlith57: nitpick: (for readability)
```suggestion
        File.Copy(steamworksLibSrc, steamworksLibDst, overwrite: true);
```

MiniInstaller/Program.cs
```diff
@@ -80,25 +80,43 @@ public static int StandardMode(string[] args) {
                 LibAndDepHandling.SetupNativeLibs();
                 LibAndDepHandling.CopyControllerDB();
 
+                if (Directory.Exists(Globals.PathMiniInstallerWorkspace)) {
+                    Logger.LogLine("MiniInstaller workspace already exists, cleaning before continuing.");
+                    Directory.Delete(Globals.PathMiniInstallerWorkspace, true);
```
microlith57: nitpick: (readability)
```suggestion
                    Directory.Delete(Globals.PathMiniInstallerWorkspace, recursive: true);
```
(& similarly for the other recursive Directory.Delete)

microlith57 (approved):
code looks good, just a few missing param names (non-blocking). still would be good to test on windows/macos :p

SnipUndercover (changes requested):
Adding a "request changes" review just so that this doesn't get accidentally merged without testing, will turn into an approval when it's done.

DashingCat:
Parameter names have been added in the latest commit.

SnipUndercover (changes requested):
Code-wise looks good, but upon testing MiniInstaller crashes when patching FNA. It attempts to resolve the Celeste assembly, but it finds `Celeste.exe` which is a Piton apphost on already modded installs. MonoMod then tries to read that as a managed .NET assembly and promptly crash.

MonoMod does not consider the coreified vanilla Celeste copy in the workspace, as it is named `Celeste.Core.dll` instead of `Celeste.dll`.

I cannot commit to your branch, so I have sent a patch file in \#modding_dev on the Celeste Discord server.
I tested it by running MiniInstaller twice from a fresh copy of Celeste and confirmed it works.

SnipUndercover:
Also this doesn't really solve the problem of `Celeste.dll` being missing when the installation is interrupted. We could skip deleting `Celeste.dll` and just have the copy from the workspace overwrite the dll?

DashingCat:
From my testing, @SnipUndercover patches work on Linux as well.

SnipUndercover (approved):
Confirmed the PR works on MacOS.
LGTM.

PR #1042, opened 2025-12-17
Add support for activating cutscenes via interact to `EventTrigger`

aonkeeper4:
This PR adds a `useInteract` option to `EventTrigger` to allow activating the cutscene via an interact prompt, rather than via entering the trigger. To complement this, it also adds a `deleteFlag` option to allow preventing the trigger from loading via a flag. The reasoning behind this is that while doing the "don't load if this flag is set" logic in the cutscene works for activation via entering the trigger, for activation via interact, the interact prompt would still be visible even though the cutscene would not play.
Requires map editor support for 2 new attributes on `eventTrigger`: `useInteract` and `deleteFlag`.

### DIFF
--- a/Celeste.Mod.mm/Patches/EventTrigger.cs
+++ b/Celeste.Mod.mm/Patches/EventTrigger.cs
@@ -19,6 +19,8 @@ namespace Celeste {
         private bool useInteract;
         // whether `OnEnter` was called via the `TalkComponent` callback
         private bool interactTriggered;
+        // the `TalkComponent` itself
+        private TalkComponent talkComponent;
 
         private patch_EventTrigger(EntityData data, Vector2 offset) : base(data, offset) { }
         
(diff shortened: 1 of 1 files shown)


SilverDorian46:
I mean... on one hand, why not patch the `InteractTrigger`?
(preferably with a new `CustomInteract` attribute / event, since it uses different strings)

but on the other hand, yeah, the `deleteFlag` option may be useful.
or... you could maybe add a parameter (or two?) to the `CustomEvent` attribute, and that will check for whether a flag is set (or unset) before doing the cutscene

aonkeeper4:
I feel patching `InteractTrigger` and using a second attribute would cause unnecessary duplication (I did want to patch it originally to use the same `CustomEvent` attribute, but that would require backwards-incompatible changes), and the only real difference in behaviour it has is the ability to trigger different cutscenes on sequential interacts. I don't mind taking this approach though if people agree it would be better.

aonkeeper4:
I noticed `InteractTrigger`s don't prevent themselves from activating multiple times in the same room, so I thought that an `onlyOnce` option (defaulting to `true`) would be a good idea to add too.
Also, I forgot to mention this in the PR body but the location of the interact prompt can be customized via the position of the trigger's node.

Kalobi:
If we have any kind of "only once" option, we should definitely have one that's only once per session (so it remembers the interaction if you leave and reenter the room). At a glance, I don't think this option does that? See the "onlyOnce" vs "oncePerSession" flags in [LuaCutscenes](https://github.com/Cruor/LuaCutscenes/blob/master/Triggers/LuaCutsceneTrigger.cs)

aonkeeper4:
Superseded by #1051.

PR #1045, opened 2025-12-24
Add a check for Camera Advance Target in GetFullCameraTargetAt

SilverDorian46:
See issue #1044

PR #1050, opened 2025-12-28
Add LevelSet check in GetFullCameraTargetAt

SilverDorian46:
the camera's position can have an effect on the gameplay in a few cases (e.g. spinner / lightning collidability).
this PR adds a check for whether the current LevelSet is "Celeste" before executing the modified method


SilverDorian46:
closing because it turns out that the behaviour of `GetFullCameraTargetAt` is even more broken than I thought (it doesn't account for leaving the triggers, for instance), and fixing that wouldn't be as easy.
#1045 will also be reverted shortly

PR #1051, opened 2025-12-29
Implement Interact Trigger API for mods

SilverDorian46:
## What is an Interact Trigger?

the Interact Trigger is a generalised trigger of which the function is to play a cutscene upon being interacted with, by pressing [Interact] while the player is within its bounds.

one particular feature of it is that it can have one or multiple events.
when it first appears, its "event index" begins at 0. this is used to get the current event, and is incremented on playing each event.
upon incrementing, a Session flag called `it_` plus the name of the current event, is also set. this is used to keep track of which events are done even if the player goes to a different room and comes back.
on the first interaction, the trigger plays event 1. then, given it has multiple events, on subsequent interactions it will play events 2, 3, and so on. when it has no more events to play (i.e. when its event index reaches the number of events it has), it disappears.

one thing to note however is that the event index increments on a local flag in the method, which is `true` by default.
certain events have that flag set to `false`, in which case the event index won't increment, the `it_` Session flag won't be set, no subsequent events will play, and the trigger won't disappear.

this PR implements an interface which modders can use to write events for the Interact Trigger.

## `Everest.Events.InteractTrigger.OnInteractTrigger`

this Everest Event accepts a method that returns a `bool` and has the following parameters: `InteractTrigger trigger, Player player, string eventID, ref bool progressEvent`.
the method should check the `eventID` if it matches a custom event. if there's a match, it should set up the event in some way (e.g. by adding the appropriate `CutsceneEntity` to the scene) and then return `true`. if there are no matches, it should return `false`.
the parameters are explained as follows:
- `InteractTrigger trigger` - the trigger that the player is using.
- `Player player` - the player using the trigger.
- `string eventID` - the ID for the current event.
- `ref bool progressEvent` - whether to progress to the next event (i.e. by incrementing the event index and setting the appropriate Session flag) or not. this is `true` by default.

## `[CustomInteract]` attribute

this attribute may be placed on a class that extends `CutsceneEntity` (or any class that extends `Entity`).
it accepts a string (can accept multiple) that represents the ID of the event. it's recommended that each string be unique to the mod (e.g. `myCampaignMod_ch1_theo`).
it can also take an optional property argument `ProgressEvent = `, which is a boolean and `true` by default. this indicates whether the trigger should progress to the next event and set the appropriate Session flag based on the event ID, or not.

when this attribute is in place, the class requires a `public` constructor of one of certain signatures for it to be registered. the following is checked in order:
1. `(InteractTrigger trigger, Player player, string eventID, ref bool progressEvent)`
2. `(InteractTrigger trigger, Player player, string eventID)`
3. `(Player player)`
4. `()`

alternatively, the modder may also use a generator method. Everest looks for one called `Load` by default, but a different generator may be specified with the ID in the attribute (e.g. `myCampaignMod_ch1_theo = LoadTheo`).
the matching generator method must be `public static`, return `Entity`, and have one of the following signatures:
1. `(InteractTrigger trigger, Player player, string eventID, ref bool progressEvent)`
2. `(InteractTrigger trigger, Player player, string eventID)`

note that if the `ref bool progressEvent` parameter is included, the `ProgressEvent` property of the attribute is ignored, and that is left to be handled within the method / constructor.

## Other changes

- Everest now accepts `(Player)` as a valid signature for entities marked with `[CustomEvent]`.
- the summary for `[CustomEvent]` has been edited, clarifying that it is for use with an `EventTrigger`.

---
### Additional notes

for map making tools such as Lönn, the Interact Trigger (entity name: `interactTrigger`) has the following features:
- it supports the attributes `event`, `event_2`, `event_3` ... up to `event_99`, whose numbers each correspond to the index of the event, and determines the order in which the events are played out.
- it can have up to 1 node, which indicates where to draw the input prompt graphic.

here's a simple Lönn plugin to try it out:
```lua
return {
    name = "interactTrigger",
    nodeLimits = {0, 1}, -- specifies the location of the input prompt graphic
    placements = {
        name = "interact",
        data = {
            event = "",
            event_2 = "",
            event_3 = ""
            --[[
              supports additional `event_` attributes up to 99.
              to include those, copy the placement into a text editor, and write the additional keys into it.
              then paste the modified data back into the map.
              ]]
        }
    }
}
```

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using _Decal = Celeste.Decal;
 using _EventTrigger = Celeste.EventTrigger;
+using _InteractTrigger = Celeste.InteractTrigger;
 using _LevelEnter = Celeste.LevelEnter;
 using _LevelLoader = Celeste.LevelLoader;
 using _Level = Celeste.Level;
@@ -357,6 +358,23 @@ namespace Celeste.Mod {
                 public static event TriggerEventHandler OnEventTrigger;
                 internal static bool TriggerEvent(_EventTrigger trigger, _Player player, string eventID)
                     => OnEventTrigger?.InvokeWhileFalse(trigger, player, eventID) ?? false;
+            }
+
+            public static class InteractTrigger {
+                public delegate bool TriggerInteractHandler(_InteractTrigger trigger, _Player player, string eventID, ref bool progressEvent);
+                public static event TriggerInteractHandler OnInteractTrigger;
+                internal static bool TriggerInteract(_InteractTrigger trigger, _Player player, string eventID, ref bool progressEvent) {
+                    if (OnInteractTrigger is null)
+                        return false;
+
+                    // the InvokeWhileFalse method won't accept ref arguments, so we replicate it
+                    foreach (TriggerInteractHandler handler in OnInteractTrigger.GetInvocationList()) {
+                        if (handler(trigger, player, eventID, ref progressEvent))
+                            return true;
+                    }
+
+                    return false;
+                }
             }
 
             public static class CustomBirdTutorial {
--- a/Celeste.Mod.mm/Patches/InteractTrigger.cs
+++ b/Celeste.Mod.mm/Patches/InteractTrigger.cs
@@ -1,0 +1,204 @@
+﻿using Celeste.Mod;
+using Microsoft.Xna.Framework;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Monocle;
+using MonoMod;
+using MonoMod.Cil;
+using MonoMod.InlineRT;
+using MonoMod.Utils;
+using System;
+using System.Collections.Generic;
+
+namespace Celeste {
+    class patch_InteractTrigger : InteractTrigger {
+
+        private static HashSet<string> _LoadStrings; // generated in MonoModRules.PatchInteractTriggerOnTalk
+
+        public delegate Entity InteractLoader(InteractTrigger trigger, Player player, string eventID, ref bool progressEvent);
+        public static readonly Dictionary<string, InteractLoader> InteractLoaders = new();
+
+        // MonoMod ignores this - this is only required to compile
+        public patch_InteractTrigger(EntityData data, Vector2 offset) : base(data, offset) { }
+
+        [MonoModIgnore] // don't change anything about this method...
+        [PatchInteractTriggerOnTalk] // ... except for injecting code into it via a MonoModRules patch.
+        public extern new void OnTalk(Player player);
+
+        public static bool TriggerCustomInteract(InteractTrigger trigger, Player player, string eventID, ref bool progressEvent) {
+            if (string.IsNullOrEmpty(eventID))
+                return false;
+
+            if (Everest.Events.InteractTrigger.TriggerInteract(trigger, player, eventID, ref progressEvent))
+                return true;
+
+            if (InteractLoaders.TryGetValue(eventID, out InteractLoader loader)) {
+                Entity loaded = loader(trigger, player, eventID, ref progressEvent);
+                if (loaded != null) {
+                    trigger.Scene.Add(loaded);
+                    return true;
+                }
+            }
+
+            if (!_LoadStrings.Contains(eventID)) {
+                Logger.Warn("InteractTrigger", $"Interact Event '{eventID}' does not exist!");
+            }
+
+            return false;
+        }
+    }
+}
+
+namespace MonoMod {
+    /// <summary>
+    /// Include a check for custom interact events.
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchInteractTriggerOnTalk))]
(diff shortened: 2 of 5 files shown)


JaThePlayer:
What's the purpose of this? Mods can very easily just make their own interactables (with the added benefit of being able to have any additional customization features they want), while mapper-defined interactions can instead be done via Lua Cutscenes so that you have an actual fully-capable programming language instead of a tiny subset you'd get from this. (And mods can easily expose API methods to be used from lua)

SilverDorian46:
> What's the purpose of this? Mods can very easily just make their own interactables (with the added benefit of being able to have any additional customization features they want), while mapper-defined interactions can instead be done via Lua Cutscenes so that you have an actual fully-capable programming language instead of a tiny subset you'd get from this. (And mods can easily expose API methods to be used from lua)

I thought it would be nice if the Interact trigger were usable for simpler things as well without having to remake it,
plus I just felt like it was really overlooked for its uses compared to Event triggers.

I don't know how to respond to that about Lua though, I'm not a Lua Cutscenes user

aonkeeper4:
> What's the purpose of this? Mods can very easily just make their own interactables (with the added benefit of being able to have any additional customization features they want), while mapper-defined interactions can instead be done via Lua Cutscenes so that you have an actual fully-capable programming language instead of a tiny subset you'd get from this. (And mods can easily expose API methods to be used from lua)

To me it seems foolish to only have Everest support for Event Triggers while vanilla has both Event and Interact Triggers, especially considering the similarity between the two.
I’m really not sure what you’re getting at with the Lua Cutscenes part; I don’t think this (just like Event Triggers and the `CustomEvent` API) is intended to be used by mappers. Since there is already Everest support for the non-interact case, I don't think code modders should have to resort to using Lua Cutscenes or making something custom themselves.

Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
```diff
@@ -359,6 +360,23 @@ internal static bool TriggerEvent(_EventTrigger trigger, _Player player, string
...
+
+                    // the InvokeWhileFalse method won't accept ref arguments, so we replicate it
+                    foreach (TriggerInteractHandler handler in OnInteractTrigger.GetInvocationList()) {
+                        if (handler(trigger, player, eventID, ref progressEvent))
+                            return true;
+                    }
+
+                    return false;
+                }
+            }
```
microlith57, line 378: **note:** after https://github.com/EverestAPI/Everest/pull/1059 this will need to be moved into `Patches/InteractTrigger.cs`

Celeste.Mod.mm/Patches/InteractTrigger.cs
```diff
@@ -0,0 +1,204 @@
...
+namespace MonoMod {
+    /// <summary>
+    /// Include a check for custom interact events.
+    /// </summary>
+    [MonoModCustomMethodAttribute(nameof(MonoModRules.PatchInteractTriggerOnTalk))]
+    class PatchInteractTriggerOnTalkAttribute : Attribute { }
+
+    static partial class MonoModRules {
+
+        public static void PatchInteractTriggerOnTalk(MethodDefinition method, CustomAttribute attr) {
```
microlith57, line 61: **issue:** this patch seems very strange to me (and therefore difficult to review); it would be much cleaner and easier to understand if it used `ILCursor`s instead of directly editing the instruction collections

Celeste.Mod.mm/Patches/InteractTrigger.cs
```diff
@@ -0,0 +1,204 @@
...
+                 ldarg.0
+                 ldfld    class [mscorlib]System.Collections.Generic.List`1<string> Celeste.InteractTrigger::Events
+                 ldarg.0
+                 ldfld    int32 Celeste.InteractTrigger::eventIndex
+                 callvirt instance !0 class [mscorlib]System.Collections.Generic.List`1<string>::get_Item(int32)
+                 stloc.1   // local string eventID
+                  <------- // we inject our instructions here.
+                 ldloc.1   // the switch block begins here
+                 ...
+                 br        // to an address just after the switch block
```
microlith57, line 119: **nitpick:** this is nice but i can't see what instructions you intend to inject; generally i write this kind of thing as a diff

microlith57 (commented):
i really don't like the `event_i` entitydata fields, since they'd be very very hard to implement well in a map editor. it would be much better if it was a comma-separated `events` field:
```diff
- Events.Add(data.Attr("event"));
+ var event = data.Attr("event");
+ if (!string.IsNullOrEmpty(event))
+   Events.Add(event);
  base.Collider = new Hitbox(data.Width, data.Height);
  for (int i = 2; i < 100 && data.Has("event_" + i) && !string.IsNullOrEmpty(data.Attr("event_" + i)); i++) {
    Events.Add(data.Attr("event_" + i));
  }
+ foreach (var e in data.Attr("events", "").Split(",")) {
+   if (!string.IsNullOrWhitespace(e))
+     Events.Add(e.Strip());
+ }
```
(just a sketch; and ofc this would have to be an ilpatch. sorry! :p)

Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
microlith57: #1059 is now merged!

microlith57 (changes requested):
apologies, the previous review should have been a request changes

PR #1053, opened 2025-12-29
Wipe parsing

SilverDorian46:
this PR adds some alternative ways for modders to apply their Screen Wipes to a map.
this mainly serves to give unique identifiers to the Screen Wipes, so that the game doesn't have to check for their full type names, and their classes can be freely reorganised if needed.

`Everest.Events.MapMeta.OnParseWipe` accepts a method that returns an `Action<Scene, bool, Action>` and takes the argument `string wipe`, which is the value of the current Wipe metadata being parsed.
the parameters of the returned `Action` are similar to the constructor for a Screen Wipe: `(Scene scene, bool wipeIn, Action onComplete)`.

the `[CustomWipe]` attribute should be placed on a class that extends `ScreenWipe`, and is used to give it a unique identifier. the class requires one of the following in order to be registered as a custom wipe:
- a `public static` generator method (called `Load` by default) that returns `ScreenWipe` and has the signature `(Scene scene, bool wipeIn, Action onComplete)`, or...
- a constructor with the signature `(Scene scene, bool wipeIn, Action onComplete)`.


Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
```diff
@@ -404,6 +404,22 @@ public static class SubHudRenderer {
...
+                internal static Action<Scene, bool, Action> ParseWipe(string wipe) {
+                    if (OnParseWipe is null)
+                        return null;
+
+                    foreach (ParseWipeHandler handler in OnParseWipe.GetInvocationList()) {
+                        if (handler(wipe) is { } wipeLoader)
+                            return wipeLoader;
+                    }
+
+                    return null;
```
microlith57: isn't this equivalent to `OnParseWipe.InvokeWhileNull`?

Celeste.Mod.mm/Mod/Entities/CustomWipeAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// A list of unique identifiers for this Screen Wipe.
+    /// </summary>
+    public string[] IDs;
+
+    /// <summary>
+    /// Mark this renderer as a custom <see cref="ScreenWipe"/> with an identifier.<br/>
+    /// If there is no match, then the full type name of the Screen Wipe is checked for.
+    /// </summary>
+    /// <param name="ids">A list of unique identifiers for this Screen Wipe.</param>
```
microlith57: nitpick: might be good to note that this follows the `ID[=MethodName]` pattern!

Celeste.Mod.mm/Patches/AreaData.cs
```diff
@@ -14,6 +14,8 @@
 namespace Celeste {
     public class patch_AreaData : AreaData {
 
+        public static readonly Dictionary<string, Action<Scene, bool, Action>> WipeLoaders = new();
```
microlith57, line 17: suggestion: can the `Action<Scene, bool, Action>` be a named `delegate` type instead?

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
```diff
@@ -161,10 +161,16 @@ public void ApplyTo(patch_AreaData area) {
                 area.ColorGrade = ColorGrade;
 
             if (!string.IsNullOrEmpty(Wipe)) {
-                Type type = Assembly.GetEntryAssembly().GetType(Wipe);
-                ConstructorInfo ctor = type?.GetConstructor(new Type[] { typeof(Scene), typeof(bool), typeof(Action) });
-                if (type != null && ctor != null) {
-                    area.Wipe = (scene, wipeIn, onComplete) => ctor.Invoke(new object[] { scene, wipeIn, onComplete });
+                string wipeStr = Wipe;
+                if (Everest.Events.MapMeta.ParseWipe(wipeStr) is { } wipeLoader
+                    || (patch_AreaData.WipeLoaders.TryGetValue(wipeStr, out wipeLoader) && wipeLoader is not null))
```
microlith57: suggestion: can this be split into an `if`/`else if`? it feels a bit cramped as a single condition

microlith57 (approved):
a few stylistic concerns but looks good!

Celeste.Mod.mm/Mod/Entities/CustomWipeAttribute.cs
SilverDorian46: should we do the same for the other attributes too?

Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
SilverDorian46: yeah, but I remember a comment saying writing it like this would "avoid dynamic dispatch" or something
so I got the impression that this would be somewhat more efficient

Celeste.Mod.mm/Mod/Entities/CustomWipeAttribute.cs
microlith57: oh, i guess i neglected to check those :p
yea if you can add wording to their docs too that'd be good!

Celeste.Mod.mm/Mod/Meta/MapMeta.cs
SilverDorian46: yep, will do

Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
microlith57: but this is called super infrequently right? is that a good tradeoff?
not a big deal though, it's not unreadable or anything, feel free to resolve

Celeste.Mod.mm/Patches/AreaData.cs
SilverDorian46: may just be my experience, but getting an `Action` and a delegate to work together like this is kind of tricky.
so I thought just using an `Action` would be easier

Celeste.Mod.mm/Mod/Everest/Everest.Events.cs
SilverDorian46: I guess it doesn't really matter in this case

microlith57 (approved):

Wartori54 (approved):
Happy to merge once conflicts are resolved!

PR #1055, opened 2026-01-01
Some fixes for Static Mover Decals

SilverDorian46:
- the Static Mover Decal Registry Handler now properly accepts the `jumpThrus` attribute.
- the Static Mover's OnShake callback now calls OnShake of each of the Decal's solids, if it has any. this ensures that entities attached to those solids have the same visual shake as well.
- the Static Mover's JumpThruChecker has a typo fixed: `(int) X + y` -> `(int) Y + y`
- a new event `EntityList.PostAwake` has been implemented. this is called at the end of `EntityList.UpdateLists()`, at which point it is exchanged for `null`. any additions to `PostAwake` during its invocation will be called the next time `UpdateLists()` is called.
- the Decal's Static Mover check on `Awake` has been delegated to the `PostAwake` event.
- there is now a public getter for `Decal.solids`.
- calls to `Destroy-`, `Disable-`, and `EnableStaticMovers` are included where appropriate.

### DIFF
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -249,7 +249,10 @@ namespace Celeste {
                         s.MoveV(v.Y, liftSpeed.Y);
                     });
                 },
-                OnShake = v => Position += v,
+                OnShake = v => {
+                    Position += v;
+                    solids.ForEach(s => s.OnShake(v));
+                },
                 OnAttach = p => {
                     p.Add(new EntityRemovedListener(() => {
                         RemoveSelf();
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -316,7 +316,7 @@ namespace Celeste {
         public override void Awake(Scene scene) {
             base.Awake(scene);
 
-            ((patch_EntityList) (object) scene.Entities).PostAwake += () => {
+            ((patch_EntityList) (object) scene.Entities).OnEndOfAwake += () => {
                 if (staticMover?.Platform == null && CoreModule.Session.AttachedDecals.Contains(string.Format("{0}||{1}||{2}", Name, Position.X, Position.Y))) {
                     RemoveSelf();
                 }
--- a/Celeste.Mod.mm/Patches/Decal.cs
+++ b/Celeste.Mod.mm/Patches/Decal.cs
@@ -301,9 +301,12 @@ namespace Celeste {
 
         public override void Awake(Scene scene) {
             base.Awake(scene);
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/Patches/Monocle/EntityList.cs
```diff
@@ -25,6 +26,8 @@ class patch_EntityList {
         /// </summary>
         public List<Entity> ToAdd => toAdd;
 
+        public event Action PostAwake;
```
microlith57: **suggestion:** can you rename this to `OnEndOfAwake`, for consistency with `Scene.OnEndOfFrame`? (just to make clear that it's the end of _this_ Awake, rather than _all_ Awakes)

microlith57 (approved):

Celeste.Mod.mm/Patches/Monocle/EntityList.cs
SilverDorian46: done 👍

microlith57 (approved):
still looks good!

Celeste.Mod.mm/Patches/Monocle/EntityList.cs
```diff
@@ -25,6 +26,8 @@ class patch_EntityList {
         /// </summary>
         public List<Entity> ToAdd => toAdd;
 
+        public event Action OnEndOfAwake;
```
Wartori54, line 29: If we are doing "post-awake" events, imo we should also do "pre-awake" events while we are at it.

Celeste.Mod.mm/Patches/Decal.cs
```diff
@@ -298,9 +315,12 @@ public void MakeOverlay() {
...
             base.Awake(scene);
-            if (staticMover?.Platform == null && CoreModule.Session.AttachedDecals.Contains(string.Format("{0}||{1}||{2}", Name, Position.X, Position.Y))) {
-                RemoveSelf();
-            }
+
+            ((patch_EntityList) (object) scene.Entities).OnEndOfAwake += () => {
+                if (staticMover?.Platform == null && CoreModule.Session.AttachedDecals.Contains(string.Format("{0}||{1}||{2}", Name, Position.X, Position.Y))) {
+                    RemoveSelf();
+                }
+            };
```
Wartori54, line 323: What is the use case that makes this valuable or required?

Wartori54 (commented):
Most of the PR looks good, just two nitpicks I'd like to argue about.

Celeste.Mod.mm/Patches/Decal.cs
SilverDorian46: I remember there was a bug that had to do with decals on moving platforms mysteriously disappearing under certain circumstances (it's been a while, so I don't remember the details right now). 
I initially wrote these changes in an attempt to fix that

PR #1005, opened 2025-10-04
Add support for nested SubMenus

ACherryJam:
Adds support for nested `SubMenu`s.

Every `SubMenu` now stores a `Parent` submenu. If the `Parent` is `null` the submenu is added to `TextMenu`, `Container`'s `Focused`, `AutoScroll` fields and `MoveSelection` method should be used. Otherwise we do the same things but for the `Parent`.
To keep parent `SubMenu` visible while not in focus, `ItemsVisible` flag was added.

Mod example: [NestedSubMenuExample.zip](https://github.com/user-attachments/files/22689263/NestedSubMenuExample.zip)

https://github.com/user-attachments/assets/5b7fd148-73e6-4eb5-a4fb-758fde2d3bde

PR #1057, opened 2026-01-03
Initial nested options implement

### LINKED ISSUE #1015: Nested options

Yes

### Describe your request

When you installed too many mods, there will be lots of options to find. There's already a search feature that allows to search between those, but you still have to look through all the options when you don't know which word should be searched. So maybe we can introduce a new feature that nest those options grouped by mods.

### Additional context

Effect:

Image

Image

Question:

1. Should it have an animation when entering a mod's options?
> If it's in-game, then no surely, but what if it's in OUI? UI switches in Overworld have an animation like "fly out and then fly in". ExtendedVariantMode has it, but when you entered its submenus away from options oui, the settings is saved somehow.  
> So it should have an animation in OUI but don't in-game?

2. Should it be implemented into Everest?
> If shouldn't, then how can we design new APIs that allow mods to implement this feature?

Saplonily:
Closes #1015

This pr also fix submenus not being expanded by its items. (It's calculated but not used somehow, I don't know much about the details 🤔)

Absolutely this will make search box a bit useless, or should we have a feature that search box can search into nested menus?

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
@@ -283,6 +283,8 @@ namespace Celeste.Mod.Core {
             }
         }
 
+        public bool NestedOptions { get; set; }
+
         // Keep in sync with https://github.com/EverestAPI/Olympus/blob/main/src/scenes/options.lua :: mirrorPreferences
         public string MirrorPreferences { get; set; } = "gb,jade,otobot,wegfan";
 
@@ -311,7 +313,7 @@ namespace Celeste.Mod.Core {
 
         [SettingIgnore]
         public int DebugRCPort { get; set; } = 32270;
-        
+
         [SettingIgnore]
         [EditorBrowsable(EditorBrowsableState.Never)]
         // This option is exclusively used internally; do not use it in mods. Use AllowDistort instead.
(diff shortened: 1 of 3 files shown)


SnipUndercover:
Doesn't this conflict with #1005?

Saplonily:
> Doesn't this conflict with #1005?

It's not the same "nest" way as it, it just focus another menu when entering a "submenu". Since we have true nested submenu supports in it, I can then base this implement on it later🤔

Saplonily:
Tried using 'true' submenus, but it looks a little ugly. I personally would prefer this previous implement.

Wartori54:
Looks like a solid new way to organize the mod options menu, I'd say that if we roll with it we should completely swap to that (so no `NestedOptions` toggle), but I can understand why people would like to have the ability to return to the current one. And definitely doesn't conflict with #1005 .

About the search option, it could still be useful regardless for overwhelmingly large mod option sections, and to search in the new initial options menu.

Given that this is a design focused PR I'd like other people to come in and share their opinion on this too.

DashingCat:
I agree with @Wartori54, in my opinion this is an improvement over the current "Mod Options" menu, and I would keep the search feature as well in both the mod list and for a specific mod options.

DashingCat (changes requested):
According to a GitHub search, there are a few mods using the `TextMenuExt` class:
- CelesteTAS
- ConsistencyTracker
- SomeSplitButtonsModule

When testing the changes with the "ConsistencyTracker" mod enabled, the mod options layout extends beyond the screen and isn't really usable (the following screenshots are not cropped):
- with nested options disabled:
image


- with nested options enabled:
image

From my testing, reverting the changes in `TextMenuExt.cs` seems to give a working result.
Would it be possible to keep the current look of nested options without changing `TextMenuExt.cs`? I'm not sure I understand why the changes in this file are needed.

Saplonily:
The change was made because I noticed that submenus do not expand to fit their items, they only calculated their own width. However this fix is incorrect and missed the indent width or other layout details. 🤔
I'll revert the changes in `TextMenuExt.cs` and open an issue for that submenu bug.

PR #1059, opened 2026-01-04
refactor events

microlith57:
- split `Events` into a partial class, split amongst the relevant classes
- add some more events

testmod build: https://github.com/microlith57/CelesteTestMod/pull/1 (check inside the workflow run)


JaThePlayer:
Can you provide a list of all the new events? All the diffs from moving existing events around are making it hard to find them.

microlith57:
@JaThePlayer please see the commit list; i split the commits per event.
the new events are `Player.OnPauseInGBJ`, `Level.OnEnd`, and `Atlas.OnGetCustomFallback`.

JaThePlayer (approved):

microlith57:
added a new event `Celeste.OnSceneTransition`, courtesy of @aonkeeper4

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -592,23 +595,27 @@ public override void End() {
 
             Scene nextScene = patch_Engine.NextScene;
 
+            bool shouldReloadPortraits = nextScene is not (LevelLoader or Pico8.Emulator or OverworldReflectionsFall);
```
Wartori54, line 605: The comments down below should get split up since they explain why these scenes are treated specially.

Wartori54 (approved):
LGTM. But lets hope no one ilhooked OnSceneTransition recently. (The ilhookviewer claims there werent any on April last year.)

Celeste.Mod.mm/Patches/Level.cs
microlith57: can you explain what you mean by that?

Wartori54: Before this change it was practically impossible to miss the comment upon seeing that `or` predicate of scenes, now it got moved and it could be missed if someone doesn't happen to scroll a bit down. I'd like to keep the part where it explains why those scenes are chosen immediately before the predicate itself (which is: those are the scenes that don't exit the map).

microlith57: reworded!

PR #1063, opened 2026-01-09
Add `CoreModeIceTileOverlay` map metadata property

SilverDorian46:
this is an effect specific to vanilla Core, which has been unusable in custom maps (without writing a code mod) until now.
if enabled, an ice graphic will be overlaid on the surface of Solid Tiles when the Core Mode is Cold.
may be useful for maps looking to emulate vanilla Core

### DIFF
--- a/Celeste.Mod.mm/Patches/LevelLoader.cs
+++ b/Celeste.Mod.mm/Patches/LevelLoader.cs
@@ -250,6 +250,7 @@ namespace Celeste {
         [MonoModIgnore] // We don't want to change anything about the method...
         [PatchLoadingThreadAddEvent] // ... except for manually manipulating the method via MonoModRules
         [PatchLoadingThreadAddSubHudRenderer]
+        [PatchLoadingThreadForMapMetadata]
         private extern void LoadingThread();
 
         private void LoadingThread_Safe() {
@@ -325,6 +326,9 @@ namespace Celeste {
             }
         }
 
+        private static bool _ShouldAddIceTileOverlay(MapData mapData)
+            => ((patch_MapData) mapData).Meta?.CoreModeIceTileOverlay ?? false;
+
     }
 }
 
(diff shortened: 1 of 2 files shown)


DashingCat (approved):

maddie480 (approved):

Celeste.Mod.mm/Patches/LevelLoader.cs
```diff
@@ -410,5 +420,51 @@ public static void PatchLoadingThreadAddEvent(ILContext context, CustomAttribute
...
+            static ILCursor EmitLdloc(ILCursor cursor, int index) {
+                return index switch {
+                    0               => cursor.Emit(OpCodes.Ldloc_0),
+                    1               => cursor.Emit(OpCodes.Ldloc_1),
+                    2               => cursor.Emit(OpCodes.Ldloc_2),
+                    3               => cursor.Emit(OpCodes.Ldloc_3),
+                    >= 4 and <= 255 => cursor.Emit(OpCodes.Ldloc_S, (byte)index),
+                    _               => cursor.Emit(OpCodes.Ldloc, index)
+                };
+            }
```
microlith57: **nitpick:** is this necessary? i thought ldlocs and such were automatically made short by cecil

microlith57 (approved):

Celeste.Mod.mm/Patches/LevelLoader.cs
SilverDorian46: I had no idea that was the case, really

SilverDorian46: removed just now, cause from what I've recently been told, the MonoMod patcher should automatically turn it into the correct opcode later down the line or something.

but apparently from what I've just been told, that only happens when it's a method with the signature `(ILContext, CustomAttribute)`
and maybe `(ILContext)` as well
(there are alternative signatures like `(MethodDefinition, CustomAttribute)` where this won't work)

anyway, this should be resolved for now

microlith57 (approved):
one last approving review, for fun

maddie480:
I said "update with rebase", thanks GitHub (since the history vanished: I hit "update branch" to run the tests one last time before merging, and that dum dum made a merge commit instead)

maddie480:
Rebased and squashed manually, no change made: https://github.com/EverestAPI/Everest/compare/b1ac854d468e419ccb49a85d62d1ca83b34a813f..06465b9b24fdf3338e4cdfe0706472c2e633cd05

PR #1079, opened 2026-01-21
Implement fast FMOD bank loading

DashingCat:
This PR implements "Fast FMOD Bank Loading" (like Fast Texture Loading, but for FMOD banks).

It relies on a non blocking bank loading API in FMOD.

The implementations loads all banks in parallel using the non blocking API, pool until all banks are loaded, and then do the same post processing of banks.

This implementation was tested against duplicated GUIDs (with duplicated bank files), which logs a warning; and with invalid bank data (with a truncated bank file), which logs an error (previously throwing an exception).

Right now, the non blocking implementation is enabled by default, and loading state is pooled every 100ms.

On my end, this reduced audio loading time when loading the "Strawberry Jam Collab", from about  6 seconds when using blocking loading, to less than 1 second when using non blocking loading.

### DIFF
--- a/Celeste.Mod.mm/Patches/Audio.cs
+++ b/Celeste.Mod.mm/Patches/Audio.cs
@@ -11,9 +11,11 @@ using MonoMod;
 using SDL2;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Threading;
 
 namespace Celeste {
     static class patch_Audio {
@@ -106,11 +108,34 @@ namespace Celeste {
 
             // Load any additional banks.
             lock (Everest.Content.Map) {
+                bool nonBlockingLoad = CoreModule.Settings.FastFMODBankLoading ?? true;
+
+                if (nonBlockingLoad) {
+                    Logger.Info("Audio", "Loading FMOD banks in parallel");
+                }
+
                 foreach (ModAsset asset in Everest.Content.Map.Values.Where(asset => asset.Type == typeof(AssetTypeBank))) {
                     if (!ingestedModBankPaths.Contains(asset.PathVirtual)) {
-                        IngestBank(asset);
+                        additionalBanksMapping.Add(asset, IngestBank(asset, nonBlockingLoad));
                     }
                 }
+
+                if (nonBlockingLoad) {
+                    // pool loading state, and do post processing (GUID ingestion)
+                    while (IsAnyBankLoading()) {
+                        Logger.Info("Audio", "Sleeping waiting for bank/sample data to be loaded");
+                        Thread.Sleep(500);
+                    }
+
+                    foreach (ModAsset asset in Everest.Content.Map.Values.Where(asset => asset.Type == typeof(AssetTypeBank))) {
+                        IngestAllGUIDs(asset.PathVirtual);
+                    }
+
+                    CacheAllBanks();
+                }
+
+                ResetBankLoadingState();
+                Logger.Info("Audio", $"Loaded {cachedBankPaths.Count} banks and {cachedModEvents.Count} events");
             }
 
             AudioInitialized = true;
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -106,11 +108,34 @@ public static void Init() {
...
+                        additionalBanksMapping.Add(asset, IngestBank(asset, nonBlockingLoad));
+                    }
+                }
+
+                if (nonBlockingLoad) {
+                    // pool loading state, and do post processing (GUID ingestion)
+                    while (IsAnyBankLoading()) {
+                        Logger.Info("Audio", "Sleeping waiting for bank/sample data to be loaded");
+                        Thread.Sleep(500);
+                    }
```
JaThePlayer: Does the main thread have to wait for all banks to be loaded before it can ingest *any* guid files?
Seems like instead there could be a list of remaining `(ModAsset asset, Bank bank)` pairs that haven't been loaded yet which would be checked in the loop (instead of IsAnyBankLoading()). When `bank.getLoadingState != FMOD_STUDIO_LOADING_STATE_LOADING`, the pair gets removed from that list, and the result is processed immediately (FMOD_STUDIO_LOADING_STATE_ERROR - Bank.unload(), ERR_EVENT_ALREADY_LOADED - log, FMOD_STUDIO_LOADING_STATE_LOADED - IngestAllGUIDs(asset)). Should allow in a bit more parallelisation (if fmod allows for this ofc)

Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -147,14 +172,64 @@ public static void Unload() {
...
+        }
+
+        private static void CacheAllBanks() {
+            foreach ((ModAsset asset, Bank bank) in additionalBanksMapping) {
+                if (!conflicts.Contains(asset)) {
+                    patch_Banks.ModCache[asset] = bank;
+                    bank.getID(out Guid id);
+                    cachedBankPaths[id] = $"bank:/mods/{asset.PathVirtual["Audio/".Length..]}";
+                }
+            }
```
JaThePlayer: Quite a bit of duplicated code between the non-blocking and blocking variants, would be good to split stuff up to more methods which can be reused for both paths (this goes for more places in this PR than just this - error logging for conflicting events for example)

Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -147,14 +172,64 @@ public static void Unload() {
...
+        private static void ResetBankLoadingState() {
+            additionalBanksMapping.Clear();
+            conflicts.Clear();
+        }
+
         /// <summary>
         /// Loads an FMOD Bank from the given asset.
         /// </summary>
-        public static Bank IngestBank(ModAsset asset) {
+        public static Bank IngestBank(ModAsset asset, bool nonBlockingLoad = false) {
```
JaThePlayer: This is a binary-breaking change to a public api. While very unlikely someone is using it, I think adding a `public static Bank IngestBank(ModAsset asset) => IngestBank(ModAsset asset, false)` wouldn't hurt. (at which point the default value for this overload doesn't make sense and should be removed)

Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -147,14 +172,64 @@ public static void Unload() {
             ready = false;
         }
 
+        // internal fields relecting bank loading state
+        internal readonly static Dictionary<ModAsset, Bank> additionalBanksMapping = new Dictionary<ModAsset, Bank>();
+        internal readonly static List<ModAsset> conflicts = new List<ModAsset>();
```
JaThePlayer: Debateable whether these need to be fields and couldn't instead be passed around to places which need them as arguments, but I'll hold off on further analysis before the rework mentioned in an earlier comment

JaThePlayer (changes requested):
From the remarks at https://documentation.help/fmod-studio-api/FMOD_Studio_System_LoadBankFile.html
> Failed asynchronous banks should be released by calling [Studio::Bank::unload](https://documentation.help/fmod-studio-api/FMOD_Studio_Bank_Unload.html).

Celeste.Mod.mm/Patches/Audio.cs
JaThePlayer: (The potentially-alternative question is whether guid ingestion could be parallelised)

DashingCat:
Thanks @JaThePlayer for the feedback, I addressed it in the latest commit.

Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -106,11 +108,25 @@ public static void Init() {
 
             // Load any additional banks.
             lock (Everest.Content.Map) {
+                bool nonBlockingLoad = CoreModule.Settings.FastFMODBankLoading ?? true;
+
+                if (nonBlockingLoad) {
+                    Logger.Info("Audio", "Loading FMOD banks in parallel");
+                }
+
+                List<KeyValuePair<ModAsset, Bank>> additionalBanksMapping = new List<KeyValuePair<ModAsset, Bank>>();
```
JaThePlayer: `KeyValuePair<ModAsset, Bank>` could be simplified to a tuple like this: `(ModAsset, Bank)`

Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -106,11 +108,25 @@ public static void Init() {
...
-                        IngestBank(asset);
+                        additionalBanksMapping.Add(new KeyValuePair<ModAsset, Bank>(asset, IngestBank(asset, nonBlockingLoad)));
                     }
                 }
+
+                if (nonBlockingLoad) {
+                    PoolAndLoadBanksAndEvents(additionalBanksMapping);
+                }
+
+                additionalBanksMapping.Clear();
```
JaThePlayer: Clearing is not necessary here, since the method is ending here and the list will be GC'able soon anyway.

Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -147,14 +163,44 @@ public static void Unload() {
...
+                        additionalBanksMapping.RemoveAt(i);
+                    } else if (loadingState != LOADING_STATE.LOADING) {
+                        // assume loading was done correctly
+                        PostProcessBank(bank, asset);
+                        additionalBanksMapping.RemoveAt(i);
+                    }
+                }
+
+                // wait before pooling loading state again
+                Logger.Info("Audio", "Sleeping waiting for bank/sample data to be loaded");
```
JaThePlayer: This will cause significant log spam if audio loading takes a while, I'd remove this log personally.

Celeste.Mod.mm/Patches/Audio.cs
```diff
@@ -147,14 +163,44 @@ public static void Unload() {
...
+                    RESULT result = bank.getLoadingState(out LOADING_STATE loadingState);
+
+                    if (CheckForBankLoadingErrors(result, asset)) {
+                        bank.unload(); // failed asynchronous banks should be released according to FMOD documentation
+                        additionalBanksMapping.RemoveAt(i);
+                    } else if (loadingState != LOADING_STATE.LOADING) {
+                        // assume loading was done correctly
+                        PostProcessBank(bank, asset);
+                        additionalBanksMapping.RemoveAt(i);
+                    }
```
JaThePlayer, line 175: I believe this logic is not quite correct here for failed banks.
CheckForBankLoadingErrors will return false if `loadingState == ERROR && result == RESULT.FMOD_ERR_VERSION` for example, at which point `loadingState != LOADING_STATE.LOADING` is true and the bank is assumed to be correct, while it very much isn't...
... except `result.CheckFMOD()` called in `CheckForBankLoadingErrors` would instead throw an exception in that case, and since there's no try-catch anywhere here, all of audio loading would fail completely due to just a single incorrect bank... which also seems to be the case in the original non-blocking version of this code???
I think that `result.CheckFMOD()` call could be changed to some alternative variant of that method which just logs the error instead of throwing, then it should be safe to call `bank.unload()` and continue loading the remaining banks...

JaThePlayer (changes requested):
I've noticed that the `IngestNewBanks` method in this class has an exact copy-paste of the last bit of `Init()` that's being changed in this PR, so it would be good to move those changes to `IngestNewBanks` and make `Init` call that method instead.

DashingCat:
Thanks again for the feedback, I addressed it in the latest commit.

This PR now introduces a behavioral change: failing to load a bank no longer throws an exception, but logs an error instead.

JaThePlayer (approved):
My testing in SJ, seems to work fine, including with invalid banks:
Pre-PR:  `- AUDIO LOAD: 6903ms`
Post-PR:  `- AUDIO LOAD: 1014ms`

maddie480 (approved):

PR #999, opened 2025-09-27
Various `DialogCutsceneTrigger` additions

SnipUndercover:
- Prevents the dialog cutscene trigger from spawning if `onlyOnce` (now `noRespawnAfterUse`) is set
- Allows the dialog cutscene to be triggered multiple times
  - Useful if someone wants to the dialog cutscene to be retriggered
- Allows delaying the dialog cutscene until after the intro state is complete
  - Makes the dialog transition smoother if the trigger is placed on the spawnpoint
  
The two new entity properties will need support from map-maker programs like Lönn.

PR #1081, opened 2026-01-29
Allow concurrent cutscenes

DashingCat:
Follow up on #999, this PR removes the check preventing multiple cutscenes from running concurrently.

Running multiple cutscenes concurrently is usually not desirable for a player/mapper, but it is currently used by TASes.
The consensus right now is to allow this behavior.

This change was tested against [Monika's D-Sides - Trust chapter TAS](https://github.com/fishmcmuffins/D-Sides-TAS/blob/5dfd6183add29cdb392b0041e4ebe778995ad458/9D/9D_10_Trust.tas):
- without this change, there is a desync in the room `hh-01`
- with this change, the TAS runs normally

### DIFF
--- a/.github/tas-check/run-locally.sh
+++ b/.github/tas-check/run-locally.sh
@@ -11,14 +11,14 @@ set -xeo pipefail
 
 case "$1" in
     "Celeste")
-        TAS_URL="https://github.com/VampireFlower/CelesteTAS/archive/0747f018e7c92abcd1bf900bd02d24de2be190e4.zip"
-        TAS_PATH="CelesteTAS-0747f018e7c92abcd1bf900bd02d24de2be190e4/0 - 100%.tas"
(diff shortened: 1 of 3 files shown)


SnipUndercover (changes requested):
As described under f6b06c322dca6928ee7b1af532cb459881a13b3e, this change was made to prevent the same Dialog Cutscene Trigger from being activated multiple times via something like Crystalline's Trigger Trigger.
Previously this was not the case, as the cutscene trigger just immediately removed itself after being activated. This check was designed to prevent this situation and bring the behavior more in line with how it was before this PR.

The consensus was to allow multiple *separate* cutscenes to play concurrently, but not the same one. I guess this could be accomplished by setting a flag for the duration of the cutscene and simply bailing out early if the flag is set. While this would also be useful for mappers, I'm not 100% sure whether this would be something that should be in Everest.
*(it probably is but i dunno)*

DashingCat:
I am unable to reproduce the issue with "Crystalline's Trigger Trigger" so far, so I'm marking this PR as draft for now.

DashingCat:
Closing in favor of #1107.

PR #1082, opened 2026-01-31
fix crash when reloading inside dream block

microlith57:
...and also any other crashes related to emitting dust while Engine.Scene is not a Level

### DIFF
--- a/.github/tas-check/run-locally.sh
+++ b/.github/tas-check/run-locally.sh
@@ -11,14 +11,14 @@ set -xeo pipefail
 
 case "$1" in
     "Celeste")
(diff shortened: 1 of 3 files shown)


DashingCat (approved):
I was able to reproduce the crash without the changes applied, and could not reproduce it after applying the changes.

maddie480 (changes requested):
There are merge conflicts on the TAS check, and it's ~~possibly~~ definitely my fault 😅

Wartori54 (approved):

SnipUndercover (approved):

PR #1088, opened 2026-02-12
Add `TagsExt.FreezeFrameUpdate`

aonkeeper4:
This PR adds a new tag (`TagsExt.FreezeFrameUpdate`) to allow entities to update during freeze frames. I'm not sure whether the logic I have is all there needs to be, but so far it is all functional.


Celeste.Mod.mm/Patches/Monocle/Engine.cs
```diff
@@ -234,6 +236,17 @@ public static void PatchEngineUpdate(ILContext context, CustomAttribute attrib)
...
+            cursor.GotoNext(MoveType.Before,
+                instr => instr.MatchLdsfld("Monocle.Engine", "FreezeTimer"),
+                instr => instr.MatchCall("Monocle.Engine", "get_RawDeltaTime"),
+                instr => instr.MatchSub(),
+                instr => instr.MatchLdcR4(0f),
+                instr => instr.MatchCall("System.Math", "Max"),
+                instr => instr.MatchStsfld("Monocle.Engine", "FreezeTimer"));
+            cursor.EmitLdarg0();
+            cursor.EmitLdfld(f_scene);
+            cursor.EmitCallvirt(m_Scene_FreezeFrameUpdate);
```
Wartori54, line 251: Would be nice to add a comment on top of this explaining what is being added, just like the other three comments in this method.

Celeste.Mod.mm/Patches/Monocle/Scene.cs
```diff
@@ -48,5 +49,22 @@ public IEnumerable<Entity> FindEntitiesWithSid(string sid) {
...
+        // todo: allow some renderers to update?
+        public virtual void FreezeFrameUpdate() {
+            if (!Paused) {
+                foreach (patch_Entity entity in this[TagsExt.FreezeFrameUpdate]) {
+                    entity._PreUpdate();
+                    if (entity.Active)
+                    {
+                        entity.Update();
+                    }
+                    entity._PostUpdate();
```
Wartori54: Not entirely sure about adding `_PreUpdate` and `_PostUpdate` calls here. When `Scene is Level`, those are only called when a "normal" update goes through (an update that updates all entities and renderers, that is, when `Scene.Update()` is called from `Level.Update()`).

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -106,6 +106,24 @@ public static void RegisterLoadOverride(Level level, LoadOverride loadOverride)
...
+            
+            base_FreezeFrameUpdate();
+            
+            foreach (PostUpdateHook component in Tracker.GetComponents<PostUpdateHook>()) {
+                if (component.Entity.Active && component.Entity.TagCheck(TagsExt.FreezeFrameUpdate)) {
+                    component.OnPostUpdate();
+                }
+            }
+            Everest.Events.Level.AfterFreezeFrameUpdate(this);
+        }
```
Wartori54, line 159: A (very niche) edge case is missing here, `Level` allows for entities with `Tags.PauseUpdate` to be updated when `Paused` is `true`, if an entity has both `Tags.PauseUpdate` and `TagsExt.FreezeFrameUpdate` it will not be updated here.

Celeste.Mod.mm/Patches/Tags.cs
```diff
@@ -20,5 +21,9 @@ public static class TagsExt {
         /// </summary>
         public static BitTag SubHUD;
 
+        /// <summary>
+        /// Tag to be used for entities that should update during freeze frames.
```
Wartori54: Entities will be updated regardless of the internal state of the `Scene`. Which when `Scene is Level` it means that those may be updated in "weird" times: for example, if freeze frames occur during a screen transition, entities with this tag and without `Tags.TransitionUpdate` will be updated. This is fair behaviour, but I think it should be noted in the comment. Something along "Tag to be used for entities that should update during freeze frames, regardless of the state of the `Scene`." would work for me.

Furthermore and connected to a previous comment, you should also add here that for entities with this tag to be updated when `Paused` is true (and the scene is a level) the tag `PausedUpdate` is also required for the update to go through.

Wartori54 (changes requested):
Looks like a nice tag to add, just some polish missing.

aonkeeper4:
I've applied your changes, but I felt unsure about having the tag ignore `Frozen` and `Transitioning`, so I updated `Level.FreezeFrameUpdate` to more closely match `Level.Update` in that regard. This is reflected in the documentation changes.

Wartori54:
Yeah, that works too.

Wartori54 (approved):

Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -797,6 +849,20 @@ internal static void BeforeUpdate(_Level level)
...
+                public static event Action<_Level> OnBeforeFreezeFrameUpdate;
+                internal static void BeforeFreezeFrameUpdate(_Level level)
+                    => OnBeforeFreezeFrameUpdate?.Invoke(level);
+                
+                /// <summary>
+                /// Called at the very end of <see cref="patch_Level.FreezeFrameUpdate"/>.
+                /// </summary>
+                public static event Action<_Level> OnAfterFreezeFrameUpdate;
+                internal static void AfterFreezeFrameUpdate(_Level level)
+                    => OnAfterFreezeFrameUpdate?.Invoke(level);
```
microlith57, line 865: **praise:** yay events! thank you :3

Celeste.Mod.mm/Patches/Tags.cs
microlith57, line 7: **idea:** this is perhaps a bit out of scope for this PR but it would be cool to have fields
```cs
public static BitTag SubHUD, PauseUpdate;
```
on `patch_Tags` (with appropriate docs ofc), set at the same time as the ones in `TagsExt`.. it would be so nice ....
(and perhaps in future we can relink the `TagsExt` ones to `Tags` and get rid of `TagsExt` entirely?)


and also **idea:** the `Tags` tags really ought to have xmldocs too lmao

Celeste.Mod.mm/Patches/Monocle/Scene.cs
```diff
@@ -23,8 +24,7 @@ class patch_Scene : Scene {
         public new bool OnInterval(float interval, float offset) {
             return Math.Floor(((double) TimeActive - offset - Engine.DeltaTime) / interval) < Math.Floor(((double) TimeActive - offset) / interval);
         }
-
-                
+
```
microlith57, line 27: **typo:** why so indented?
```suggestion

```

microlith57 (approved):

PR #1089, opened 2026-02-20
Fix threading issue of `Atlas.FallbackStack`, and introduce overloads with `fallback` parameter

Saplonily:
This pr changes the `Atlas.FallbackStack` implementation into `ThreadLocal<T>`, since the expected usage of it is that, pushing a fallback, get something from `Atlas` in this context, and then pop it, which make sense to make it thread local.
If we're using `ConcurrentStack`, then one thread may get a fallback that pushed from another thread, which is not expected.

---
There's an optional commit that marks stack operation methods as obsolete, and add some overloads that accept a `fallback` parameter, because this is a much easier way to reach that.
```diff cs
namespace Monocle;

public class Atlas
{
+   [Obsolete(MessageFallbackStackObsolete)]
    public void PushFallback(MTexture fallback)

+   [Obsolete(MessageFallbackStackObsolete)]
    public MTexture PopFallback()

+   public MTexture GetDefaultFallback();
+   public MTexture GetAtlasSubtexturesAt(string key, int index, MTexture fallback)
+   public List<MTexture> GetAtlasSubtextures(string key, MTexture fallback)
+   public MTexture GetWithFallback(string id, MTexture fallback) 
}
```

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/Atlas.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/Atlas.cs
@@ -9,6 +9,7 @@ using MonoMod;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Threading;
 using System.Xml;
 using Logger = Celeste.Mod.Logger;
 
@@ -24,9 +25,9 @@ namespace Monocle {
         /// The internal string-MTexture dictionary.
         /// </summary>
         public Dictionary<string, MTexture> Textures => textures;
-        private Dictionary<string, string> links = new Dictionary<string, string>();
+        private Dictionary<string, string> links;
         private Dictionary<string, List<MTexture>> orderedTexturesCache;
-        private Stack<MTexture> FallbackStack;
+        private ThreadLocal<Stack<MTexture>> FallbackStack;
 
         public MTexture DefaultFallback;
 
@@ -35,6 +36,15 @@ namespace Monocle {
         public string RelativeDataPath;
         public string[] DataPaths;
         public AtlasDataFormat? DataFormat;
+
+        public extern void orig_ctor();
+
+        [MonoModConstructor]
+        public void ctor() {
+            links = new Dictionary<string, string>();
+            FallbackStack = new(() => new());
+            orig_ctor();
+        }
 
         [MonoModReplace]
         private static void ReadAtlasData(Atlas _atlas, string path, AtlasDataFormat format) {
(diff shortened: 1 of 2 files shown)


maddie480 (approved):

DashingCat (approved):

microlith57 (approved):
LGTM pending conflict resolution

Saplonily:
Conflict resolved. But there's a change that `MTexture GetWithFallback(string id, MTexture fallback)` will not log `does not exist` messages, because I noticed that in #1059, if a fallback is provided from `OnGetCustomFallback` then the message is not logged as if it's "found".

AAA1459:
I heard that this PR has a new `GetAtlasSubtextures` overload so there are two `GetAtlasSubtextures` now.

MMHook will automatically generate two new references for these methods with the same name. and then `On.Monocle.Atlas.GetAtlasSubtextures` is no longer exists and causing SMH+ not work (although not a big deal like just need an update)

I'm just curious if the new method could be renamed to `GetAtlasSubtexturesWithDefault`-like as there does a `fallback`.
33eline

JaThePlayer (changes requested):
I wouldn't mark the stack-based methods as obsolete, as there are valid usecases for them, such as providing fallbacks when calling into vanilla code (inherited from a base entity class for example), where it's impossible to switch to the new methods.

Saplonily:
> I heard that this PR has a new `GetAtlasSubtextures` overload so there are two `GetAtlasSubtextures` now.
> 
> MMHook will automatically generate two new references for these methods with the same name. and then `On.Monocle.Atlas.GetAtlasSubtextures` is no longer exists and causing SMH+ not work (although not a big deal like just need an update)
> 
> I'm just curious if the new method could be renamed to `GetAtlasSubtexturesWithDefault`-like as there does a `fallback`. 33eline

yeah surely, I haven't realized that MMHook will be broken. I'd prefer a `GetAtlasSubtexturesWithFallback` name.

Saplonily:
> I wouldn't mark the stack-based methods as obsolete, as there are valid usecases for them, such as providing fallbacks when calling into vanilla code (inherited from a base entity class for example), where it's impossible to switch to the new methods.

I made a search that there're two usages in Everest, one at `GetAtlasSubtextures` that wrap a `orig_GetAtlasSubtextures` call, but I can't find any fallback accesses there 🤔, another at `HasFrames` where `GetAtlasSubtexturesAt` in it accessed, so this pr `MonoModReplace`d it to use a `null` fallback provided version `GetAtlasSubtexturesAt`.

PR #1092, opened 2026-03-08
Flush modsettings stream before writing to file

DashingCat:
This PR flushes the `StreamWriter` before flushing the `FileStream`.

From my testing, not flushing the `StreamWriter` before flushing the `FileStream` causes data to remain in the `StreamWriter`, which needs to be written after the `FileStream` flush.

Flushing the `StreamWriter` before flushing the `FileStream` leads to one less `write` system calls, and may fix the issue with `modsettings` files being truncated/corrupted.


microlith57 (approved):
lgtm

SnipUndercover (approved):

PR #1094, opened 2026-03-20
Add customization for room colors on the debug map

aonkeeper4:
This PR allows mappers to specify custom room colors on the debug map, instead of only being limited to the small number of hardcoded ones. This is done by adding a new `color` field to `level` elements, which will be used for the room color in the debug map if present.

### DIFF
--- a/Celeste.Mod.mm/Patches/LevelTemplate.cs
+++ b/Celeste.Mod.mm/Patches/LevelTemplate.cs
@@ -12,53 +12,45 @@ using System.Runtime.CompilerServices;
 namespace Celeste.Editor {
     class patch_LevelTemplate : LevelTemplate {
         // expose this private field to our patch
-        private static Color[] fgTilesColor;
+        private static new Color[] fgTilesColor;
+
+        // custom room color
+        // if null, `EditorColorIndex` will be used to determine the room color
+        private Color? editorColor;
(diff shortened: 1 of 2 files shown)


microlith57 (approved):
yea

Celeste.Mod.mm/Patches/LevelTemplate.cs
```diff
@@ -10,26 +11,41 @@
 
 namespace Celeste.Editor {
     class patch_LevelTemplate : LevelTemplate {
+        // expose this private field to our patch
+        private static new Color[] fgTilesColor;
+
+        // custom room color
+        // if null, `EditorColorIndex` will be used to determine the room color
+        public Color? CustomEditorColor;
```
Wartori54, line 19: Is this public intentionally?

Wartori54 (approved):
LGTM, only a single non-blocking nitpick.

Celeste.Mod.mm/Patches/LevelTemplate.cs
aonkeeper4: My thought was that if `EditorColorIndex` is public, this should probably also be public. I don't mind changing it to be internal though.

Wartori54: It looks like a backing field so using `private` would make sense here.

PR #1096, opened 2026-03-24
Add TemporarySpriteBatch

SnipUndercover:
Adds a helper `TemporarySpriteBatch` type to easily interrupt an existing `SpriteBatch`, optionally swap in a `RenderTarget2D`, and start a new `SpriteBatch`.
When disposed, the new `SpriteBatch` is ended, the old `RenderTarget2D` is restored and the previous `SpriteBatch` is resumed.

~~Also bumps the language version to C# 13, as this is necessary for ref structs to implement interfaces.~~
~~We already need .NET 9 to compile Everest, so we might as well make use of C# 13.~~

**EDIT**: Implicitly implementing `IDisposable` lets us go back to `<LangVersion>` 12, so the version bump isn't necessary. Ideally we should implement `IDisposable` but this will work for now, until we jump to .NET 10.

### DIFF
--- a/Celeste.Mod.mm/Celeste.Mod.mm.csproj
+++ b/Celeste.Mod.mm/Celeste.Mod.mm.csproj
@@ -5,8 +5,6 @@
     <AssemblyName>Celeste.Mod.mm</AssemblyName>
     <RootNamespace>Celeste</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <!-- C# 13 was introduced with the .NET 9 SDK, and we need .NET 9 to compile MonoMod anyway -->
-    <LangVersion>13</LangVersion>
     <ShouldIncludeNativeLua>false</ShouldIncludeNativeLua>
   </PropertyGroup>
 
--- a/Celeste.Mod.mm/Celeste.Mod.mm.csproj
+++ b/Celeste.Mod.mm/Celeste.Mod.mm.csproj
@@ -5,7 +5,8 @@
     <AssemblyName>Celeste.Mod.mm</AssemblyName>
     <RootNamespace>Celeste</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <LangVersion>11</LangVersion>
+    <!-- C# 13 was introduced with the .NET 9 SDK, and we need .NET 9 to compile MonoMod anyway -->
+    <LangVersion>13</LangVersion>
     <ShouldIncludeNativeLua>false</ShouldIncludeNativeLua>
   </PropertyGroup>
 
--- a/Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
@@ -1,8 +1,8 @@
-﻿using System;
-using Microsoft.Xna.Framework;
+﻿using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using Monocle;
 using MonoMod.Utils;
+using System.Buffers;
 using System.Diagnostics.CodeAnalysis;
 
 namespace Celeste.Mod.Helpers;
@@ -25,6 +25,11 @@ namespace Celeste.Mod.Helpers;
 /// <seealso cref="TemporarySpriteBatchBuilder"/>
 public ref struct TemporarySpriteBatch
 {
+    // we don't really expect the RenderTargetBinding[] array size to exceed 1,
+    // so a max of 16 seems reasonable
+    private static readonly ArrayPool<RenderTargetBinding> _renderTargetPool
+        = ArrayPool<RenderTargetBinding>.Create(0x10, 50);
+
     /// <summary>
     ///   The <see cref="SpriteSortMode"/> of this <see cref="TemporarySpriteBatch"/>.
     /// </summary>
@@ -179,7 +184,7 @@ public ref struct TemporarySpriteBatch
             int renderTargetCount = graphicsDevice.GetRenderTargetsNoAllocEXT(null);
             if (renderTargetCount > 0)
             {
-                PreviousRenderTargets = new RenderTargetBinding[renderTargetCount];
+                PreviousRenderTargets = _renderTargetPool.Rent(renderTargetCount);
                 graphicsDevice.GetRenderTargetsNoAllocEXT(PreviousRenderTargets);
             }
             CurrentRenderTarget = renderTarget;
@@ -220,6 +225,8 @@ public ref struct TemporarySpriteBatch
             PreviousRasterizerState,
             PreviousCustomEffect,
             PreviousTransformMatrix);
+
+        _renderTargetPool.Return(PreviousRenderTargets, clearArray: true);
     }
 
     private static void GetSpriteBatchFields(
--- a/Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
@@ -1,0 +1,244 @@
+﻿using System;
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
+using Monocle;
+using MonoMod.Utils;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Celeste.Mod.Helpers;
+
+/// <summary>
+///   A temporary <see cref="SpriteBatch"/>.
+/// </summary>
+/// <remarks>
+///   When constructed, the current <see cref="SpriteBatch"/> properties and <see cref="RenderTarget2D"/> are preserved.
+///   Then, the <see cref="SpriteBatch"/> is ended, the <see cref="RenderTarget2D"/> is swapped if necessary,
+///   and finally the <see cref="SpriteBatch"/> is restarted with the new properties.<br/>
+///   When disposed, the previous <see cref="SpriteBatch"/> properties and <see cref="RenderTarget2D"/> are restored.
+///   <br/>
+///   Useful when interrupting a <see cref="SpriteBatch"/> mid-render to, for example, render a specific entity to a
+///   temporary <see cref="RenderTarget2D"/> while applying a custom shader, all while preserving the previous
+///   configuration.<br/>
+///   <br/>
+///   Note: Restarting a spritebatch flushes it to the GPU, which is costly.
+/// </remarks>
+/// <seealso cref="TemporarySpriteBatchBuilder"/>
+public ref struct TemporarySpriteBatch : IDisposable
+{
+    /// <summary>
+    ///   The <see cref="SpriteSortMode"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    public readonly SpriteSortMode CurrentSortMode;
+
+    /// <summary>
+    ///   The <see cref="BlendState"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    [NotNull]
+    public readonly BlendState CurrentBlendState;
+
+    /// <summary>
+    ///   The <see cref="SamplerState"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    [NotNull]
+    public readonly SamplerState CurrentSamplerState;
+
+    /// <summary>
+    ///   The <see cref="DepthStencilState"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    [NotNull]
+    public readonly DepthStencilState CurrentDepthStencilState;
+
+    /// <summary>
+    ///   The <see cref="RasterizerState"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    [NotNull]
+    public readonly RasterizerState CurrentRasterizerState;
+
+    /// <summary>
+    ///   The custom <see cref="Effect"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    [MaybeNull]
+    public readonly Effect CurrentCustomEffect;
+
+    /// <summary>
+    ///   The transformation <see cref="Matrix"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    public readonly Matrix CurrentTransformMatrix;
+
+    /// <summary>
+    ///   The <see cref="RenderTarget2D"/> swapped in for the duration of this <see cref="TemporarySpriteBatch"/>
+    ///   or <c>null</c> to draw to the screen if <see cref="HasRenderTarget"/> is <c>true</c>; else always <c>null</c>.
+    /// </summary>
+    [MaybeNull]
+    public readonly RenderTarget2D CurrentRenderTarget;
+
+
+    /// <summary>
+    ///   The <see cref="SpriteSortMode"/> that was used prior to the start of this
+    ///   <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    public readonly SpriteSortMode PreviousSortMode;
+
+    /// <summary>
+    ///   The <see cref="BlendState"/> that was used prior to the start of this
+    ///   <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    [NotNull]
+    public readonly BlendState PreviousBlendState;
+
+    /// <summary>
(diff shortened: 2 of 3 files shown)


Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
```diff
@@ -0,0 +1,244 @@
...
+    private static void GetSpriteBatchFields(
+        out SpriteSortMode sortMode,
+        [NotNull] out BlendState blendState,
+        [NotNull] out SamplerState samplerState,
+        [NotNull] out DepthStencilState depthStencilState,
+        [NotNull] out RasterizerState rasterizerState,
+        [MaybeNull] out Effect customEffect,
+        out Matrix transformMatrix)
+    {
+        // life would be good if we could just access these directly...
```
microlith57, line 233: **query:** why can't we? surely monomod has some way do do this right?

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatchBuilder.cs
```diff
@@ -0,0 +1,281 @@
...
+///   A <see cref="TemporarySpriteBatch"/> builder.
+/// </summary>
+/// <remarks>
+///   This class lets users configure the creation of a new <see cref="TemporarySpriteBatch"/>, which allows
+///   users to interrupt an existing <see cref="SpriteBatch"/>, optionally swap in a <see cref="RenderTarget2D"/>
+///   and restart the <see cref="SpriteBatch"/> with custom properties.<br/>
+///   When done, the <see cref="SpriteBatch"/> is ended, the previous <see cref="RenderTarget2D"/> is restored and the
+///   old <see cref="SpriteBatch"/> is resumed.<br/>
+///   <br/>
+///   Note: Restarting a spritebatch flushes it to the GPU, which is costly.
```
microlith57: **nitpick:** this cost is not actually that significant (since we are flushing spritebatches way too much already), it would probably be good to note that here to avoid scaring people away from this

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
```diff
@@ -0,0 +1,244 @@
...
+using MonoMod.Utils;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Celeste.Mod.Helpers;
+
+/// <summary>
+///   A temporary <see cref="SpriteBatch"/>.
+/// </summary>
+/// <remarks>
+///   When constructed, the current <see cref="SpriteBatch"/> properties and <see cref="RenderTarget2D"/> are preserved.
```
microlith57, line 15: **nitpick:** this comment doesn't mention _how_ to construct these, probably good to have a `see`

microlith57 (commented):
initial comments, will review more thoroughly later!

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
SnipUndercover: There is a `<seealso cref="TemporarySpriteBatchBuilder"/>` on line 25.

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
SnipUndercover: I am using `DynamicData` here.
It'd be great if we didn't have to resort to `DynamicData` or reflection and just access fields directly.

microlith57: yes that's what i mean, i don't know if monomodlinkto does what we want but if not i would expect a similar thing exists
i guess in the worst case we can always write the body of this method in il

SnipUndercover: The JIT would still check access permissions, unless we `[IgnoreAccessChecksTo("FNA")]`, but I don't feel like this is a good idea.

Saplonily:
Since we've already on .net 8, what about enable null context for this file instead of using these attributes?

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        GraphicsDevice graphicsDevice = Engine.Graphics.GraphicsDevice;
+        if (hasRenderTarget)
+        {
+            int renderTargetCount = graphicsDevice.GetRenderTargetsNoAllocEXT(null);
+            if (renderTargetCount > 0)
+            {
+                PreviousRenderTargets = new RenderTargetBinding[renderTargetCount];
+                graphicsDevice.GetRenderTargetsNoAllocEXT(PreviousRenderTargets);
+            }
```
JaThePlayer, line 180: The use of the NoAlloc variant of GetRenderTargets is useless here, since we're allocating a new `RenderTargetBinding[]` each time anyway, maybe worth switching to `ArrayPool<T>` here to avoid that allocation.

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
```diff
@@ -0,0 +1,244 @@
...
+            CurrentTransformMatrix);
+    }
+
+    /// <summary>
+    ///   End this <see cref="TemporarySpriteBatch"/>, restore the previous render targets if necessary, and restore
+    ///   the previous <see cref="SpriteBatch"/> properties.
+    /// </summary>
+    public void Dispose()
+    {
+        ObjectDisposedException.ThrowIf(!Active, typeof(TemporarySpriteBatch));
```
JaThePlayer: Generally, Dispose doesn't throw exceptions like this:
From [msdocs](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) "To help ensure that resources are always cleaned up appropriately, a Dispose method should be idempotent, such that it's callable multiple times without throwing an exception. Furthermore, subsequent invocations of Dispose should do nothing."

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
JaThePlayer: [Everest already applies patches to FNA](https://github.com/EverestAPI/Everest/tree/dev/Celeste.Mod.mm/Patches/FNA), I think it wouldn't be too crazy to do this, especially since this api is by definition going to be called on a hot path. I could see it being useful for mods to be able to access the current sprite batch state in general as well. (Doesn't Everest already have a different place (Parallax?) where it gets this information as well?)

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
```diff
@@ -0,0 +1,244 @@
...
+            out PreviousCustomEffect,
+            out PreviousTransformMatrix);
+
+        CurrentSortMode = hasSortMode ? sortMode!.Value : PreviousSortMode;
+        CurrentBlendState = hasBlendState ? blendState : PreviousBlendState;
+        CurrentSamplerState = hasSamplerState ? samplerState : PreviousSamplerState;
+        CurrentDepthStencilState = hasDepthStencilState ? depthStencilState : PreviousDepthStencilState;
+        CurrentRasterizerState = hasRasterizerState ? rasterizerState : PreviousRasterizerState;
+        CurrentCustomEffect = hasCustomEffect ? customEffect : PreviousCustomEffect;
+        CurrentTransformMatrix = hasTransformMatrix ? transformMatrix!.Value : PreviousTransformMatrix;
```
JaThePlayer: I don't like the separate bool arguments here, I think it's only relevant for `customEffect`, as passing null could either mean you want the default effect or inherit the effect from the last sprite batch.
From what I could tell, all other parameters could instead be done by deciding that `null = use previous value`.

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
SnipUndercover: In theory if there's no render targets then we do avoid an allocation but that won't be the case most of the time I believe. I can try to look into this.

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
SnipUndercover: I suppose that's right, this was written when the fields could in theory have `null` values and I wasn't sure whether `SpriteBatch.Begin` would accept them just fine. The `bool` is also relevant for the `RenderTarget2D` field.

JaThePlayer:
The `[MaybeNull]` annotations used wreak havoc on actual Nullable Reference Types: (this is from Frost Helper, which has NRT enabled, so the `target` parameter is specified to be not-null here)
obraz

EDIT:
It gets worse:
obraz

The attribute seemingly should be `[AllowNull]` instead, or the files could instead have NRT enabled via `#nullable enable` if we're going with nullability annotations anyway.

SnipUndercover:
> The `[MaybeNull]` annotations used wreak havoc on actual Nullable Reference Types: (this is from Frost Helper, which has NRT enabled, so the `target` parameter is specified to be not-null here)

I missed this, huh. I guess we can go with `#nullable enable` and using NRT here.

Celeste.Mod.mm/Celeste.Mod.mm.csproj
```diff
@@ -5,7 +5,6 @@
     <AssemblyName>Celeste.Mod.mm</AssemblyName>
     <RootNamespace>Celeste</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <LangVersion>11</LangVersion>
```
microlith57, line 8: **nitpick/query:** did you mean to remove this entirely? why not explicitly specify 12? i guess if it's the default that's fine but seems strange nevertheless

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
microlith57: yea, definitely agree. it's well within everest's remit to do this. if we just ignore access checks from everest to fna (while still requiring mods to do that manually themselves if applicable), that should be fine right?

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
```diff
@@ -0,0 +1,244 @@
...
+///   <b>Note: Restarting a spritebatch flushes it to the GPU.</b>
+///   While the cost is not that significant, it's best to avoid restarting the spritebatch too often per frame.
+/// </remarks>
+/// <seealso cref="TemporarySpriteBatchBuilder"/>
+public ref struct TemporarySpriteBatch
+{
+    // we don't really expect the RenderTargetBinding[] array size to exceed 1,
+    // so a max of 16 seems reasonable
+    private static readonly ArrayPool<RenderTargetBinding> _renderTargetPool
+        = ArrayPool<RenderTargetBinding>.Create(0x10, 50);
```
microlith57, line 32: **query:** assuming a max `RenderTargetBinding[]` size of 1, would this mean only 16 `TemporarySpriteBatch`es (that change RTs) can be nested inside each other?
..maybe that's fine, but would be good to document more explicitly here, and have a `catch` in case it happens for a more clear error message. just in case!

(also, what's the cost of this arraypool? if `RenderTargetBinding` is referencetype then isn't it basically free to make the capacity 256 or something and not worry about it at all?)

microlith57 (commented):
still uncomfortable with the DynamicData so this isn't an approval yet. but good progress, i like the `ArrayPool`!

Celeste.Mod.mm/Celeste.Mod.mm.csproj
SnipUndercover: I thought we were using C#11 since we're in .NET 8 but apparently not.
Not setting the language version explicitly makes it default to the latest one at the time of the release of the version specified by the TFM. In this case it would be C#12.
Though, since several other PRs have suggested bumping the language version, I think we should move that change to a separate PR.

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
SnipUndercover: I'm wondering what sort of patches would be necessary here to avoid having to use `DynamicData` without breaking mods who might be reflecting on it...

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
SnipUndercover: 16 refers to the max size of an individual array. 50 refers to how many of these can be pooled together into buckets.
I considered a max individual array size of 16 to be reasonable enough, and I just copied the second parameter from the parameterless constructor that `ArrayPool<T>.Create` calls. I can't say I understand its internals very well as I just glanced at them briefly.

`RenderTargetBinding` is a value type that is just a pointer and an `int`.

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
microlith57: i assume this is now resolved?

SnipUndercover: I didn't mark it as resolved as I wanted Ja to take a look to confirm this is fine, but I guess...

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
```diff
@@ -0,0 +1,244 @@
...
+    ///   The transformation <see cref="Matrix"/> of this <see cref="TemporarySpriteBatch"/>.
+    /// </summary>
+    public readonly Matrix CurrentTransformMatrix;
+
+    /// <summary>
+    ///   The <see cref="RenderTarget2D"/> swapped in for the duration of this <see cref="TemporarySpriteBatch"/>
+    ///   or <c>null</c> to draw to the screen if <see cref="HasRenderTarget"/> is <c>true</c>; else always <c>null</c>.
+    /// </summary>
+    public readonly RenderTarget2D? CurrentRenderTarget;
+
```
JaThePlayer, line 74: These `Current...` fields seem a bit questionable, since they're only used in the ctor and bloat the already-large struct even more.
The fields *could* be useful as a way to pass rendering info around various methods, but the Builder seems better to use for this purpose (since the code receiving a builder can use it to restart/change the batch etc).
If Everest exposes a public way of retrieving this info from the SpriteBatch directly, these become less important as well.

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatchBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+/// <remarks>
+///   This class lets users configure the creation of a new <see cref="TemporarySpriteBatch"/>, which allows
+///   users to interrupt an existing <see cref="SpriteBatch"/>, optionally swap in a <see cref="RenderTarget2D"/>
+///   and restart the <see cref="SpriteBatch"/> with custom properties.<br/>
+///   When done, the <see cref="SpriteBatch"/> is ended, the previous <see cref="RenderTarget2D"/> is restored and the
+///   old <see cref="SpriteBatch"/> is resumed.<br/>
+///   <b>Note: Restarting a spritebatch flushes it to the GPU.</b>
+///   While the cost is not that significant, it's best to avoid restarting the spritebatch too often per frame.
+/// </remarks>
+public sealed class TemporarySpriteBatchBuilder
```
JaThePlayer, line 20: Maybe worth making a record (struct?) to allow easily making a copy of the builder (to later change some of its settings)? Or maybe just a `Clone()` method instead.

If this doesn't become a `struct`, would be good to make the `TemporarySpriteBatch` ctor public for cases where caching the builder is not reasonable. (In which case, the ctor should probably have default values for all parameters so it's easy to pick and choose named parameters as needed).

Celeste.Mod.mm/Mod/Helpers/TemporarySpriteBatch.cs
JaThePlayer: I think adding get-only props to SpriteBatch would be fine, and wouldn't conflict with reflection.

microlith57 (changes requested):
formally requesting changes about the dynamicdata thing

PR #1097, opened 2026-03-25
make animationSpeed decalregistry handler accept floats

microlith57:
the underlying field is a `float`, so the handler should also accept floats

also fixed some whitespace


JaThePlayer (approved):

SnipUndercover (approved):
Considering it compiles I don't think testing before dev is necessary. It's just a type change  that we can easily roll back if for whatever reason it combusts.

PR #1098, opened 2026-03-31
implement ISpeed interface

microlith57:


swoolcock:
As I mentioned in Discord, I would recommend calling this something like `IHasSpeed`.

`INoun` implies that an implementation has a sole purpose (example, `IEnumerator`, `IList`).  `IAdjective` implies that the class can perform a certain functionality but it's not its only purpose (example, `IDisposable`, `IEnumerable`).  `IHasNoun` implies that the class provides data in addition to anything else it does (`IHasXmlNode`).

Edit:
Additionally, I would recommend NOT adding it to `Holdable`, since it's a component with convenience methods and doesn't actually use the property itself.

balt-dev:
don't forget PlayerSeeker!

PR #1104, opened 2026-04-06
implement a super fast data component.

wuke32767:
I'm not too glad with the current way to attach fields to an entity. `DynamicData.For` and `Component` are both too slow (to me).

That's why I made this. *Why would someone do this?*

## Advantage

Super fast.

## Disadvantage

- Add confusing sht to EVERY (EVERY!) entity.

- It's slower for the first time loading.

## Implementation

#1105 The two-dimensional array is torn apart by depth and scattered into EVERY subtype.


wuke32767:
:sobeline: i have noticed an issue which causes it to be not super fast. i'm going to close this pr.

PR #1105, opened 2026-04-06
Implement a fast enough data component.

wuke32767:
I'm not too glad with the current way to attach fields to an entity. `DynamicData.For` and `Component` are both slow (to me).

That's why I made this. *Why has no one ever done this before?*

## Advantage

Fast.

## Implementation

Now `Entity` holds a two-dimensional array. It stores all attached fields.

The first dimension stands for the depth of the entity in the inheritance tree.

Users can register their fields to a type, and the second dimension is the unique index they got.

Any fields are attached to only one type (and it's subtype), it can't work as a universal component.

It handles inheritance by depth dimension. Any child entity will be able to access fields attached to the parent entity, naturally.

Any type safe issue is done by hidding the implementation details.

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.Loader.cs
@@ -211,6 +211,8 @@ namespace Celeste.Mod {
                 Logger.Verbose("loader", $"ALL MODS LOADED IN {watch.ElapsedMilliseconds}ms");
                 Logger.Info("loader", $"Loaded {Everest._Modules.Count} modules");
 
+                DataComponentRegistry.Optimize();
+
                 try {
                     Watcher = new FileSystemWatcher {
                         Path = PathMods,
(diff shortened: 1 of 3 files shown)


balt-dev:
imo the exception messages here could _really_ use some work

microlith57 (changes requested):
apologies for the late review.

i strongly dislike this; it feels brittle and likely to cause difficult-to-diagnose issues later. perhaps this could be resolved by more thorough safety justification and better error messages, but i am still reluctant to introduce this risk into the codebase without good reason.

note that tools like ConditionalWeakTable are already available; is there a compelling reason why this is insufficient? are you able to measure the performance of this implementation in comparison to CWT? (and, if this is indeed more performant, is that benefit sufficient to justify the risks?)

wuke32767:
a) It's not fragile.

b) As an instance field, it works better with speedrun tool; And as a simple array, it doesn't introduce any DependentHandles and ConcurrentDictionaries.

c) Benchmark shows that
- CWT read is not too expensive (3 or 7 ns versus my 1 ns) ([see here](https://github.com/wuke32767/Private-SomeRandomBenchmark/wiki/Hate));
- CWT has gc impact ([see here](https://github.com/wuke32767/Private-SomeRandomBenchmark/wiki/Collect)).
- I have no idea about how to measure the impact to concurrent gc, cpu cache and gameplay, can't say what does it mean to real world program.

d) wait, am i responding to your question or change as requested?

PR #1106, opened 2026-04-07
Prevent inlining of Player.RefillStamina

maddie480:
`On.Celeste.Player.RefillStamina` seems to be a frequent enough use case... and with how the method is literally `Stamina = 110f;`, it's easily getting inlined _and_ it's not automatically slapped with NoInlining. I ran into inlining issues with Extended Variants, and the force-no-inlining-at-runtime method didn't help.


microlith57 (approved):
:yay:

DashingCat (approved):

PR #1107, opened 2026-04-07
Allow different cutscenes to play at the same time

maddie480:
The whole reason why #999 is held in beta since forever (4 months? wow) is that triggering multiple cutscenes at once was used... by TASers I think? The compromise is to block triggering of the cutscene if _a duplicate of it_ (as in, the same dialog, unless I missed something) is already running, not if _any_ cutscene is running.

### DIFF
--- a/Celeste.Mod.mm/Mod/Entities/DialogCutsceneTrigger.cs
+++ b/Celeste.Mod.mm/Mod/Entities/DialogCutsceneTrigger.cs
@@ -58,8 +58,8 @@ namespace Celeste.Mod.Entities {
             if (ignoreIntroState && ((patch_Player) player).IsIntroState)
                 return;
 
-            // don't activate if some dialog is already in progress
-            if (level.Tracker.GetEntity<Textbox>() is not null)
+            // don't activate if the same dialog is already in progress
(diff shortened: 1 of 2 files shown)


SnipUndercover (approved):

maddie480:
Branch updated to include the restoration of PR #999 as a separate commit

DashingCat (approved):

Celeste.Mod.mm/Mod/Entities/DialogCutsceneTrigger.cs
```diff
@@ -1,40 +1,77 @@
...
+            if (deathCount >= 0 && level.Session.DeathsInCurrentLevel != deathCount)
+                return;
+
+            if (ignoreIntroState && ((patch_Player) player).IsIntroState)
+                return;
+
+            // don't activate if the same dialog is already in progress
+            if (DialogCutscene.IsInProgress(dialogEntry))
                 return;
-            }
```
microlith57, line 29: **praise**: very nice, clean guard chain

microlith57 (approved):
cool! please inform lönn team about the new fields ^_^

PR #1108, opened 2026-04-12
Update Japanese localization

danesky-jp:
After more than two years, I would like to update Japanese.txt and relevant font files. The changes are limited to language files and should not affect other functionalities.


microlith57 (approved):
👍

DashingCat (approved):

PR #1112, opened 2026-04-23
Restore float jank

maddie480:
Violently ripping apart a dozen commits to try bringing back float jank :tm: (more seriously, preparing for the eventual switch of vanilla to Core, where the fixes should be removed if we want Everest to still match vanilla)


Celeste.Mod.mm/Patches/LavaRect.cs
```diff
@@ -14,127 +7,9 @@ public patch_LavaRect(float width, float height, int step) : base(width, height,
...
-            cursor.EmitDup();
-            cursor.EmitNewobj(m_Vector2_ctor);
-            cursor.EmitCall(m_Vector2_op_Addition);
-        }
-
-        public static void PatchLavaRectUpdate(ILContext context, CustomAttribute attrib) {
-            ILCursor cursor = new ILCursor(context);
-            
-            // Insert this code at the start of the method:
-            // if (!IsVisible())
```
Kalobi, line 129: Definitely some non-float-jank changes being caught in the crossfire here, this will need a manual pass of the actual patches that need to be removed.

maddie480:
Reverting to draft, it'll only become relevant if vanilla gets upgraded to .NET Core

PR #1113, opened 2026-04-29
prevent `ParticleType` instances from being kept alive forever

earthwise01:
Currently, the constructors for `Monocle.ParticleType` add all created instances to a private static `AllTypes` list, which is never used anywhere else. While this is fine in vanilla, this causes minor issues for mods which create new `ParticleType`s on the fly (e.g. to allow customization through `EntityData`, etc), as all created instances will be kept alive forever.

This PR fixes this by removing the `AllTypes.Add(this)` calls from both constructors, and also applies an `[Obsolete]` attribute to the list itself.

### DIFF
--- a/Celeste.Mod.mm/Patches/Monocle/ParticleType.cs
+++ b/Celeste.Mod.mm/Patches/Monocle/ParticleType.cs
@@ -1,0 +1,41 @@
+using MonoMod;
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


microlith57 (approved):
cool

Wartori54 (approved):

SnipUndercover:
:yay:

PR #1117, opened 2026-05-07
Fall back to mirrors on any exception when downloading mods

maddie480:
Someone in modding_help ran into this

```
System.Security.Authentication.AuthenticationException: Authentication failed because the remote party sent a TLS alert: 'AccessDenied'.
```

Turns out they had a security thing blocking files.gamebanana.com, and Everest didn't fall back to a mirror, because this exception is not part of the caught exceptions (`TimeoutException`, `WebException` or `IOException`). 

This PR makes Everest fall back to a mirror on any exception like Olympus, most errors come from networking anyway.

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/AutoModUpdater.cs
+++ b/Celeste.Mod.mm/Mod/UI/AutoModUpdater.cs
@@ -9,7 +9,6 @@ using System.IO;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Security.Authentication;
 
 namespace Celeste.Mod.UI {
     public class AutoModUpdater : Scene {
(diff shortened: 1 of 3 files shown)


Celeste.Mod.mm/Mod/UI/AutoModUpdater.cs
```diff
@@ -142,7 +143,7 @@ private void autoUpdate(SortedDictionary<ModUpdateInfo, EverestModuleMetadata> u
                             ModUpdaterHelper.VerifyChecksum(update, zipPath);
 
                             break; // out of the loop
-                        } catch (Exception e) when (e is WebException or TimeoutException or IOException) {
+                        } catch (Exception e) when (e is WebException or TimeoutException or IOException or AuthenticationException) {
```
SnipUndercover: The actual exception is a `System.AggregateException` that has an inner `System.Net.Http.HttpRequestException`, which *then* includes an `System.Security.Authentication.AuthenticationException`.
###### *(which finally includes an `System.ComponentModel.Win32Exception`.)*
```suggestion
                        } catch (Exception e) when (e is WebException or TimeoutException or IOException or AggregateException { InnerException: HttpRequestException }) {
```

SnipUndercover (changes requested):

SnipUndercover (approved):

microlith57 (approved):
no don't fall back on the mirror, that's fragile, you might break it

PR #1121, opened 2026-06-01
Add debug map option to critical error handler

balt-dev:
Adds an option to enter the debug level editor to the critical error handler.

### DIFF
--- a/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
+++ b/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
@@ -46,9 +46,9 @@ namespace Celeste.Mod.UI {
         public static CriticalErrorHandler CurrentHandler { get; private set; }
 
         // Done so that older mods that use this don't crash due to a linking error
-        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error) => HandleCriticalError(error, null);
+        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error) => HandleCriticalError(error, DisplayState.Initial);
         
-        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error, DisplayState? defaultDisplayState) {
+        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error, DisplayState defaultDisplayState) {
             if (!CoreModule.Settings.UseInGameCrashHandler)
                 return error;
 
--- a/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
+++ b/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
@@ -73,7 +73,7 @@ namespace Celeste.Mod.UI {
             
             // If this is a compatible scene, display as an overlay
             if (
-                defaultDisplayState is null or <= DisplayState.Overlay
+                defaultDisplayState <= DisplayState.Overlay
                 && CurrentHandler.State < DisplayState.Overlay 
                 && Celeste.Scene is Level lvl
             ) {
@@ -119,7 +119,7 @@ namespace Celeste.Mod.UI {
             }
 
             if (
-                defaultDisplayState is null or <= DisplayState.CleanScene
+                defaultDisplayState <= DisplayState.CleanScene
                 && CurrentHandler.State < DisplayState.CleanScene
             ) {
                 Scene oldScene = CurrentHandler.Scene;
@@ -139,7 +139,7 @@ namespace Celeste.Mod.UI {
 
             // If we aren't on a bluescreen scene, try that as a last resort
             if (
-                (defaultDisplayState is null or <= DisplayState.BlueScreen)
+                defaultDisplayState <= DisplayState.BlueScreen
                 && CurrentHandler.State < DisplayState.BlueScreen
             ) {
                 Scene oldScene = CurrentHandler.Scene;
@@ -827,13 +827,13 @@ namespace Celeste.Mod.UI {
                 "bluescreen" => DisplayState.BlueScreen,
                 _ => null
             };
-            if (state is null) {
+            if (state is not DisplayState validState) {
                 Engine.Commands.Log("Expected a display state of 'initial', 'overlay', 'cleanscene', or 'bluescreen'.", Color.Yellow);
                 return;
             }
             Exception exc = new(Calc.Random.NextSingle() > 0.95 ? "smots gaming" : "Hello, world!");
             ExceptionDispatchInfo.SetCurrentStackTrace(exc);
-            HandleCriticalError(ExceptionDispatchInfo.Capture(exc), state);
+            HandleCriticalError(ExceptionDispatchInfo.Capture(exc), validState);
         }
     }
 }
--- a/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
+++ b/Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
@@ -45,7 +45,10 @@ namespace Celeste.Mod.UI {
 
         public static CriticalErrorHandler CurrentHandler { get; private set; }
 
-        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error) {
+        // Done so that older mods that use this don't crash due to a linking error
+        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error) => HandleCriticalError(error, null);
+        
+        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error, DisplayState? defaultDisplayState) {
             if (!CoreModule.Settings.UseInGameCrashHandler)
                 return error;
 
@@ -64,12 +67,16 @@ namespace Celeste.Mod.UI {
                 CurrentHandler.EncounteredAdditionalErrors = true;
                 AmendLogFile(CurrentHandler.LogFile, "Encountered an additional critical error", error.SourceException);
             }
-
+            
             // Invoke the critical error event
             Everest.Events.CriticalError(CurrentHandler);
-
+            
             // If this is a compatible scene, display as an overlay
-            if (CurrentHandler.State < DisplayState.Overlay && Celeste.Scene is Level lvl) {
+            if (
+                defaultDisplayState is null or <= DisplayState.Overlay
+                && CurrentHandler.State < DisplayState.Overlay 
+                && Celeste.Scene is Level lvl
+            ) {
                 CurrentHandler.State = DisplayState.Overlay;
                 lvl.Add(CurrentHandler);
 
@@ -111,7 +118,10 @@ namespace Celeste.Mod.UI {
                 }
             }
 
-            if (CurrentHandler.State < DisplayState.CleanScene) {
+            if (
+                defaultDisplayState is null or <= DisplayState.CleanScene
+                && CurrentHandler.State < DisplayState.CleanScene
+            ) {
                 Scene oldScene = CurrentHandler.Scene;
                 CurrentHandler.RemoveSelf();
                 oldScene?.Entities?.UpdateLists();
@@ -128,7 +138,10 @@ namespace Celeste.Mod.UI {
             }
 
             // If we aren't on a bluescreen scene, try that as a last resort
-            if (CurrentHandler.State < DisplayState.BlueScreen) {
+            if (
+                (defaultDisplayState is null or <= DisplayState.BlueScreen)
+                && CurrentHandler.State < DisplayState.BlueScreen
+            ) {
                 Scene oldScene = CurrentHandler.Scene;
                 CurrentHandler.RemoveSelf();
                 oldScene?.Entities?.UpdateLists();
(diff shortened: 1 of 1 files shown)


Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -591,10 +616,16 @@ public override void Update() {
...
 
             // Update the options menu
-            if (Fade == 1)
-                optMenu?.Update();
+            if (Fade == 1) optMenu?.Update();
 
             base.Update();
+            
+            // Handle debug map key
+            if (Session != null && Celeste.PlayMode == Celeste.PlayModes.Debug && CoreModule.Settings.DebugMap.Pressed && Engine.Scene.Tracker.GetEntity<KeyboardConfigUI>() == null && Engine.Scene.Tracker.GetEntity<ButtonConfigUI>() == null) {
```
microlith57: **style:** can this be split onto multiple lines? it's a very long condition

microlith57 (approved):
lgtm, thanks for the contribution!

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
balt-dev: done!

Kalobi:
I'm not a fan of having an option cut off at the bottom and having stuff move around for no reason while navigating the menu when there's plenty of space on screen to hold all of the options.
Also, this is not within the scope of this PR since you were just following the structure of the existing code, but I think it might be a good idea to make these texts localizable.

SnipUndercover:
> Also, this is not within the scope of this PR since you were just following the structure of the existing code, but I think it might be a good idea to make these texts localizable.

We have a drafted #991 that does that.

balt-dev:
> I'm not a fan of having an option cut off at the bottom and having stuff move around for no reason while navigating the menu when there's plenty of space on screen to hold all of the options. Also, this is not within the scope of this PR since you were just following the structure of the existing code, but I think it might be a good idea to make these texts localizable.

The localization isn't my wheelhouse, and I didn't want to mess with the size of the player drawing in case someone else would take issue with that. It's also future-proofing for if more entries are added to the handler.

balt-dev:
Went and redid a lot of the UI code as per discussion with Kalobi on Discord.

balt-dev:
If I may request, could this be squash-merged if/when it's approved? It should make things cleaner.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -759,13 +845,29 @@ void DrawLineWrap(string text, float scale, Color color, Vector2 posOff = defaul
...
+            if (displayState is null) {
+                Engine.Commands.Log("Expected a display state of .", Color.Yellow);
+                return;
+            }
+
+            DisplayState state = displayState switch {
+                "overlay" => DisplayState.Overlay,
+                "cleanscene" => DisplayState.CleanScene,
+                _ => DisplayState.Initial,
+            };
```
Wartori54: Apart from having missed the `DisplayState.BlueScreen` in the switch, this is a bit misleading.
Calling `HandleCriticalError` with a given `DisplayState` will advance to the next possible display state, meaning `criterror overlay` will use a clean scene instead of being an overlay.
I don't think we should add the ability to force a `DisplayState` but this method should call `HandleCriticalError` with the greatest `DisplayState` that comes before the specified `displayState`, in my opinion.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -494,23 +517,65 @@ private bool ExecuteUserChoice(UserChoice choice) {
             return true;
         }
 
+        // Forcibly saves the game.
+        private void ManuallySaveGame() {
+            if (SaveData.Instance is null) return;
```
Wartori54: I'd rather throw here rather than return, all callers already check for this.
This could be static too.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -494,23 +517,65 @@ private bool ExecuteUserChoice(UserChoice choice) {
...
+            float availableSpace = MaxOptionsHeight - optMenu.Height;
+
+            float optY = (Celeste.TargetHeight - MaxOptionsHeight) / 2f;
+            if (UsePlayerSprite)
+                optY += Celeste.TargetHeight * PlayerSpriteSize;
+            else
+                optY += availableSpace / 2;
+
+            optMenu.Position = new Vector2(Celeste.TargetWidth * 0.15f, optY + optMenu.Height / 2);
+            optMenu.Justify = new Vector2(0.5f, 0.5f);
```
Wartori54: Not sure about this, feels like an overly complicated way to auto layout something that will only have two distributions: overworld menu, and in game menu (+ player or no player sprite).
Wouldn't something like the old code with adjusted constants work? Or just a fixed y position for the start of the buttons, and then "smartly" nudge them up if the list would go out of frame?
This menu will have to get rewritten entirely with submenus if we keep adding more buttons anyway.

Wartori54 (changes requested):
Good first PR! Just needs a bit more polish and taking a step back and thinking what is truly needed in here.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
balt-dev: I'm really torn on this. On the one hand, I physically can't fit the debug editor option into the old layout without changing the code somewhat, and so if I'm going to be changing the code at all, I'd rather do it well than half-ass it. On the other, though, I'd honestly rather go back to the old source code and shrink the scope of this PR, if at least to get the main feature I want (the debug editor option) merged, with the layout being tackled later on - I'm not the most comfortable working with Everest internals, and given I'm not a maintainer, I feel like anything greater than this in scope, and even what I have now, should fall under, while deserved, _very_ harsh criticism, since ANY change to Everest's internal systems has a very far-reaching impact. [I've been laughed out of a project before due to terrible code quality](https://github.com/0x0ade/CelesteNet/pull/151), and I don't want to repeat that.

balt-dev: If you really want me to, I can completely rewrite the menu to have submenus and such, but as it stands, that feels like the only option apart from rejecting this PR entirely. As I said, I'd rather do it well than half-ass it.

balt-dev:
!? I didn't close the PR...
Github?

balt-dev:
Here.

balt-dev:
Okay... one moment, then.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
balt-dev: > Not sure about this, feels like an overly complicated way to auto layout something that will only have two distributions: overworld menu, and in game menu (+ player or no player sprite). Wouldn't something like the old code with adjusted constants work? Or just a fixed y position for the start of the buttons, and then "smartly" nudge them up if the list would go out of frame? This menu will have to get rewritten entirely with submenus if we keep adding more buttons anyway.

@Wartori54 ...thinking about it more, I feel I might've gotten the wrong message from this. Could you elaborate on what you want me to _do_ here?

balt-dev:
I've decided that doing anything than just simply adding the debug map option is probably overscoping for this PR, given my lack of experience with Everest's codebase.

balt-dev:
> Ambiguous match found for 'Celeste.Mod.UI.CriticalErrorHandler System.Runtime.ExceptionServices.ExceptionDispatchInfo HandleCriticalError(System.Runtime.ExceptionServices.ExceptionDispatchInfo)'.

why would you compile my code then

balt-dev:
```
(08/29/2026 17:33:26) [Everest] [Error] [CelesteTAS] Failed to get method 'HandleCriticalError' on type 'Celeste.Mod.UI.CriticalErrorHandler'
--------------------------------
System.Reflection.AmbiguousMatchException: Ambiguous match found for 'Celeste.Mod.UI.CriticalErrorHandler System.Runtime.ExceptionServices.ExceptionDispatchInfo HandleCriticalError(System.Runtime.ExceptionServices.ExceptionDispatchInfo)'.
```
Are you kidding me?

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -595,6 +623,29 @@ public override void Update() {
...
+                    Input.MenuCancel.ConsumePress();
+                    ForceOptionImmediately(UserChoice.ReturnToMainMenu);
+                }
+            }
+        }
+
+        private void ForceOptionImmediately(UserChoice choice) {
+            if (!ExecuteUserChoice(choice)) return;
+            if (CurrentHandler == this) return;
+            RemoveSelf(); // The rest of the coroutine is never run
```
Wartori54, line 648: What's the purpose of these two lines?

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -766,6 +817,24 @@ void DrawLineWrap(string text, float scale, Color color, Vector2 posOff = defaul
...
+        
+        [Command("criterror", "Shows the Everest critical error handler. Must take a display state of 'initial', 'overlay', or 'cleanscene'.")]
+        internal static void CmdCriticalError(string displayState) {
+            DisplayState? state = displayState switch {
+                "initial" => DisplayState.Initial,
+                "overlay" => DisplayState.Overlay,
+                "cleanscene" => DisplayState.CleanScene,
+                "bluescreen" => DisplayState.BlueScreen,
+                _ => null
+            };
```
Wartori54: You forgot "bluescreen" in the help text.
Also having a default value of `displayState` was nice imo, you should bring that back.

Wartori54 (commented):
LGTM, just two non-blocking comments.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
balt-dev: Doing what happens after Fade immediately, so that I don't have to hack in a NoFade flag.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
balt-dev: Done.

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
Wartori54: Those two lines ensure there are no duplicate crash handlers, they are run at the end of the `Routine` as well, but don't affect the fade.

psyGamer (commented):
I've checked and the only mod which uses this method directly and would break is [FFFFF](https://gamebanana.com/mods/506049).
With the change however, it breaks CelesteTAS due to an ambigous match.

I am of course a bit biased, but imo it would make more sense to remove the backwards compat, since:
a) more people are using CelesteTAS and even if an update exists, people are known to not update their mods
b) It reduces tech debt
c) The developer of FFFFF has been made aware that internal Everest changes will not take the resonsbility for keeping the map working

Celeste.Mod.mm/Mod/UI/CriticalErrorHandler.cs
```diff
@@ -38,12 +39,16 @@ private enum UserChoice {
...
+            OpenDebugMap
         }
 
         public static CriticalErrorHandler CurrentHandler { get; private set; }
 
-        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error) {
+        // Done so that older mods that use this don't crash due to a linking error
+        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error) => HandleCriticalError(error, DisplayState.Initial);
+        
+        public static ExceptionDispatchInfo HandleCriticalError(ExceptionDispatchInfo error, DisplayState defaultDisplayState) {
```
psyGamer, line 51: I've checked and the only mod which uses this method directly and would break is [FFFFF](https://gamebanana.com/mods/506049).
With the change however, it breaks CelesteTAS due to an ambigous match.

I am of course a bit biased, but imo it would make more sense to remove the backwards compat, since:
a) more people are using CelesteTAS and even if an update exists, people are known to not update their mods
b) It reduces tech debt
c) The developer of FFFFF has been made aware that internal Everest changes will not take the resonsbility for keeping the map working

Wartori54: The counter argument to this is that `HandleCriticalError` is a public API and we should keep those working.
Also in CelesteTAS you could push out an update now that filters for the overload with the most amount of arguments, and that would work for current Everest and for future Everest once this change lands.
Adding a new overload when you need to introduce a new parameter on a public API is something that the BCL does extensively and I'd say its a good practice even.

psyGamer: I didn't really consider it a public API. I suppose it's `public` which is all the qualification it needs?

balt-dev: Public is public is public.

PR #1122, opened 2026-06-02
Fix the wind fix making the wires nearly static when there's no wind

SnipUndercover:
#981 introduced a bug that makes wires basically static when there's no wind if `fixWindBehavior` is set.
This is because `level.WindSine` was also being divided by `100f` which made it basically insignificant.

This PR changes this by just implementing a `ReducedVisualWind` property and using a ternary in `Render`.
image

The changes were tested in Celestecord and confirmed to work as intended.

### DIFF
--- a/Celeste.Mod.mm/Patches/Wire.cs
+++ b/Celeste.Mod.mm/Patches/Wire.cs
@@ -10,6 +10,10 @@ namespace Celeste {
     class patch_Wire : Wire {
         private bool fixWindBehavior;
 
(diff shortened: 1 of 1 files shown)


microlith57 (approved):
yup that wind sure is reduced (but not too much)

DashingCat (approved):
Works as expected, tested by setting the `fixWindBehavior` to ` true` to all wires, then comparing the behavior visually:
- before, the wires were nearly static
- after, the wires were moving slightly, as expected

PR #1103, opened 2026-04-05
potentially fix freeze with fast FMOD bank loading

wuke32767:
i'm not too sure how did it work

PR #1124, opened 2026-06-13
Fast FMOD Bank Loading Redux

maddie480:
Reapply #1079 + apply fixes from #1103

### DIFF
--- a/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
+++ b/Celeste.Mod.mm/Mod/Core/CoreModuleSettings.cs
@@ -162,6 +162,11 @@ namespace Celeste.Mod.Core {
         [SettingNeedsRelaunch]
         [SettingInGame(false)]
         [SettingIgnore] // TODO: Show as advanced setting.
+        public bool? FastFMODBankLoading { get; set; } = null;
+
+        [SettingNeedsRelaunch]
+        [SettingInGame(false)]
+        [SettingIgnore] // TODO: Show as advanced setting.
         public bool? ThreadedGL { get; set; } = null;
 
         [YamlMember(Alias = "FastTextureLoading")]
--- a/Celeste.Mod.mm/Patches/Audio.cs
+++ b/Celeste.Mod.mm/Patches/Audio.cs
@@ -11,9 +11,11 @@ using MonoMod;
 using SDL2;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Threading;
 
 namespace Celeste {
     static class patch_Audio {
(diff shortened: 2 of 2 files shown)


maddie480:
even though #1079 and #1103 were approved, I'm not merging this because I heard there were other issues with fast FMOD bank loading, not sure what though

Wartori54:
The bug-hunt status of the Fast FMOD Loading currently is:
- [x] Possible deadlock due to accesses to the Content map in the main thread, while audio is being loaded in parallel. (Fixed in what was #1103 and in this PR.)
- [ ] Duplicated GUIDs in banks can cause load order issues, where, if the same GUID is associated with multiple (and different) paths, the `IngestBank` call order matters. (There may be extra issues inside of FMOD itself for when a GUID is also duplicated in the `.bank` file as well, this is yet to be investigated.)

wuke32767:
my theory was, 

in old sequential load, it's:

load A.bank -> ingest A.guids.txt -> load B.bank -> ingest B.guids.txt -> ...

in new async load, there's one of the possibilities:

load A.bank -> load B.bank -> ingest A.guids.txt -> ingest B.guids.txt -> ...

the difference is, 
https://github.com/EverestAPI/Everest/blob/985fd82290fe47798bcd46cbbaa7574fe02780c8/Celeste.Mod.mm/Patches/Audio.cs#L223-L228
if A.guids.txt and B.guids.txt both declared an path<->guid whose corresponding event is only defined in B.bank, 

in old sequential load, A.guids.txt is filtered out because `only ingest the GUID if the corresponding event exists.` it is not loaded so does not exists.

in new async load, A.guids.txt is applied. then B.guids.txt is filtered out because the given ID is already registered.

if unfortunately A.guids.txt is wrong, it breaks.

---------------

i made two branches to bypass it. any of them can work, but which one is better?

### <https://github.com/wuke32767/Everest/commit/10b9bb505f0db91502c450200997b5cd65b0ab89>

this branch allows multiple paths to be mapped to the same guid.
side effect:
- guid will only be mapped to the first path.
- some event paths which is originally filtered out will be available again.
- there might be multiple different eventdescription with the same guid becomes accessible.

### <https://github.com/wuke32767/Everest/commit/d6bb3ee482f07f1d2ce7f0fd7c59317d710cd0cc>

in this branch, guid<->path in A.guids.txt is only registered when it's defined in A.bank.
side effect:
- will someone rely on it? like make multiple bank and only put one guid here?

DashingCat:
Thanks @wuke32767 for looking into the issues related to Fast FMOD Bank Loading.

About your two suggestions, I believe the second approach is better, because I think we can reasonably expect GUIDs to be defined in the related bank. If this assumption does not hold in practice, we could still fallback to the first approach.

Magedeline:
heya I am having an issue that the fmod bank as far I know of has been muted and try everything to get it fixes and it didn't work I was curious of if it is might be the update of everest

[log.txt](https://github.com/user-attachments/files/28931005/log.txt)

https://github.com/user-attachments/assets/84e199f0-733b-422a-9a12-842b5404cf3c

Magedeline:
I saw that snip

PR #1133, opened 2026-06-21
add more useful fields to InvisibleBarrier

microlith57:
seems useful

note the actual default soundindex is `Grass = 33` (apparently for bird nest reasons). a more sensible editor default would be `0` which is silent.

basic loenn plugin for testing:
```lua
local invisibleBarrier = {}

invisibleBarrier.name = "invisibleBarrier"
invisibleBarrier.fillColor = {0.4, 0.4, 0.4, 0.8}
invisibleBarrier.borderColor = {0.0, 0.0, 0.0, 0.0}
invisibleBarrier.placements = {
    name = "invisible_barrier",
    data = {
        width = 8,
        height = 8,
        allowClimbing = false, -- If disabled, blocks all climbing when the player is horizontally next to the barrier.
        surfaceIndex = 0, -- Footstep sound when walking on the barrier (or climbing, if enabled).
		allowStaticMovers = true -- If disabled, static movers like spikes won't attach to the barrier.
    }
}

return invisibleBarrier
```

### DIFF
--- a/Celeste.Mod.mm/Patches/InvisibleBarrier.cs
+++ b/Celeste.Mod.mm/Patches/InvisibleBarrier.cs
@@ -13,6 +13,8 @@ namespace Celeste {
                 Remove(Get<ClimbBlocker>());
 
             SurfaceSoundIndex = data.Int("surfaceIndex", SurfaceSoundIndex);
+
+            AllowStaticMovers = data.Bool("allowStaticMovers", true);
         }
 
     }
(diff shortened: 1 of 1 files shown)


Wartori54 (approved):
LGTM!

DashingCat (approved):

SnipUndercover:
Same here. @microlith57 Can you verify whether the patch is correct? My main concern is the patch deleting field initializers that happen before the `.ctor` call.

SnipUndercover:
Discussed with @Wartori54 on Discord, merging.

PR #1135, opened 2026-06-24
rework LegacyHook IsLegacyMMCaller

wuke32767:
relying on the stacktrace is not efficient and not safe.

jit may do tail call optimization, which can remove methods it should check from the stacktrace.

--------

this branch doesn't check StackTrace, instead it simply checks hookDelegate a hook uses.

downside:

will someone use some hook helper from other mods?


microlith57 (approved):
i am happy with this, i don't think the cross mod hook util thing is a big concern

Wartori54 (approved):
Since this only impacts hookgen its probably good enough :P

PR #1138, opened 2026-07-01
TileGrid public Tile type

TheDavSmasher:
This PR modifies the Autotiler/Tiles and TileGrid classes so TileGrid inherently keeps a reference to the tileset id that was used to fill that portion of the grid, as well as a method to get an id at a position from that grid. This is done in an effort to access the specific tileset on contact for further optionability based on tileset id instead of only via sound index.

### DIFF
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -11,7 +11,6 @@ using MonoMod.Utils;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
 using System.Xml;
 
 namespace Celeste {
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -606,9 +606,9 @@ namespace MonoMod {
             TypeDefinition autotiler = context.Method.DeclaringType;
 
             // char Celeste.Autotiler/TerrainType::ID
-            FieldReference f_terrainTypeID = autotiler.NestedTypes.First(t => t.Name == "TerrainType").FindField("ID");
+            FieldReference f_TerrainType_ID = autotiler.NestedTypes.First(t => t.Name == "TerrainType").FindField("ID");
             // char Celeste.Autotiler/Tiles::ID
-            FieldReference f_tilesID = autotiler.NestedTypes.First(t => t.Name == "Tiles").FindField("ID");
+            FieldReference f_Tiles_ID = autotiler.NestedTypes.First(t => t.Name == "Tiles").FindField("ID");
 
             cursor.GotoNext(MoveType.After, instr => instr.MatchStloc(5));
             cursor.GotoNext(MoveType.After, instr => instr.MatchStloc(4));
@@ -617,8 +617,8 @@ namespace MonoMod {
 
             cursor.EmitLdloc(4); // tiles (Celeste.Autotiler/Tiles)
             cursor.EmitLdarg1(); // data (Celeste.Autotiler/TerrainType)
-            cursor.EmitLdfld(f_terrainTypeID); // data.ID (char)
-            cursor.EmitStfld(f_tilesID);
+            cursor.EmitLdfld(f_TerrainType_ID); // data.ID (char)
+            cursor.EmitStfld(f_Tiles_ID);
         }
     }
 }
--- a/Celeste.Mod.mm/Patches/Autotiler.cs
+++ b/Celeste.Mod.mm/Patches/Autotiler.cs
@@ -62,7 +62,7 @@ namespace Celeste {
                         for (int num = Math.Min(i + 50, startX + tilesX); k < num; k++) {
                             int l = j;
                             for (int num2 = Math.Min(j + 50, startY + tilesY); l < num2; l++) {
-                                char tile = TileIdHandler(mapData, k, l, forceFill, forceID, behaviour);
+                                char tile = TileHandler(mapData, k, l, forceFill, forceID, behaviour).ID;
                                 tileGrid.TileIds[k - startX, l - startY] = tile;
                             }
                         }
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -515,5 +533,75 @@ public static void PatchAutotilerCtor(ILContext context, CustomAttribute attrib)
             cursor.EmitLdarg1(); // string filename (the xml path)
             cursor.EmitCallvirt(m_throwOnDuplicateId);
         }
+
+        public static void PatchAutotilerGenerate(ILContext context, CustomAttribute attrib) {
+            ILCursor cursor = new(context);
+
+            int x = 0;
+            int y = 0;
+            int tiles = 0;
```
microlith57: **nitpick:** personally i prefer `loc_` prefixes on variables holding local indices, so that it's clear these don't hold actual x, y coords or tile counts or whatever

microlith57 (approved):
yup

TheDavSmasher:
victory

Celeste.Mod.mm/Patches/Monocle/TileGrid.cs
```diff
@@ -1,13 +1,25 @@
...
+﻿#pragma warning disable CS0626 // Method, operator, or accessor is marked external and has no attributes on it
+
+using Celeste;
 using Microsoft.Xna.Framework;
+using MonoMod;
 
 namespace Monocle {
     class patch_TileGrid : TileGrid {
 
+        public VirtualMap<char> TileIds;
```
Wartori54, line 10: In `TileGrid` there are several methods that mutate the vanilla `VirtualMap<MTexture> Tiles` which should also get appropriate handling.
Unfortunately, those are hard to implement since there's no id we can trivially use, and getting proper values would imply creating a very ugly `Dictionary<MTexture, char>` or something of the sort. So I suggest adding an unknown value to the `TileIds` for when an id was lost.

Edit: Just realized that `Celeste.TileInterceptor` would need the same handling too.

Celeste.Mod.mm/Patches/Monocle/TileGrid.cs
```diff
@@ -38,5 +50,15 @@ public patch_TileGrid() : base(0, 0, 0, 0) {
...
+
+        public char TilesetIdAt(Vector2 readPosition) {
+            Vector2 position = Entity.Position + Position;
+            int num = (int) ((readPosition.X - position.X) / 8f);
+            int num2 = (int) ((readPosition.Y - position.Y) / 8f);
+            if (num >= 0 && num2 >= 0 && num < TilesX && num2 < TilesY) {
+                return TileIds[num, num2];
+            }
+            return default;
+        }
```
Wartori54, line 62: `readPosition` being a global position feels weird here, do you have any argument on why this should not be relative to the entity position?

Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -515,5 +533,92 @@ public static void PatchAutotilerCtor(ILContext context, CustomAttribute attrib)
...
+                instr => instr.MatchLdloc(out loc_tiles),
+                instr => instr.MatchLdfld(out f_tilesTextures),
+                instr => instr.MatchCall("Monocle.Calc", "Choose"),
+                instr => instr.MatchCallvirt("Monocle.VirtualMap`1<Monocle.MTexture>", "set_Item")))
+            {
+                // tileGrid.TileIds[{x} - startX, {y} - startY] = {tiles}.ID;
+
+                // tileGrid.TileIds
+                cursor.EmitLdloc0();
+                cursor.EmitLdfld(f_tileGridTiles.DeclaringType.Resolve().FindField("TileIds"));
```
Wartori54: Instead of resolving and finding the field each time just find the type in the method and start just like you do for `t_VirtualMap`.

Celeste.Mod.mm/Patches/Autotiler.cs
```diff
@@ -515,5 +533,92 @@ public static void PatchAutotilerCtor(ILContext context, CustomAttribute attrib)
...
+                cursor.EmitLdloc(loc_x);
+                cursor.EmitLdarg2();
+                cursor.EmitSub();
+                // y - startY
+                cursor.EmitLdloc(loc_y);
+                cursor.EmitLdarg3();
+                cursor.EmitSub();
+                // tiles.ID;
+                cursor.EmitLdloc(loc_tiles);
+                cursor.EmitLdfld(f_tilesTextures.DeclaringType.Resolve().FindField("ID"));
```
Wartori54: Same as the other `.Resolve().FindField` line above.

Wartori54 (changes requested):
The idea sounds good but (apart from the few comments I made) theres a major issue (related to the `TileIds` comment).
Mods also modify the values of `TileGrid.Tiles` opening up the possibility of `TileIds` to getting desynced, examples that I found are: [KoseiHelper](https://github.com/koseidiamond/KoseiHelper/blob/5754f10553bb0776f6448d4d6cba0f6125507d87/Source/Triggers/SwapTilesetTrigger.cs#L142), [FancyTileEntities](https://github.com/coloursofnoise/FancyTileEntities/blob/39087810b4735c1499b8c66dcd79e8418b7a2acb/FancyTileEntities/Entities/TileSeedController.cs#L74), [ScugHelper](https://github.com/balt-dev/ScugHelper/blob/3cca4b59ba8ee3141f6094f6b48f3e0e369841d9/Source/Entities/ProceduralTilemap.cs#L381)... and many others likely.
I feel like this still is an issue that should get addressed, even if the current impl covers 99% of cases, because users of the new API will get very bizarre bugs once something goes wrong.
The two solutions that come to mind for me are:
- Use relink patches to find mods modifying `TileGrid.Tiles`, and insert the appropriate code besides it, OR use relink patches to change the `TileGrid.Tiles` from a `TileGrid` to a new type were we manage both the tiles `VirtualMap<MTexture>` and the `VirtualMap<char>`.
- Introduce a dirty flag inside `VirtualMap<T>` for each element so we can catch assignments.

(Or just ask all modders to fix their code. :) )

But any other suggestion is welcome too.

Celeste.Mod.mm/Patches/Monocle/TileGrid.cs
TheDavSmasher: Not really, besides utility and code I based this on, but I can make an overload/alternative that doesn't utilize global position and instead uses room position?

Celeste.Mod.mm/Patches/Autotiler.cs
TheDavSmasher: I'll be honest and say I'm not sure what this means exactly you're asking to be changed

PR #1139, opened 2026-07-02
Fix MiniInstaller not copying FNA.dll into its workspace

Daxcr:
MiniInstaller wouldn't copy over FNA.dll into its workspace. Seems to be a Windows thing. Small fix though.

### DIFF
--- a/MiniInstaller/Program.cs
+++ b/MiniInstaller/Program.cs
@@ -91,6 +91,8 @@ namespace MiniInstaller {
 
                 DepCalls.LoadModders();
 
+                File.Copy(Path.Combine(Globals.PathGame, "FNA.dll"), moddedFNA, overwrite: true);
+
                 // DepCalls.ConvertToNETCore also converts dependencies, so FNA will also be copied to the workspace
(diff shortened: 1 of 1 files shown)


microlith57 (approved):
yup

MiniInstaller/Program.cs
```diff
@@ -91,6 +91,8 @@ public static int StandardMode(string[] args) {
 
                 DepCalls.LoadModders();
 
+                File.Copy(Path.Combine(Globals.PathGame, "FNA.dll"), moddedFNA, overwrite: true);
```
SnipUndercover, line 94: Note the comment below.

The copy does work correctly on FNA copies, but it seems that we completely forgot to test XNA copies which naturally don't have FNA.dll in orig.
*(partly because i didn't have an itch xna copy)*

I'm not sure whether just copying FNA.dll to the workspace without converting to .NET Core is fine.
Maybe we can check if vanilla is XNA and copy + coreify FNA.dll after coreifying vanilla?

SnipUndercover (changes requested):
See comment on line 94.

Daxcr:
Superseded by https://github.com/EverestAPI/Everest/pull/1140

PR #1141, opened 2026-07-02
Fix MiniInstaller not copying FNA.dll for XNA installs

Daxcr:


microlith57 (approved):
sure

MiniInstaller/Program.cs
```diff
@@ -91,9 +91,14 @@ public static int StandardMode(string[] args) {
 
                 DepCalls.LoadModders();
 
+                string libFNA = Path.Combine(Globals.PathGame, Path.Combine("everest-lib", "FNA.dll"));
+
                 // DepCalls.ConvertToNETCore also converts dependencies, so FNA will also be copied to the workspace
                 DepCalls.ConvertToNETCore(Path.Combine(Globals.PathOrig, "Celeste.exe"), moddedCeleste);
 
+                File.Copy(libFNA, moddedFNA, overwrite: true);
+                DepCalls.ConvertToNETCore(moddedFNA);
```
DashingCat:
```suggestion
                // Always bring FNA from "everest-lib"                                                                
                // XNA Celeste does not ship with FNA.dll, and the version we need may differ from the version shipped
                DepCalls.ConvertToNETCore(libFNA, moddedFNA);
```

I believe the `File.Copy` is not needed because `DepCalls.ConvertToNETCore` already overwrites the destination.

DashingCat (changes requested):
First of all, thanks for the PR!

I tested the change on Linux and it still works there.

I left a comment, and will approve once it is taken into account.

MiniInstaller/Program.cs
SnipUndercover: I agree, I suggested earlier to modify the PR as you suggested.

DashingCat (approved):

MiniInstaller/Program.cs
```diff
@@ -91,9 +91,15 @@ public static int StandardMode(string[] args) {
 
                 DepCalls.LoadModders();
 
+                string libFNA = Path.Combine(Globals.PathGame, Path.Combine("everest-lib", "FNA.dll"));
+
                 // DepCalls.ConvertToNETCore also converts dependencies, so FNA will also be copied to the workspace
```
SnipUndercover: Non-blocking nitpick: maybe clarify that the copied FNA is from lib-vanilla due to the SetupNativeLibs call?

SnipUndercover (approved):
Looks good, just one non-blocking nitpick. I'd also like to test whether installing this on FNA and XNA copies works fine as a sanity check.

SnipUndercover:
Do not merge yet, I haven't yet gotten around to making sure this works on both FNA and XNA.

SnipUndercover:
Tested and works on both XNA and FNA. Will commit some miscellaneous changes and merge.

PR #1142, opened 2026-07-06
Fix "start vanilla Celeste" on Steam macOS

maddie480:
I don't know much about that wizardry but according to my testing it seems to work

- Steam: holding right at the title screen works, running from Olympus works
- Steam with appid.txt: holding right at the title screen relaunches Everest
- itch: holding right at the title screen closes the game, but running from Olympus works


microlith57 (approved):

DashingCat (approved):
Works as expected on Linux, I am not able to test on other platforms.

SnipUndercover:
Windows is N/A since there is an earlier Windows-only branch, plus the two envvars do not do anything on Windows.

PR #1144, opened 2026-07-12
fix lua not load type in a safe manner

wuke32767:
i got a zip that works

[quirk.zip](https://github.com/user-attachments/files/29933741/quirk.zip) (compile it yourself)

### DIFF
--- a/Celeste.Mod.mm/Mod/Everest/Everest.LuaLoader.cs
+++ b/Celeste.Mod.mm/Mod/Everest/Everest.LuaLoader.cs
@@ -162,6 +162,10 @@ namespace Celeste.Mod {
                 }
 
                 foreach (Type type in asm.GetTypesSafe()) {
+                    PrecacheType(type);
+                }
+
+                static CachedType PrecacheType(Type type) {
                     // Non-public type instances can still be passed / returned.
                     /*
                     if (!type.IsPublic)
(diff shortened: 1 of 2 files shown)


DashingCat:
The mod in the zip is:

- quirkModule.cs
```cs
using System;

namespace Celeste.Mod.quirk;

public class quirkModule : EverestModule
{
    [MonoMod.Game]
    class YetAnother;
    public override void Load()
    {
    }

    public override void Unload()
    {
    }
}
```

- Ruler.cs
```cs
using Mono.Cecil;
using System;

namespace MonoMod
{
    [MonoModCustomAttribute(nameof(MonoModRules.Start))]
    public class GameAttribute : Attribute;

    public class MonoModRules
    {
        public static void Start(TypeDefinition type, CustomAttribute __)
        {
            var m = type.Module;
            var scope = new AssemblyNameReference("can you get it", new(1, 0, 0, 0));
            var btype = m.ImportReference(new TypeReference("who cares", "missing name", null, scope));
            type.BaseType = btype;
        }
    }
}
```

When attempting to load the mod without the changes, when Everest starts, it crashes with the following exception:
`System.IO.FileNotFoundException: Could not load file or assembly 'can you get it, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.`

It also logs the following line:
`[Warn] [modasmctx] Failed to load assembly 'can you get it, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' for module 'quirk'`

The exception does not show up when applying the changes, only leaving the warning.


If I understand correctly what is happening, the issue is that the non-public class `YetAnother` is getting cached by `LuaLoader`, which should not happen as the class is not public.

Did this behavior cause issue with existing or upcoming mods?

wuke32767:
the only real-world program may be ILHookDebugger 0.99.9, if it's real-world program

i think it's because GetNestedTypes is trying to load a type that can't be loaded

Celeste.Mod.mm/Mod/Everest/Everest.LuaLoader.cs
microlith57, line 173: **issue?:** isn't this the same for nonpublic nested types? but on line 178 you have `if (!type.IsNestedPublic) return null;`

wuke32767: i think it's because of this?
https://github.com/EverestAPI/Everest/pull/1144/changes#diff-bbae25c1a7b486378c5e49ec58129d0d598ed1c682fe6920dd2fcf0a845dd5c6L78-L79

microlith57: hm, i see
so i guess you're at least not introducing a new problem lol

realistically i don't think this is a big deal since nested private types aren't used that much and it should be easy to change later if it turns out to be an issue ^_^

microlith57 (approved):
yup

DashingCat (approved):

Issue #1126, opened 2026-06-13
Add trigger to call Level.TeleportTo()


grog398:
**I plan on implementing this myself**

**Why is this necessary?**

Level.TeleportTo is vanilla behavior typically used in cutscenes, however there is no way of calling the method outside of a custom cutscene.

Level.TeleportTo is a simple teleport that effectively acts as "starting a map" from another room (that is, it plays an arbitrary intro type like you started the map from that room), and is
- effectless
- stable
- seamless (with the Transition) intro type

and these qualities do not exist simultaneously in existing modded teleport triggers.

**Is there precedent for this being an Everest feature?**

Yes, features such as the Custom NPC and Dialog Cutscene triggers both expose vanilla cutscene functionality in an accessible way to mappers, and Level.TeleportTo() is exactly that, perhaps even moreso as it would be literally exposing a vanilla helper function to mappers through a trigger.

**Does this follow Everest's rules for new content?**

```
Content added must not change the functionality of the base game, or change any public facing API.
While many mods also make use of non-public code it is not as important to maintain backwards compatibility for it.
```
Yes, this does not modify anything, it simply adds a way to access existing vanilla behavior.

```
New features should only be added to improve or un-hardcode vanilla features, especially for general use in code mods (f.e. StrawberryRegistry, CustomNPC, Custom Events).
```
This improves vanilla features in the same way CustomNPC does, by exposing very basic behavior (Textbox.Say in the case of CustomNPC) to mappers.

```
Gameplay mechanics, entities, and other features not present in the base game are usually better suited for a mod, although some exceptions will be made on a case by case basis.
```
This is present in the base game.

**Planned implementation**
Create a trigger that adds a call to Level.TeleportTo to the Scene.OnEndOfFrame delegate.

**TLDR**

The feature has nearly identical precedent in other Everest features, would be trivial to implement, fits the template for new Everest content perfectly, and would provide very useful functionality for mappers that cannot be found in its particular form in existing mods.

PR #1146, opened 2026-07-19
Add trigger to call Level.TeleportTo()

grog398:
The rationale for this trigger is already described in the [linked issue for this feature](https://github.com/EverestAPI/Everest/issues/1126).

Some additional functionality added to the trigger not directly required to call Level.TeleportTo():
- An "Only Once" condition
- A flag condition


Celeste.Mod.mm/Mod/Entities/TeleportTrigger.cs
```diff
@@ -0,0 +1,55 @@
...
+        if (OnlyOnceFlag) RemoveSelf();
+    }
+
+    public override void OnEnter(Player player) {
+        base.OnEnter(player);
+
+        if (
+            string.IsNullOrEmpty(_nextLevel) ||
+            !Flag
+            ) return;
```
microlith57: **nitpick:** this formatting is. very strange
why not just this?:
```suggestion
        if (string.IsNullOrEmpty(_nextLevel) || !Flag)
            return;
```

Celeste.Mod.mm/Mod/Entities/TeleportTrigger.cs
```diff
@@ -0,0 +1,55 @@
...
+        
+        _onlyOnce = data.Bool("onlyOnce");
+        _flag = data.Attr("flag");
+
+        _onlyOnceFlag = $"{data.Level.Name}:{data.ID}_teleport_doNotTrigger";
+    }
+
+    public override void Added(Scene scene) {
+        base.Added(scene);
+        if (OnlyOnceFlag) RemoveSelf();
```
microlith57: **suggestion:** i dislike having a whole property that's only used once, especially since it casts the scene to Level. please remove the property and do something like this:
```suggestion
        if (Scene is not Level level || (_onlyOnce && level.Session.GetFlag(_onlyOnceFlag)))
            RemoveSelf();
```

Celeste.Mod.mm/Mod/Entities/TeleportTrigger.cs
```diff
@@ -0,0 +1,55 @@
...
+    private readonly string _flag;
+    private bool Flag => string.IsNullOrEmpty(_flag) || SceneAs<Level>().Session.GetFlag(_flag);
+    private readonly string _onlyOnceFlag;
+    private bool OnlyOnceFlag => _onlyOnce &&  SceneAs<Level>().Session.GetFlag(_onlyOnceFlag);
+    
+    public TeleportTrigger(EntityData data, Vector2 offset) : base(data, offset)
+    {
+        _nextLevel = data.Attr("nextLevel");
+        _introType = data.Enum("introType", Player.IntroTypes.None);
+        _nearestSpawn = new Vector2(data.Float("nearestSpawnX"),  data.Float("nearestSpawnY"));
```
microlith57, line 27: **question:** is this in destination-room coords or world coords? (will check when i get to my computer)

grog398: it's world coords

microlith57: in that case it would be nice to have some way to use dest room coords. perhaps a boolean or enum attr
this way people can move the destination room around without breaking things

microlith57 (approved):
yay!

Celeste.Mod.mm/Mod/Entities/TeleportTrigger.cs
```diff
@@ -0,0 +1,65 @@
...
+    private readonly string _onlyOnceFlag;
+
+    private readonly bool _useWorldCoordinates;
+    
+    public TeleportTrigger(EntityData data, Vector2 offset) : base(data, offset)
+    {
+        
+        _nextLevel = data.Attr("nextLevel");
+        _introType = data.Enum("introType", Player.IntroTypes.None);
+        _useWorldCoordinates = data.Bool("useWorldCoordinates");
```
Wartori54, line 26: Does it make sense to ever use world coordinates? I feel like using world coords is just more cumbersome than room relative coords. Furthermore this could be a node instead.

Celeste.Mod.mm/Mod/Entities/TeleportTrigger.cs
```diff
@@ -0,0 +1,65 @@
...
+        base.Added(scene);
+
+        if (scene is not Level level || (_onlyOnce && level.Session.GetFlag(_onlyOnceFlag))) {
+            RemoveSelf();
+            return;
+        }
+        
+        if (!_useWorldCoordinates || level.Session.MapData.Levels.FirstOrDefault(l => l.Name == _nextLevel) is not { } nextLevel) return;
+
+        _nearestSpawn += nextLevel.Position;
```
Wartori54, line 45: I don't think this is right, `TeleportTo` takes dest room relative coordinates, so it should be subtracted instead.

Celeste.Mod.mm/Mod/Entities/TeleportTrigger.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    public override void Added(Scene scene) {
+        base.Added(scene);
+
+        if (scene is not Level level || (_onlyOnce && level.Session.GetFlag(_onlyOnceFlag))) {
+            RemoveSelf();
+            return;
+        }
+        
+        if (!_useWorldCoordinates || level.Session.MapData.Levels.FirstOrDefault(l => l.Name == _nextLevel) is not { } nextLevel) return;
```
Wartori54, line 43: We should fail the teleport entirely (and maybe warn the user) instead of delaying the crash to somewhere else if the `_nextLevel` doesn't exist.

Celeste.Mod.mm/Mod/Entities/TeleportTrigger.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        if (string.IsNullOrEmpty(_nextLevel) || !Flag)
+            return;
+        
+        Scene.OnEndOfFrame += () => {
+            SceneAs<Level>().TeleportTo(
+                player, 
+                _nextLevel, 
+                _introType, 
+                _nearestSpawn);
```
Wartori54, line 59: It would be nice to allow for null values of `_nearestSpawn` since vanilla also implements that case.

Wartori54 (changes requested):

Wartori54 (approved):

PR #1149, opened 2026-08-01
Intern BinaryPacker.Element and EntityData values, and PlaybackData animation ids.

JaThePlayer:
Saves ~50MB RAM and reduces allocated object count by ~2 millions with Strawberry Jam.

More results starting from here: https://discord.com/channels/403698615446536203/429775439423209472/1533091706080067657

### DIFF
--- a/Celeste.Mod.mm/Mod/Helpers/InternHelper.cs
+++ b/Celeste.Mod.mm/Mod/Helpers/InternHelper.cs
@@ -1,0 +1,59 @@
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Celeste.Mod.Helpers;
+
+/// <summary>
+/// Helper class for interning boxed representations of simple types, used extensively in EntityData and BinaryPacker elements.
+/// </summary>
+public static class InternHelper {
+    private static readonly object True = true;
+    private static readonly object False = false;
+    
+    /// <summary>
+    /// Returns a boxed representation of the given value, cached between calls.
+    /// </summary>
+    public static object Intern(bool value) => value ? True : False;
+    
+    /// <summary>
+    /// Returns a boxed representation of the given value, cached between calls.
+    /// Be careful when using this method on arbitrary values, as once interned, the value will never get un-interned and will leak memory.
+    /// Use only when you can prove it's beneficial to do so.
+    /// </summary>
+    public static object Intern(int value) => InternHelper<int>.Intern(value);
+    
+    /// <summary>
+    /// Returns a boxed representation of the given value, cached between calls.
+    /// Be careful when using this method on arbitrary values, as once interned, the value will never get un-interned and will leak memory.
+    /// Use only when you can prove it's beneficial to do so.
+    /// </summary>
+    public static object Intern(float value) => InternHelper<float>.Intern(value);
+    
+    /// <summary>
+    /// Tries to intern the given value if deemed profitable.
+    /// Returns either the interned value, or the original if no interning occured.
+    /// Be careful when using this method on arbitrary values, as once interned, the value will never get un-interned and will leak memory.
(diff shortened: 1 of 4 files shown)


microlith57 (approved):

DashingCat (approved):
As discussed on Discord, this PR initially broke the "Fast Level Loading" (FLL) PR #1072, but this was taken into account in the latest commits.

Besides the RAM save, this PR has an impact on loading performance:
```
Without changes, FLL disabled (current state):
 - LEVELS LOAD: 8829ms
 - LEVELS LOAD: 8696ms
 - LEVELS LOAD: 8646ms

With changes, FLL disabled:
 - LEVELS LOAD: 8799ms
 - LEVELS LOAD: 8956ms
 - LEVELS LOAD: 9002ms

Without changes, FLL enabled:
 - LEVELS LOAD: 7072ms
 - LEVELS LOAD: 7172ms
 - LEVELS LOAD: 7186ms

With changes, FLL enabled:
 - LEVELS LOAD: 8452ms
 - LEVELS LOAD: 8247ms
 - LEVELS LOAD: 8169ms
```

While this PR introduces a slowdown while loading levels, assuming we can merge both this PR and #1037, we should still have better both loading times and RAM usage, so I approve this PR.

wuke32767:
here is my test result (only strawberry jam):
```
Slow Level Loading
 - LEVELS LOAD: 3046ms
 - LEVELS LOAD: 3165ms
 - LEVELS LOAD: 3074ms
Fast Level Loading
 - LEVELS LOAD: 5281ms
 - LEVELS LOAD: 4267ms
 - LEVELS LOAD: 4116ms
 - LEVELS LOAD: 4513ms
 - LEVELS LOAD: 4211ms
```
i'm trying two ways to optimize it a little, by replace the ConcurrentDictionary with _, but failed:
```
Fast Level Loading (Immutable)
 - LEVELS LOAD: 4556ms
 - LEVELS LOAD: 4536ms
 - LEVELS LOAD: 5490ms
 - LEVELS LOAD: 4158ms
 - LEVELS LOAD: 4539ms
Fast Level Loading (ThreadStatic)
 - LEVELS LOAD: 4272ms
 - LEVELS LOAD: 4608ms
 - LEVELS LOAD: 4408ms
```

wuke32767:
wait, i think i should nitpick it.
the FLL compatible stuff should be removed for now and opened in a new pr (or in FLL pr). currently this pr makes FLL slower.
[non-blocking]

Wartori54:
FLL got delayed, we should probably switch to `Dictionary` to reduce overhead before merging.

JaThePlayer:
> FLL got delayed, we should probably switch to `Dictionary` to reduce overhead before merging.

Wrote a quick benchmark for it, I think its fine to leave as `ConcurrentDictionary` as it only incurs a ~1ns overhead if there's no contention, and this way future multithreading PRs don't have to worry about changing this / mods that wish to use these helpers don't have to worry about thread safety.

```
public class DictAccess {
    [Benchmark]
    public object Dict() => InternHelper<int>.Intern(3);
    [Benchmark]
    public object ConcurrentDict() => InternHelperThreadSafe<int>.Intern(3);
}

internal static class InternHelper<T> {
    private static readonly Dictionary<T, object> Cache = new();
    public static object Intern(T value) {
        return Cache.TryGetValue(value, out object result) ? result : Cache[value] = value;
    }
}

internal static class InternHelperThreadSafe<T> {
    private static readonly ConcurrentDictionary<T, object> Cache = new();
    public static object Intern(T value) {
        return Cache.GetOrAdd(value, static x => x);
    }
}

| Method         | Mean     | Error     | StdDev    |
|--------------- |---------:|----------:|----------:|
| Dict           | 3.736 ns | 0.0528 ns | 0.0468 ns |
| ConcurrentDict | 4.634 ns | 0.0390 ns | 0.0365 ns |
```

PR #1152, opened 2026-08-03
Rework and port HookGen from MonoMod

psyGamer:
This ports the HookGen code from MonoMod and adjusts to do avoid running MonoMod over it and instead directly integrates our patches into the generated assembly.
With this port, there are also three notable changes:
1. IL-hooks now target the `orig_` method if available. This avoids breaking mods when Everest starts patching a method, since that will move the vanilla IL instructions into the `orig_` method. If the intention is to specifically hook Everest, manual hooks should be used, just like with all other Everest hooks.
2. Some vanilla methods were inaccessible via HookGen. To further strengthen the above argument, HookGen now generates events for properties, lambdas and state-machines.
3. The critiria for "Everest-internal" now matches to publicizer in the sense that everything not present inside the vanilla assembly will now be an error. This won't break existing mods, but might require some code changes to compile them again.

There are a handful of mods for which this is a breaking change, those would need to be updated before getting merged.


Everest.slnx
microlith57: ❤️

PR #1153, opened 2026-08-03
Optimize Launch Speed: Parallelize startup mod loading and batch per-target ILHook application

std-microblock:
Replaces the serial startup loader with a dependency-aware DAG scheduler that runs independent modules in parallel, and batches startup ILHooks so each target method's IL chain is rebuilt only once.

Loader:
- Build the module dependency graph once (Kahn topological sort) instead of rescanning the delayed list after every module; dynamic DAG scheduling
- Run assembly load, GetTypes, module construction, LoadSettings and Load in parallel across a bounded worker pool, with a dedicated PGO reserved lane for historically expensive nodes.
- Profile-guided list scheduling starts the longest critical paths first; per-mod timings are persisted and learned across boots (LoaderTimingCache).

Caches:
- Persistent ZIP metadata cache (everest.yaml/yml, .everestignore and a DLL/PDB entry index) avoids reopening every container on each boot.
- Persistent stat-validated checksum cache with single-flight hashing avoids re-reading unchanged files (LoaderChecksumCache).

Relinker:
- Warm relink-cache hits load under the per-mod assembly context lock only, keeping cached loads parallel; cache misses retain the global relinker lock.

ILHook:
- Added [ILHookTransaction to MonoMod](https://github.com/std-microblock/MonoMod/tree/ilhook-transaction): hooks queue during startup and are committed once per target method, then flushed with per-mod manipulator serialization and bounded target parallelism. A process-wide safety gate serializes ILHook apply and legacy ModInterop calls during parallel startup. (thanks to @WEGFan for inspiration here xD).

Diagnostics:
- Low-overhead LoaderProfiler and StartupProfiler instrumentation across the loader, boot and content phases.

Measured on a ~170-mods instance (24 logical CPUs, warm caches):
- Loader internal: ~40.1s serial -> ~5s.
- Process start to "Loaded 180 modules": ~44.98s serial -> ~9.0s.
- ILHook apply aggregate dropped from ~80s to a single per-target flush of ~1-3.6s.

Added envvars for disabling optimizations: `EVEREST_PARALLEL_LOAD` `EVEREST_ILHOOK_STARTUP_TRANSACTION` `EVEREST_LOADER_PGO_REORDER`

Note: This pull request is written under the assistance of AI


std-microblock:
about strange changes for `Crawl`: CollabUtils2 hooked the Crawl method and the normal way breaks it.

PR #1154, opened 2026-08-03
Expand contributing guidelines

Kalobi:
Add some points that have recently come up to the contributing guidelines. Suggestions welcome.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
 # Contributing
-Contributions are very welcome and greatly appreciated! Make sure to join the [Celeste Community Discord](discord.gg/Celeste) for any help you need.
+Contributions are very welcome and greatly appreciated! Make sure to join the [Celeste Community Discord](discord.gg/Celeste) for any help you need. The #modding_dev channel there is the best place to discuss contributions to Everest.
 
 ## Table of Contents
 - [Contributing](#contributing)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -32,14 +32,14 @@ Once you have a fork, clone it to your machine and follow the instructions [here
 These are guidelines for some of the different types of contributions you can make to Everest.
 
 ## Bug Reports
-Make sure bug reports have enough information to be reproduced easily!  
+Make sure bug reports have enough information to be reproduced easily!
```
SnipUndercover, line 35:
```suggestion
Make sure bug reports have enough information to be reproduced easily!\
```
In my opinion the flow of the document is better with the line breaks preserved, so this applies to all the line diffs where the only change is the removal of the double trailing whitespace.
Fortunately, ending a line with a backslash is another way to achieve a line break that's less prone to breaking due to autoformatters.

This suggestion applies to other diffs where the double trailing whitespace was removed.

SnipUndercover (changes requested):
Moving this from `#modding_dev` to here to keep a paper trail.
Markdown has significant trailing whitespace; putting two or more whitespaces at the end of a line inserts a line break (HTML `<br/>`), as opposed to two blank lines, which create a paragraph break (HTML `</p>` and `<p>`).

PR #1159, opened 2026-08-09
Fix symlinks on unsupported filesystems on OSes other than Windows

SnipUndercover:
Currently, installing Everest to a drive with a filesystem that does not support symlinks (like FAT32, exFAT, and others) will crash MiniInstaller completely, instead of falling back to copying the file.

Normally Windows would have shown a message box for this case, but neither Linux or macOS had fallback mechanisms for that.

This PR adds the prompt to other filesystems, leveraging SDL2's [`SDL_ShowMessageBox`](https://wiki.libsdl.org/SDL2/SDL_ShowMessageBox) API that's cross-compatible. Might as well use the SDL2 we ship, right?
It will fall back to copying files if the user chooses to proceed, and will exit early if the user chooses to cancel.

Furthermore, the elevation prompt has also been changed to use SDL's API.

Keeping as draft until all OSes have been tested following the TODO list below.

> [!NOTE]
> Most of the daunting +9300 additions are caused by a copy-paste of `EverestSplash`'s SDL bindings.

## TODOs
- [x] Windows
  - [x] Install on a filesystem that DOES support symlinks and WITH elevation *(NTFS tested)*
  - [x] Install on a filesystem that DOES support symlinks but WITHOUT elevation *(NTFS tested)*
  - [x] Install on a filesystem that DOES NOT support symlinks *(FAT32 tested)*
- [x] Linux
  - [x] Install on a filesystem that DOES support symlinks
  - [x] Install on a filesystem that DOES NOT support symlinks
- [ ] macOS
  - [ ] Install on a filesystem that DOES support symlinks
  - [ ] Install on a filesystem that DOES NOT support symlinks


MiniInstaller/LibAndDepHandling.cs
```diff
@@ -89,7 +89,14 @@ void CopyNativeLib(string src, string dst) {
 
                 if (symlinkPath != null && symlinkPath != dst) {
                     File.Delete(symlinkPath);
-                    File.CreateSymbolicLink(symlinkPath, Path.GetRelativePath(Path.GetDirectoryName(symlinkPath)!, dst));
+
+                    if (Symlinks.Supported)
+                        File.CreateSymbolicLink(symlinkPath, Path.GetRelativePath(Path.GetDirectoryName(symlinkPath)!, dst));
+                    else if (Symlinks.NeedElevation)
+                        // we never had to do this for Windows
+                        throw new NotImplementedException();
```
microlith57: **nitpick:** is this necessary? why not simply remove this:
```suggestion
```
and fall back to the copy? (it would make more sense, and you can keep the comment saying it shouldn't happen if you want)

DashingCat:
I tested the changes on Linux, and as far as I can tell it works properly on my end:
- running MiniInstaller normally creates the symlinks as expected,
- simulating an unsupported filesystem using `strace -e 'trace=symlink' -e 'fault=symlink:error=EPERM' --follow-forks` raises the SDL/zenity dialog, and my filesystem was correctly identified

PR #1161, opened 2026-08-21
Separate search key from QuickRestart

Saplonily:
Added a separate key binding entry `MenuSearch` for Search. This includes:

- in-game options
- oui mod options
- oui mod toggler
- SearchToolTip texture
- oui chapter select


microlith57 (approved):

DashingCat (approved):
Tested and works as expected.

PR #1162, opened 2026-08-22
Fix `CassetteListener.EntityAdded` to use `Level.ShouldCreateCassetteManager`

aonkeeper4:
This PR makes `CassetteListener.EntityAdded` use `Level.ShouldCreateCassetteManager` directly when checking whether to add a `CassetteBlockManager`, instead of duplicating the logic. This makes mods hooking `ShouldCreateCassetteManager` (e.g. custom cassette blocks, most notably Quantum Mechanics' wonky cassette blocks) work properly with cassette listeners, as previously this would have required a hook into Everest.

### DIFF
--- a/Celeste.Mod.mm/Patches/Level.cs
+++ b/Celeste.Mod.mm/Patches/Level.cs
@@ -46,6 +46,11 @@ namespace Celeste {
 
         public SubHudRenderer SubHudRenderer;
 
+        // we want to expose `ShouldCreateCassetteManager` to `CassetteListener` without modifying its visibility, so create an internal property that forwards to it
+        [MonoModIgnore]
+        private bool ShouldCreateCassetteManager { get; }
+        internal bool _ShouldCreateCassetteManager => ShouldCreateCassetteManager;
+
         public class LoadOverride {
 
             public Player NextLoadedPlayer = null;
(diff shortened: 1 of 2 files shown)


Celeste.Mod.mm/Patches/Level.cs
```diff
@@ -46,6 +46,11 @@ class patch_Level : Level {
 
         public SubHudRenderer SubHudRenderer;
 
+        // we want to expose `ShouldCreateCassetteManager` to `CassetteListener` without modifying its visibility, so create an internal property that forwards to it
+        [MonoModIgnore]
+        private bool ShouldCreateCassetteManager { get; }
+        internal bool _ShouldCreateCassetteManager => ShouldCreateCassetteManager;
```
microlith57, line 52: why not just. make it public?

microlith57 (approved):

Celeste.Mod.mm/Patches/Level.cs
aonkeeper4: people hook it manually :(

microlith57: :cryplant:

ah well..

PR #1168, opened 2026-08-31
Build mirror URLs using the MirrorName field of everest_update.yaml

maddie480:
... instead of using rad URL substrings that are sure to break if the URLs deviate from the current GameBanana formats in any way.

This mirrors (🥁) the change made in https://github.com/EverestAPI/Olympus/commit/a193d0c46c273711bfbdc8a3ca45af402122c3b2.


maddie480:
The pipeline fails because Azure is doing a death

maddie480:
okay we're good

microlith57 (approved):

